PR #16, opened 2019-10-17
Add conversion between float and int

johnwikman:
Adds conversion expressions for int-to-float and for float-to-int. The mnemonic for the float-to-int conversions (`floorfi`/`ceilfi`/`roundfi`) is
 - `%fi`: Performs `%` on a **f**loat and returns an **i**nt.


EliasC:
This looks good! I fixed some conflicts that showed up since it took some time to merge this.

PR #18, opened 2019-11-12
Added some seq operations

br4sco:
Added operations over sequences for appending, reversing, concatenation and foldr

### DIFF
--- a/stdlib/seq.mc
+++ b/stdlib/seq.mc
@@ -19,12 +19,8 @@ let foldl1 = lam f. lam l. foldl f (head l) (tail l)
 let foldr = fix (lam foldr. lam f. lam acc. lam seq.
     if null seq
     then acc
-    else f (foldr f acc (tail seq)) (head seq)
+    else f (head seq) (foldr f acc (tail seq))
 )
-
-let rev = lam seq.
-    let f = lam acc. lam x. cons x acc in
-    foldl f [] seq
 
 let foldr1 = lam f. lam seq. foldl1 f (reverse seq)
 
@@ -45,12 +41,8 @@ let all = fix (lam all. lam p. lam seq.
   then true
   else and (p (head seq)) (all p (tail seq)))
 
--- Append and concat
-let append = lam seq1. lam seq2.
-    let f = lam acc. lam x. cons x acc in
-    foldr f seq2 seq1
-
-let concat = lam seqs. foldl append [] seqs
+-- Join
+let join = lam seqs. foldl concat [] seqs
 
 -- Searching
 let filter = fix (lam filter. lam p. lam seq.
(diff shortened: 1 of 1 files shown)


stdlib/seq.mc
```diff
@@ -31,6 +43,13 @@ let all = fix (lam all. lam p. lam xs.
   then true
   else and (p (nth xs 0)) (all p (slice xs 1 (length xs))))
 
+-- Append and concat
+
+let append = lam seq1. lam seq2.
+    let f = lam acc. lam x. cons x acc in
+    foldr f seq2 seq1
+
+let concat = lam seqs. foldl append [] seqs
```
EliasC: `append` already exists in the built-ins with the name `concat`. I suggest replacing this chunk by:
```
let join = lam  seqs. foldl concat [] seqs
```

stdlib/seq.mc
```diff
@@ -14,14 +14,26 @@ let foldl = fix (lam foldl. lam f. lam acc. lam seq.
...
 
+let foldr = fix (lam foldr. lam f. lam acc. lam seq.
+    if null seq
+    then acc
+    else f (foldr f acc (tail seq)) (head seq)
+)
+
+let rev = lam seq.
+    let f = lam acc. lam x. cons x acc in
+    foldl f [] seq
```
EliasC: There is already a built-in version of this called `reverse`.

stdlib/seq.mc
```diff
@@ -14,14 +14,26 @@ let foldl = fix (lam foldl. lam f. lam acc. lam seq.
...
+
+let foldr1 = lam f. lam seq. foldl1 f (reverse seq)
```
EliasC, line 25: Not sure if `foldr f (head seq) (tail seq)` would be faster (no reverse, but also not tail-recursive).

stdlib/seq.mc
```diff
@@ -14,14 +14,26 @@ let foldl = fix (lam foldl. lam f. lam acc. lam seq.
 )
 let foldl1 = lam f. lam l. foldl f (head l) (tail l)
 
+let foldr = fix (lam foldr. lam f. lam acc. lam seq.
+    if null seq
+    then acc
+    else f (foldr f acc (tail seq)) (head seq)
```
EliasC: Should the arguments to `f` be flipped? My feeling is that `foldr cons [] [1,2,3]` should be `[1,2,3]`, whereas now it is an error.

EliasC:
Resolved some conflicts, so you might want to pull before addressing any issues!

stdlib/seq.mc
br4sco: If you do `head [1,2,3]` you get `1` but when folding from right you want `3` to be the first value of the accumulator right?

stdlib/seq.mc
br4sco: I guess we could, but I think the above definition is consistent with foldl, i.e. `f : acc -> x -> acc` so `foldl cons [] [1,2,3]` also gives an error.

stdlib/seq.mc
EliasC: You're right! Maybe you have to implement it from scratch to get rid of the reverse. Let's just leave as is then!

stdlib/seq.mc
br4sco: No your right, for `foldr` we should have  `f: x -> acc -> acc` but for `foldl` then `f: acc -> x -> acc`.

EliasC:
Looks good! Merging...

PR #19, opened 2019-11-12
Functions for seq.mc

lingmar:
Extends the sequence library with filter, find and paritition functions.


EliasC:
Looks good! Merging!

PR #20, opened 2019-11-12
Add simple string operations to stdlib

johnwikman:
Adds some string operations to stdlib. Some of these are copied from the stringops.mc example.

### DIFF
--- a/stdlib/string.mc
+++ b/stdlib/string.mc
@@ -33,15 +33,30 @@ let int2string = fix (lam int2string. lam n.
 
 // Returns an option with the index of the first occurrence of c in s. Returns
 // None if c was not found in s.
-let strfind = lam c. lam s.
-  let strfind_abs = fix (lam strfind. lam i. lam c. lam s.
+let strindex = lam c. lam s.
+  let strindex_abs = fix (lam strindex. lam i. lam c. lam s.
     if eqi (length s) 0
     then None
     else if eqchar c (head s)
          then Some(i)
-         else strfind (addi i 1) c (tail s)
+         else strindex (addi i 1) c (tail s)
   ) in
-  strfind_abs 0 c s
+  strindex_abs 0 c s
+
+// Returns an option with the index of the last occurrence of c in s. Returns
+// None if c was not found in s.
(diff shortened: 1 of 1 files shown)


stdlib/string.mc
```diff
@@ -29,6 +31,63 @@ let int2string = fix (lam int2string. lam n.
...
+    if eqi (length s) 0 then
+      if eqi acc (negi 1)
+      then None
+      else Some(acc)
+    else
+      if eqchar c (head s)
+      then strrindex (addi i 1) i   c (tail s)
+      else strrindex (addi i 1) acc c (tail s)
+  ) in
+  strrindex_abs 0 (negi 1) c s
```
EliasC: May I suggest a name that is more than one letter different from `strindex`? Maybe call them `strIndex` and `strLastIndex`?

EliasC: (or non-camelCase, if you prefer. We're going to have to go over the whole standard library at some later point to make naming convention the same all over)

stdlib/string.mc
```diff
@@ -29,6 +31,63 @@ let int2string = fix (lam int2string. lam n.
     concat (int2string (divi n 10)) d
 )
 
+// Returns an option with the index of the first occurrence of c in s. Returns
+// None if c was not found in s.
+let strindex = lam c. lam s.
+  let strindex_abs = fix (lam strindex. lam i. lam c. lam s.
```
EliasC: Minor comment: the `foo_abs` naming convention is something I used in the experiments where I was separating `fix` from definition. It does not "mean" anything in particular.

EliasC:
Looking good in general! Some minor comments regarding naming. Let me know when you have adressed (or refuted!) these comments, and I will merge.

stdlib/string.mc
johnwikman: Done. I usually prefer a non-camelCase naming scheme, but it is better to have a consistent naming scheme (I saw that you use camelCase for `zipWith` etc.)

EliasC: Awesome! Merging now...

PR #21, opened 2019-11-13
Add functions for options to the stdlib

segeljakt:
This PR adds the following functions to the stdlib (derived from [rust-stdlib](https://doc.rust-lang.org/std/option/enum.Option.html), also MIT-licensed).

* `optionIsNone`: Returns `true` if the option is a `Some` value.
* `optionIsSome`: Returns `true` if the option is a `None` value.
* `optionContains`:  Returns `true` if the option contains a value which satisfies the specified predicate.
* `optionMap`: Applies a function to the contained value (if any).
* `optionFlatMap`: Applies a function to the contained value (if any), and flattens the result.
* `optionMapOr`: Applies a function to the contained value (if any), or returns the provided default (if not).
* `optionMapOrElse`: Applies a function to the contained value (if any), or computes a default (if not).
* `optionAnd`: Returns `None` if either option is `None`, otherwise returns the first option.
* `optionFilter`: Filters the contained value (if any) using the specified predicate.
* `optionOr`: Returns the first option if it contains a value, otherwise returns the second option.
* `optionOrElse`: Returns the option if it contains a value, otherwise calls the specified function and returns the result.
* `optionXor`: Returns the first option if exactly one option is `Some`, otherwise returns `None`.
* `optionFlatten`: Converts from `Option<Option<T>>` to `Option<T>`.

Some tests are not passing yet, I'm not sure why 😕.

### DIFF
--- a/stdlib/option.mc
+++ b/stdlib/option.mc
@@ -2,3 +2,169 @@ type Option
 
 con Some : Dyn -> Option
 con None : () -> Option
+
+-- Returns `true` if the option is a `Some` value.
+let optionIsSome: Option -> Bool = lam o.
+  match o with Some _ then
+    true
+  else
+    false
+
+-- Returns `true` if the option is a `None` value.
+let optionIsNone: Option -> Bool = lam o.
+  match o with None () then
+    true
+  else
+    false
+
+-- Returns `true` if the option contains a value which
+-- satisfies the specified predicate.
+let optionContains: Option -> (a -> Bool) -> Bool = lam o. lam p.
+  match o with Some t then
+    p t
+  else
+    false
+
+-- Applies a function to the contained value (if any).
+let optionMap: Option -> (a -> a) -> Option = lam o. lam f.
+  match o with Some t then
+    Some(f t)
+  else
+    None ()
+    
+-- Applies a function to the contained value (if any),
+-- or returns the provided default (if not).
+let optionMapOr: Option -> a -> (a -> a) -> a = lam o. lam d. lam f.
+  match o with Some t then
+    f t
+  else
+    d
+
+-- Applies a function to the contained value (if any),
+-- or computes a default (if not).
+let optionMapOrElse: Option -> (Unit -> a) -> (a -> a) -> a = lam o. lam d. lam f.
+  match o with Some t then
+    f t
+  else
+    d unit
+
+-- Returns `None` if either option is `None`, otherwise returns
+-- the first option.
+let optionAnd: Option -> Option -> Option = lam o1. lam o2.
+  match o1 with Some _ then
+    match o2 with Some _ then
+      o1
+    else
+      None ()
+  else
+    None ()
+
+-- Returns `None` if the option is `None`, otherwise calls the
+-- specified function on the wrapped value and returns the result.
+let optionFlatMap: Option -> (a -> Option) -> Option = lam o. lam f.
+  match o with Some t then
+    f t
+  else
+    None ()
+
+-- Filters the contained value (if any) using the specified predicate.
+let optionFilter: Option -> (a -> a) -> Option = lam o. lam p.
+  match o with Some t then
+    if p t then
+      o
+    else
+      None ()
+  else
+    None ()
+ 
+-- Returns the first option if it contains a value, otherwise returns
+-- the second option.
+let optionOr: Option -> Option -> Option = lam o1. lam o2.
+  match o1 with Some _ then
+    o1
+  else
+    o2
+  
+-- Returns the option if it contains a value, otherwise calls the specified
+-- function and returns the result.
+let optionOrElse: Option -> (Unit -> Option) -> Option = lam o. lam f.
+  match o with Some _ then
+    o
+  else
+    f unit
+
+-- If exactly one option is `Some`, that option is returned,
+-- otherwise returns `None`.
+let optionXor: Option -> Option -> Option = lam o1. lam o2.
+  match o1 with Some _ then
+    match o2 with None () then
+      o1
+    else
+      None ()
+  else match o1 with None () then
+    match o2 with Some _ then
+      o2
+    else
+      None ()
+  else
+    None ()
+
+-- Converts from `Option<Option<T>>` to `Option<T>`
+let optionFlatten: Option -> Option = lam o.
+  match o with Some t then
+    t
+  else
+    None ()
+
+mexpr
+
+  utest optionIsNone (None ()) with true in
+  utest optionIsNone (Some 1) with false in
+
+  utest optionIsSome (Some 1) with true in
+  utest optionIsSome (None ()) with false in
+
+  utest optionContains (Some 1) (lam t. eqi t 1) with true in
+  utest optionContains (Some 2) (lam t. eqi t 1) with false in
+  utest optionContains (None ()) (lam t. eqi t 1) with false in
+
+  utest optionMap (None ()) (lam t. addi t 1) with (None ()) in
+  utest optionMap (Some 1) (lam t. addi t 1) with (Some 2) in
+
+  utest optionFlatMap (None ()) (lam t. Some (addi t 1)) with (None ()) in
+  utest optionFlatMap (Some 1) (lam t. Some (addi t 1)) with (Some 2) in
+  utest optionFlatMap (Some 1) (lam t. None ()) with (None ()) in
+
+  utest optionMapOr (Some 1) 3 (lam t. addi t 1) with 2 in
+  utest optionMapOr (None ()) 3 (lam t. addi t 1) with 3 in
+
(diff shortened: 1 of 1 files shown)


stdlib/option.mc
```diff
@@ -1,4 +1,74 @@
...
+let isSome: Option -> Bool = lam o.
+  match o with Some t then
+    true
+  else
+    false
+
+-- Returns `true` if the option contains the specified value.
+let contains: Option -> Dyn -> Bool = lam o. lam v.
+  match o with Some t then
+    if eqi t v then
```
segeljakt: Woops, the equality check here should not work.

EliasC: Maybe provide a predicate instead? (cf. `find` in `seq.mc`)

segeljakt:
I got these errors:

```
 ** Unit test FAILED on line 10 **
    LHS: clos p. match o'1 with Some t then p'1 t'0 else false
    RHS: true

 ** Unit test FAILED on line 11 **
    LHS: clos p. match o'1 with Some t then p'1 t'0 else false
    RHS: false

 ** Unit test FAILED on line 12 **
    LHS: clos p. match o'1 with Some t then p'1 t'0 else false
    RHS: false

 ** Unit test FAILED on line 14 **
    LHS: clos f. match o'1 with Some t then Some'7 (f'1 t'0) else None'5
    RHS: con(None,sym2)

 ** Unit test FAILED on line 15 **
    LHS: clos f. match o'1 with Some t then Some'7 (f'1 t'0) else None'5
    RHS: con(Some,sym1,2)

 ** Unit test FAILED on line 17 **
    LHS: clos d. lam f:Dyn. match o'2 with Some t then f'1 t'0 else d'1
    RHS: 2

 ** Unit test FAILED on line 18 **
    LHS: clos d. lam f:Dyn. match o'2 with Some t then f'1 t'0 else d'1
    RHS: 3

 ** Unit test FAILED on line 20 **
    LHS: clos d. match o'1 with Some t then t'0 else d'0 unit'18
    RHS: 2

 ** Unit test FAILED on line 21 **
    LHS: clos d. match o'1 with Some t then t'0 else d'0 unit'18
    RHS: 3
```

stdlib/option.mc
```diff
@@ -1,4 +1,132 @@
...
+let orElse: Option -> (Unit -> Option) -> Option = lam o. lam f.
+  match o with Some _ then
+    o
+  else
+    f unit
+
+-- Returns the first option if exactly one option is `Some`,
+-- otherwise returns `None`.
+let xor: Option -> Option -> Option = lam o1. lam o2.
+  match o1 with Some _ then
```
segeljakt, line 99: Is it possible to pattern match on tuples? I was not able to find the syntax for it. :/

stdlib/option.mc
```diff
@@ -1,4 +1,132 @@
...
+    match o2 with None then
+      o1
+    else
+      None
+  else match o1 with None then
+    match o2 with Some _ then
+      o1
+    else
+      None
+  else
```
segeljakt, line 109: This `else` branch was needed to get the code compiling, but I think it will never be executed.

stdlib/option.mc
```diff
@@ -1,4 +1,132 @@
...
+    None
+
+-- Converts from `Option<Option<T>>` to `Option<T>`
+let flatten: Option -> Option = lam o.
+  match o with Some t then
```
segeljakt: I was not able to find a way to pattern match on nested sum-types.

segeljakt:
One other thing, I was not able to find a way to pattern match on nested sum-types, e.g.

```ocaml
match o with Some Some t then
    ...
else
    ...
```

stdlib/option.mc
```diff
@@ -1,4 +1,132 @@
...
+    p t
+  else
+    false
+
+-- Applies a function to the contained value (if any).
+let map: Option -> (Dyn -> Dyn) -> Option = lam o. lam f.
+  match o with Some t then
+    Some(f t)
+  else
+    None
```
johnwikman: @EliasC Will MCore support function overloading? Just curious considering that there is a different map for sequences, and it would be nice to have access to both.

stdlib/option.mc
EliasC: You have to use tuple projection to extract stuff from a tuple.
```
let left = t.0 in
let right = t.1 in
...
```
Also, it is not possible to pattern match on two things at the same time (although my feeling is that this is something that we will want to support sooner or later).

stdlib/option.mc
EliasC: That sounds right! A `match` currently always needs an `else`, even when we know that matching is exhaustive. For now, just put `error "Not an option"` or something like this. When we have a type system, we will be able to avoid this!

EliasC:
> @segeljakt: I was not able to find a way to pattern match on nested sum-types

That is also a current limitation (that might be lifted soon, the jury is still out...). For now, you have to write it as two nested `match`es.

stdlib/option.mc
EliasC: That is a good question! My feeling is that MCore will _not_ support function overloading, but that Ragnar (or some other language built on top of MCore) will. For now, when there isn't even a module system to disambiguate conflicting names, we need to use a naming scheme that separates them, e.g., `optionMap` and `seqMap` (although I would very much prefer to let `map` be an alias of `seqMap` in that case).

There is also the possibility to use the same trick as in your PR for formatted printing (#25). Something like this (but with polymorphism rather than dynamic typing):

```
lang Functor
  syn FunctorWrapper =
  -- Intentionally left blank

  sem fmap (f : Dyn -> Dyn) =
  -- Intentionally left blank
end

lang SeqFunctor = Functor
  syn FunctorWrapper =
  | SeqFunctor [Dyn]

  sem fmap (f : Dyn -> Dyn) =
  | SeqFunctor s -> seqMap f s
end
```

```
utest fmap (addi 1) (SeqFunctor([1,2,3])) with [2,3,4] in
...
```
We should however agree on this before we start reimplementing the entire standard library like this.

EliasC:
> @segeljakt: I got these errors

An error where you get `clos p. ...` instead of some value you were expecting is a sign that some function (**clos**ure) was not provided with enough arguments. In particular, a function was not provided an argument for the parameter called `p`. I will have a look at your test cases and see what I can find!

stdlib/option.mc
```diff
@@ -1,4 +1,132 @@
...
+
+  utest isNone(None) with true in
+  utest isNone(Some(1)) with false in
+
+  utest isSome(Some(1)) with true in
+  utest isSome(None) with false in
+
+  utest contains(Some(1), 1) with true in
+  utest contains(Some(2), 1) with false in
+  utest contains(None, 1) with false in
```
EliasC: All of these function calls are missing the predicate `p`. Also, function calls are _curried_ in MCore, meaning that you pass them one by one rather than as a single tuple. What you want is probably something like this:
```
utest contains (Some 1) (lam n. eqi n 1) with true in
...
```

stdlib/option.mc
```diff
@@ -1,4 +1,132 @@
...
+
+  utest map(None, 1) with None in
+  utest map(Some(1), lam v. addi v 1) with Some(2) in
```
EliasC: The first call to `map` is missing a function argument (`1` is not a function). Because we don't have a type system and the function is not applied in the `None` case, this won't lead to an error, but it should be fixed nonetheless. Also, arguments should be in curried form, not as a tuple.

segeljakt:
Thank you for the feedback 😄

stdlib/option.mc
segeljakt: oops.. I overlooked the part in the README that said
>As part of the design of MExpr, we try to make it simple. Hence, MExpr does not support pattern variables or nested matches.

stdlib/option.mc
david-broman: MCore will most likely not be supporting overloading. As I see it, it should be implemented on a language on top of MCore. Ragnar (which should be the language we use for implementing the compiler in the end) should of course support it in a controlled way (most likely some variant of type classes).

segeljakt:
I did not port the following methods from https://doc.rust-lang.org/std/option/enum.Option.html listed below.

These required references `&`:
* `as_ref`
* `as_mut`
* `as_pin_ref`
* `as_pin_mut`
* `get_or_insert`
* `get_or_insert_with`
* `take`
* `replace`
* `copied`
* `cloned`
* `as_deref`
* `as_deref_mut`

These required `panic`s.
* `expect`
* `unwrap`
* `unwrap_or`
* `unwrap_or_else`
* `expect_none`
* `unwrap_none`

These required `Result` types:
* `ok_or`
* `ok_or_else`
* `transpose`

These required `Iterator` type classes:
* `iter`
* `iter_mut`

These required `Default` constructor type classes:
* `unwrap_or_default`

PR #22, opened 2019-11-13
Make signature of f in foldr1 consistent with signature in foldr

br4sco:


EliasC:
Good call! Merging...

PR #23, opened 2019-11-13
Add float2string and support for negative numbers in int2string and string2int

johnwikman:
Adds naive float2string conversion, sufficient for printing results.

Also adds negative integer support for int2string and string2int conversion.


EliasC:
Good stuff! Didn't check the details of `float2string`, but I think it should do for now.

PR #25, opened 2019-11-13
Add a simple string formatter written in MLang

johnwikman:
A simple (but verbose) implementation of string formatting in MLang. Put this in `test/mlang` instead of `stdlib` since I am not 100% about the implementation specifics but still want it "for the record".

### DIFF
--- a/test/mlang/strformat.mc
+++ b/test/mlang/strformat.mc
@@ -4,78 +4,77 @@ include "char.mc"
 include "seq.mc"
 include "string.mc"
 
-lang TypeInteger
-	syn Type =
-	| TyInt (Int)
+lang FmtInteger
+	syn Fmt =
+	| FmtInt (Int)
 
 	sem toString =
-	| TyInt v -> int2string v
+	| FmtInt n -> int2string n
 end
 
-lang TypeFloat
-	syn Type =
-	| TyFloat (Float)
+lang FmtFloat
+	syn Fmt =
+	| FmtFloat (Float)
 
 	sem toString =
-	| TyFloat v -> float2string v
+	| FmtFloat f -> float2string f
 end
 
-lang TypeString
-	syn Type =
-	| TyStr (String)
+lang FmtString
+	syn Fmt =
+	| FmtStr (String)
 
 	sem toString =
-	| TyStr v -> v
+	| FmtStr s -> s
 end
 
-lang TypeChar
-	syn Type =
-	| TyChar (Char)
+lang FmtChar
+	syn Fmt =
+	| FmtChar (Char)
 
 	sem toString =
-	| TyChar v -> [v]
+	| FmtChar c -> [c]
 end
 
 lang StrFormatBase
-	syn Type =
-
-	syn Operand =
-	| CStrFormat (String)
+	syn Fmt =
+	-- Intentionally left blank
 
 	sem toString =
-	| _ -> error "StrFormatBase: toString: Unknown type"
+	| _ -> error "StrFormatBase: toString: Unknown Fmt"
 
-	sem eval (args : [Type]) =
-	| CStrFormat s ->
+	sem strFormat (args : [Fmt]) =
+	| s ->
 		if lti (length s) 2 then
-			s 
+			s
 		else if eqchar '%' (head s) then
 			let c = head (tail s) in
 			if eqchar '%' c then
-				cons '%' (eval args (CStrFormat (tail (tail s))))
+				cons '%' (strFormat args (tail (tail s)))
 			else if is_alpha c then
 				-- At the moment just accept any alpha char to represent a format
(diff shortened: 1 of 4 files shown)


test/mlang/strformat.mc
```diff
@@ -0,0 +1,81 @@
...
+use StrFormat in
+let sprintf = lam s. lam args. eval args (CStrFormat (s)) in
+let printf = lam s. lam args. print (sprintf s args) in
+
+utest sprintf "%d + %d = %d" [TyInt(2), TyInt(3), TyInt(addi 2 3)] with "2 + 3 = 5" in
+utest sprintf "Give it %T%%" [TyInt(101)] with "Give it 101%" in
+utest sprintf "Hello, %s!" [TyStr("John Doe")] with "Hello, John Doe!" in
+utest sprintf "My initials are %c.%c." [TyChar('J'), TyChar('D')] with "My initials are J.D." in
+
+printf "\n >Test Print:\n >%a/%a = %a\n" [TyInt(10), TyInt(3), TyFloat(divf (int2float 10) (int2float 3))]
```
EliasC: I would avoid putting prints in a test case as it will clutter output when running the full test suite (`make test`).

test/mlang/strformat.mc
```diff
@@ -0,0 +1,81 @@
...
+
+	sem toString =
+	| TyChar v -> [v]
+end
+
+lang StrFormatBase
+	syn Type =
+
+	syn Operand =
+	| CStrFormat (String)
```
EliasC: Is this datatype needed? Could the pattern of `eval` (is this the right name?) be a catch-all named `s` which is just a string?

test/mlang/strformat.mc
```diff
@@ -0,0 +1,81 @@
...
+include "seq.mc"
+include "string.mc"
+
+lang TypeInteger
+	syn Type =
+	| TyInt (Int)
+
+	sem toString =
+	| TyInt v -> int2string v
+end
```
EliasC, line 20: I'm wondering if you should choose names that is more specific for this particular library. The name `Type` is very general, and without looking at the entire library it isn't obvious why something denoting a type should also carry with it a value of that type. How about changing `Type` to `Fmt`:

```
lang FmtInteger
  syn Fmt =
  | FmtInt Int

  sem toString =
  | FmtInt n -> int2string n
end
```

EliasC (commented):
I think this is good as a first stab at the problem. I have added some comments and suggestions. Making the library extensible in the actual format codes (i.e. `%s`, `%d`, etc.) is still an open problem. I have some ideas, but I don't think they are possible in the current system. Of course, it might be the case that the right design is to have a fixed number of format codes and just use `toString` to print arbitrary data.

test/mlang/strformat.mc
johnwikman: The though process here was more that all types probably should have some form of toString and that they should be easy to combine with the StrFormatBase language. Changed it to Fmt now.

johnwikman:
I addressed the specific issues you mentioned in the latest commit. Regarding the format codes I have some that I want to try out, hence I changed to PR to WIP.

EliasC:
Great! Not merging yet then...

johnwikman:
Added specific format strings for different types (and `%*` to represent `toString` for all types).

This is close to what I initially imagined, but it would also be nice if there was a way to add on additional format strings to existing types without having to rewrite the original language. For example to add `"%.3f"` to `FmtFloat` while keeping the same formats from `FormatFloat`.

johnwikman:
To add to my last comment: The feature that I am looking for would be something similar to this:
```
-- Original Implementation
lang FormatString
	syn Fmt =
	| FmtStr (String)

	sem toString =
	| FmtStr s -> s

	sem toFormat (fmtstr : String) =
	| FmtStr s where (eqstr fmtstr "%s") -> s
	| _ -> error (concat "FormatString: toFormat: Unrecognized format: " fmtstr)
end

...

-- My own extension with additional formats
lang MyFormatString = FormatString
	sem toFormat (fmtstr : String) =
	| FmtStr s where (eqstr fmtstr "%^s") -> str2upper s
	| FmtStr s where (eqstr fmtstr "%_s") -> str2lower s
end

lang MyStrFormat = MyFormatString + StrFormat
```

Some form of matching on both syntax as well as an optional conditional would aid the modularity in this case. While the case above could be solved with fixed case syntax in MLang, more complex format arguments in the case of floats could require some form of regex matching.

EliasC:
@johnwikman I think this could be merged more or less as is! Could you translate your recursive MExpr functions to the new `recursive` style (while resolving the conflicts), and change the layout to follow the prescribed style (especially "scope indentation should be 2 spaces")? I will translate some of the insights from this discussion to an issue so it doesn't get lost.

johnwikman:
@EliasC done, should be able to merge now.

PR #26, opened 2019-11-15
Rename 'id' to 'identity'

elegios:
Free up a rather useful name for users


EliasC:
Good point! Merging...

PR #27, opened 2019-11-21
Recursive lets

elegios:
Adds full mutual recursion to the OCaml interpreter (including the top-level), as well as the interpreter in `mexpr.mc`. It also adds a few constants to `mexpr.mc`, since they were needed for the test case.

### DIFF
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -124,7 +124,7 @@ and program = Program of include_ list * top list * tm
 and tm =
 | TmVar     of info * ustring * int                                 (* Variable *)
 | TmLam     of info * ustring * ty * tm                             (* Lambda abstraction *)
-| TmClos    of info * ustring * ty * tm * env                       (* Closure *)
+| TmClos    of info * ustring * ty * tm * env Lazy.t                (* Closure *)
 | TmLet     of info * ustring * tm * tm                             (* Let *)
 | TmRecLets of info * (info * ustring * tm) list * tm               (* Recursive lets *)
 | TmApp     of info * tm * tm                                       (* Application *)
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -111,10 +111,12 @@ and decl = (* TODO: Local? *)
 
 and mlang   = Lang of info * ustring * ustring list * decl list
 and let_decl = Let of info * ustring * tm
+and rec_let_decl = RecLet of info * (info * ustring * tm) list
 and con_decl = Con of info * ustring * ty
 and top =
 | TopLang of mlang
 | TopLet  of let_decl
+| TopRecLet of rec_let_decl
 | TopCon of con_decl
 
 and include_ = Include of info * ustring
--- a/src/boot/lexer.mll
+++ b/src/boot/lexer.mll
@@ -30,7 +30,6 @@ let reserved_strings = [
   ("let",           fun(i) -> Parser.LET{i=i;v=()});
   ("recursive",     fun(i) -> Parser.REC{i=i;v=()});
   ("lam",           fun(i) -> Parser.LAM{i=i;v=()});
-  ("fix",           fun(i) -> Parser.FIX{i=i;v=()});
   ("in",            fun(i) -> Parser.IN{i=i;v=()});
   ("end",           fun(i) -> Parser.END{i=i;v=()});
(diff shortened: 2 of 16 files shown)


src/boot/ast.ml
```diff
@@ -124,7 +126,7 @@ and program = Program of include_ list * top list * tm
 and tm =
 | TmVar     of info * ustring * int                                 (* Variable *)
 | TmLam     of info * ustring * ty * tm                             (* Lambda abstraction *)
-| TmClos    of info * ustring * ty * tm * env                       (* Closure *)
+| TmClos    of info * ustring * ty * tm * env Lazy.t                (* Closure *)
```
EliasC, line 129: Could you say something about why `Lazy` is needed here (in the PR in general)? I'm guessing it has something to do with preventing diverting fixpoint computations?

src/boot/lexer.mll
```diff
@@ -30,7 +30,6 @@ let reserved_strings = [
   ("let",           fun(i) -> Parser.LET{i=i;v=()});
   ("recursive",     fun(i) -> Parser.REC{i=i;v=()});
   ("lam",           fun(i) -> Parser.LAM{i=i;v=()});
-  ("fix",           fun(i) -> Parser.FIX{i=i;v=()});
```
EliasC, line 33: We should also remember to remove `fix` from the MCore parser.

test/mlang/mexpr.mc
```diff
@@ -99,6 +116,35 @@ lang LetEval = Let + VarEval
...
+
+lang RecLetsEval = RecLets + VarEval + Fix + FixEval
+  sem eval (env : Env) =
+  | TmRecLets t ->
+    let bindings = t.0 in
+    let body = t.1 in
+    let foldli = lam f. lam init. lam seq.
+      (foldl (lam acc. lam x. (addi acc.0 1, f acc.0 acc.1 x)) (0, init) seq).1 in
+    utest foldli (lam i. lam acc. lam x. concat (concat acc (int2string i)) x) "" ["a", "b", "c"]
+      with "0a1b2c" in
```
EliasC, line 132: Is this meant as documentation about what `foldli` does, or was it left by accident?

EliasC: (also, does it make sense to put in the standard library?)

EliasC (approved):
I think this looks good! Let's resolve the comments and get it merged.

src/boot/ast.ml
elegios: Conceptually, each mutually recursive lambda should become a closure with a paired environment. That environment should include the closures for all of the mutually recursive lambdas, i.e., the environment needs to be cyclic. This can be achieved in two ways: with a `fix` that creates one nesting of the cycle at a time, or an actual recursive value. OCaml allows cyclic values, but it can only handle it in very simple circumstances, e.g., `let rec x = 1 :: x`. Suffice to say that this isn't quite simple enough to support that directly, so we insert a layer of laziness, which makes it simple again.

src/boot/lexer.mll
elegios: I quite forgot to look at that. Is it just changes in `meta.mc`? What's praxis for testing it?

test/mlang/mexpr.mc
elegios: I wrote it as a generic helper, so it felt sensible to test it as one. Not sure about putting it in the standard library. It doesn't tend to be a part of standard libraries for other functional languages, to my knowledge anyway. I don't really have any preference either way.

EliasC: I actually liked having the test there, since it helped me understand what it did (and it has negligible performance overhead, at least in theory). Maybe it's too early to introduce the `listExtras.mc` library ;)

src/boot/lexer.mll
EliasC: Until we've sorted out having multiple `use`s of the same language, you should change it in `meta.mc` and `mcore_parser.mc`. The latter you can test by feeding it the test files in `test/mexpr` (this is currently not done automatically).

elegios: I've now pushed updates that make `meta.mc` and `mcore_parser.mc` both use recursive lets and parse programs that contain them.

PR #28, opened 2019-11-21
Slightly more friendly output in error messages

elegios:
Some changes I made for easier debugging while working on #27. Most things should be fairly uncontroversial, except for changes in `pprintME`. I changed printing of constructors from
```
con(TmLam, sym74, (...))
```
to
```
TmLam_74 (...)
```
since it was easier to figure out a good way to pretty-print then.

### DIFF
--- a/src/boot/mexpr.ml
+++ b/src/boot/mexpr.ml
@@ -111,7 +111,10 @@ let arity = function
 
 
 
-let fail_constapp fi = raise_error fi "Incorrect application "
+let fail_constapp f v fi = raise_error fi ("Incorrect application. function: "
+                                         ^ Ustring.to_utf8 (pprint_const f)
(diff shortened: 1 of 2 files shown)


src/boot/mexpr.ml
```diff
@@ -120,6 +123,7 @@ let fail_constapp fi = raise_error fi "Incorrect application "
    The reason for this is that if-expressions return expressions
    and not values. *)
 let delta fi c v  =
+    let fail_constapp = fail_constapp c v in
```
EliasC, line 126: Some would object to this shadowing. I don’t mind that much.

EliasC (approved):
This is very welcome! I for one also think that the proposed printing of algebraic data types is the right one.

PR #29, opened 2019-11-21
Remember generated symbols

EliasC:
Whenever a new constructor is defined, instead of blindly
generating a new symbol, this commit makes it so that the symbol
for each combination of constructor and datatype is remembered and
reused. This means that two datatypes with the same names are
assumed to be "the same".

Two obvious TODOs:
- Names (and thereby, symbols) should not be resolved during
runtime, but via static analysis (i.e., typechecking).
- We currently ignore parameters of datatypes and look only at the
rightmost type in a chain of arrows.


elegios:
Some initial thoughts:

I assume that this is done since `use` re-runs `con` expressions (I don't know that this is the case, so please correct me if I'm wrong)?

Following on that I suspect that the more long term way to fix this is to introduce namespaces and have `use` utilize that mechanism.

As a user I would probably expect it to be impossible to define two constructors in the same scope. As someone that generates code I'd maybe like it to shadow a previous definition (but probably be a distinct constructor).

As a final note, the memoization makes some assumptions about the type system:
- A constructor can have arbitrary arity, but the implementation currently has 0-arity and 1-arity (i.e., a constructor can be "not yet applied" or "applied to a single argument", and both cases can be matched against, regardless of the declared type of the constructor).
- Each constructor declaration belongs to exactly one type, specified as the last type in an arrow-chain. I personally like the idea of either ad-hoc sum types or row-type sums, which runs a little bit counter to this since it allows a single constructor to be a part of many types.

EliasC:
> I assume that this is done since use re-runs con expressions (I don't know that this is the case, so please correct me if I'm wrong)?

That is correct. This means that the following test will currently fail
```
lang L
  syn S =
  | Foo
end

mexpr

let x = use L in Foo in
let y = use L in Foo in
utest x with y in
()
```

> Following on that I suspect that the more long term way to fix this is to introduce namespaces and have use utilize that mechanism.

Indeed. Name resolution should not be dynamic when it can be done in a static fashion.

> As a user I would probably expect it to be impossible to define two constructors *[with the same name]* in the same scope. As someone that generates code I'd maybe like it to shadow a previous definition (but probably be a distinct constructor).

It should most likely be disallowed if they belong to the same type (and it's probably a bad idea even if they don't). 

> A constructor can have arbitrary arity, but the implementation currently has 0-arity and 1-arity (i.e., a constructor can be "not yet applied" or "applied to a single argument", and both cases can be matched against, regardless of the declared type of the constructor).

Do you mean that `gencon` should check for constructors with arity higher than 1 (and raise an error)? I think it makes sense to use an OCaml-style "zero or one tuple argument" from a simplicity-standpoint, and higher-arity constructors (and implicit currying) can be a feature of some higher-level language.

> Each constructor declaration belongs to exactly one type, specified as the last type in an arrow-chain. I personally like the idea of either ad-hoc sum types or row-type sums, which runs a little bit counter to this since it allows a single constructor to be a part of many types.

I don't know about this... I might be thinking very much from an MLang perspective rather than an MExpr one, but I think a nice feature of MLang is that you always have the type of a constructor (as well as its entire interface). When you want to add new stuff, you do it in such a way that you know that you are complete (with respect to interpreters) and backwards-compatible with the existing language fragment. Being able to mix-and-match constructors across types might be powerful, but also seems quite complicated from a user-perspective (and possibly difficult to compile to efficient code, but I don't know this for sure). I have to think a bit more about this...

PR #30, opened 2019-12-02
Nested Patterns in MExpr

elegios:
This adds nested patterns to the OCaml interpreter and the `mexpr.mc` interpreter. Notably absent is support for nested patterns in `sem` declarations in MLang, but those have some remaining semantic questions (ordering of patterns in composition) and so won't be a part of this PR.

I also haven't yet finished updating `mcore_parser.mc` and `meta.mc`, but I wanted to get the rest looked at sooner rather than later.

As a usability note, it feels surprising when `match e with Con then ...` highlights `Con` as a constructor/type, but the semantics are equivalent with `let Con = e in ...`. It's also annoying to write `None ()` everywhere, but that's expected. As a sidenote, I believe the current implementation allows us to write
```
con None : () -> Option
let None = None ()

let map = lam f. lam o.
  match o with Some x then Some (f x) else
  match o with None _ then None else error "impossible"
```

Note that while we can't get away from using `None _` as the pattern, we can still use `None` as a value.

This gets somewhat close to the inconsistency noted in `mexpr.mc`:

```
con Foo in
let Bar = Foo in
match Foo () with Bar _ then true else false
```

This evaluates to `true` in `mexpr.ml`, but I believe it evaluates to `false` in `mexpr.mc` (both before and after this pull request) (I haven't quite had the time to construct the ast to test it yet).

### DIFF
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -138,10 +138,18 @@ and tm =
 | TmProj    of info * tm * int                                      (* Projection of tuple *)
 | TmCondef  of info * ustring * ty * tm                             (* Constructor definition *)
 | TmConsym  of info * ustring * sym * tm option                     (* Constructor symbol *)
-| TmMatch   of info * tm * ustring * int * ustring option * tm * tm (* Match data *)
+| TmMatch   of info * tm * pat * tm * tm                            (* Match data *)
 | TmUse     of info * ustring * tm                                  (* Use a language *)
 | TmUtest   of info * tm * tm * tm                                  (* Unit testing *)
 
+(* Patterns *)
+and pat =
+| PatNamed of info * ustring                      (* Named, capturing wildcard *)
+| PatTuple of info * pat list                     (* Tuple pattern *)
+| PatCon   of info * ustring * int * pat option   (* Constructor pattern *)
+| PatInt   of info * int                          (* Int pattern *)
+| PatBool  of info * bool                         (* Boolean pattern *)
+| PatUnit  of info                                (* Unit pattern *)
 
 (* Types *)
 and ty =
@@ -173,9 +181,12 @@ module Option = struct
   let map f = function
     | Some x -> Some (f x)
     | None -> None
+  let bind f = function
+    | Some x -> f x
+    | None -> None
 end
 
-(* General map over terms *)
+(* General (bottom-up) map over terms *)
 let rec map_tm f = function
   | TmVar (_,_,_) as t -> f t
   | TmLam(fi,x,ty,t1) -> f (TmLam(fi,x,ty,map_tm f t1))
@@ -192,8 +203,8 @@ let rec map_tm f = function
   | TmProj(fi,t1,n) -> f (TmProj(fi,map_tm f t1,n))
   | TmCondef(fi,x,ty,t1) -> f (TmCondef(fi,x,ty,map_tm f t1))
   | TmConsym(fi,k,s,ot) -> f (TmConsym(fi,k,s,Option.map (map_tm f) ot))
-  | TmMatch(fi,t1,k,n,x,t2,t3) ->
-    f (TmMatch(fi,map_tm f t1,k,n,x,map_tm f t2,map_tm f t3))
+  | TmMatch(fi,t1,p,t2,t3) ->
+    f (TmMatch(fi,map_tm f t1,p,map_tm f t2,map_tm f t3))
   | TmUse(fi,l,t1) -> f (TmUse(fi,l,map_tm f t1))
   | TmUtest(fi,t1,t2,tnext) -> f (TmUtest(fi,map_tm f t1,map_tm f t2,map_tm f tnext))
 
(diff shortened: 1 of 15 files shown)


EliasC:
> it feels surprising when `match e with Con then ...` highlights `Con` as a constructor/type.

It doesn't exactly help to avoid that error... I will have a look later to see if we can adress that in the Emacs mode.

> `let None = None ()`

I would consider this an anti-pattern.

> [regarding the] the inconsistency noted in `mexpr.mc`

I feel like that program shouldn't compile at all, which lets us avoid making the choice :)

src/boot/ast.ml
```diff
@@ -138,10 +138,18 @@ and tm =
...
 | TmUtest   of info * tm * tm * tm                                  (* Unit testing *)
 
+(* Patterns *)
+and pat =
+| PatNamed of info * ustring                      (* Named, capturing wildcard *)
+| PatTuple of info * pat list                     (* Tuple pattern *)
+| PatCon   of info * ustring * int * pat          (* Constructor pattern *)
+| PatInt   of info * int                          (* Int pattern *)
+| PatBool  of info * bool                         (* Boolean pattern *)
+| PatUnit  of info                                (* Unit pattern *)
```
EliasC, line 153: Should this also include character literals?

src/boot/ast.ml
```diff
@@ -173,9 +181,15 @@ module Option = struct
   let map f = function
     | Some x -> Some (f x)
     | None -> None
+  let bind f = function
+    | Some x -> f x
+    | None -> None
+  let value o ~default = match o with
+    | Some x -> x
+    | None -> default
```
EliasC: Is `value` ever used?

src/boot/mexpr.ml
```diff
@@ -393,7 +393,20 @@ let rec debruijn env t =
   let rec find fi env n x =
     (match env with
      | VarTm(y)::ee -> if y =. x then n else find fi ee (n+1) x
-     | [] -> raise_error fi ("Unknown variable '" ^ Ustring.to_utf8 x ^ "'"))
+     | [] -> raise_error fi ("Unknown variable '" ^ Ustring.to_utf8 x ^ "'")) in
+  let rec dbPat env = function
+    | PatNamed(fi,x) -> (VarTm(x)::env, PatNamed(fi,x))
+    | PatTuple(fi,ps) -> (* NOTE: this causes patterns to introduce names right-to-left, which will cause errors if a later pattern binds a name that is seen as a constructor in a pattern to the left *)
```
EliasC, line 399: This should also be solved by type checking, right?

stdlib/char.mc
```diff
@@ -1,3 +1,5 @@
+include "seq.mc"
```
EliasC, line 1: Why is this needed?

test/mexpr/nestedpatterns.mc
```diff
@@ -0,0 +1,25 @@
+mexpr
+
+let classify = lam x.
+  match x with (true, true) then "one" else
+  match x with (true, false) then "two" else
+  match x with (false, true) then "three" else
+  match x with (false, false) then "four"
+  else "five" in
```
EliasC, line 8: Your interpretation of binary numbers is flawed ;)

test/mlang/mexpr.mc
```diff
@@ -216,25 +235,36 @@ end
...
 end
 
+lang BoolPat = Bool
+  syn Pat =
+  | BoolPat Bool
+
+  sem tryMatch (t : Expr) =
+  | BoolPat b ->
+    match t with TmConst CBool b2 then
+      if or (and b b2) (and (not b) (not b2)) then Some [] else None () -- TODO: is there a nicer way to do equality on bools? 'eqb' is unbound
```
EliasC, line 271: We should probably add an `eqb`!

EliasC (approved):
I think this looks good! Do you want this merged as is (after rebasing and possible fixes) or are you planning on adding any features?

src/boot/ast.ml
elegios: Definitely

src/boot/ast.ml
elegios: It used to be, but no longer, I'll remove it

stdlib/char.mc
elegios: A use of `head`. It works when `char.mc` is included somewhere else, since `prelude` is already included at that point, but when running tests on the standard library I run them without `prelude`, since it would induce an import cycle otherwise (well, only on the files that are included in `prelude`, but still, that's annoying to detect for the test runner).

test/mexpr/nestedpatterns.mc
elegios: :)

src/boot/mexpr.ml
elegios: Mostly yes, but it will at least be solved by #31.

It does still leave shadowing for normal names though, so 

```
match e with (a, a) then a else ...
```

will use the left value, since it shadows the right. This may or may not be expected.

EliasC: My feeling is that shadowing in the same pattern is also something that a typechecker should prevent.

PR #33, opened 2019-12-18
RFC: Support for Records

EliasC:
This RFC explains the ongoing work to add records to MCore.

A record is an unordered map from labels (strings) to values:
```
let r1 = {age = 42, name = "foobar"} in
let r2 = {name = "foobar", age = 42} in
utest r1 with r2 in
...
```
Like with tuples, values are projected from a record by specifying their labels:
```
utest r1.age with 42 in
...
```
Unlike with tuples, new records can be created from existing tuples with individual tuples changed:
```
let r3 = {r2 with age = 43} in
...
```
The reason for this is to be able to support functions where the full record is not known (e.g. row polymorphism). For example, we might write a function that updates a single field but leaves the rest untouched, regardless of what these fields are:
```
let bumpAge = lam r : {age: Int, ..}. {r with age = r.age + 1} in
utest bumpAge r2 with r3 in
()
```
Note that this is not the same thing as subtyping, where the return type of `bumpAge` would necessarily be the tuple type `{age : Int}`, as there is no way to reason about the fields of a record that cannot be seen.

The details of row-polymorphism are being worked out and will be posted here when ready for scrutiny.

### DIFF
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -24,10 +24,11 @@ let utest_ok = ref 0            (* Counts the number of successful unit tests *)
 let utest_fail = ref 0          (* Counts the number of failed unit tests *)
 let utest_fail_local = ref 0    (* Counts local failed tests for one file *)
 
+(* Map type for record implementation *)
+module Record = Map.Make(Ustring)
 
 (* Evaluation environment *)
 type env = tm list
-
 
 and const =
 (* MCore intrinsic: unit - no operation *)
@@ -86,6 +87,8 @@ and const =
 | Ccons    of tm option
 | Cslice   of (tm list) option * int option
 | Creverse
+(* MCore intrinsic: records *)
+| CRecord of tm Record.t
 (* MCore debug and I/O intrinsics *)
 | Cprint
 | Cdprint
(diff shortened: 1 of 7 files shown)


johnwikman:
I have not been present in the recent discussions, but is there a reason for why this functionality should be implemented in MCore? It feels quite high-level.

EliasC:
> I have not been present in the recent discussions, but is there a reason for why this functionality should be implemented in MCore?

Good question! The reason for this is that records (with row-polymorphism) are needed for extending the payload of a constructor of a data type:

```
lang LambdaSyntax
  syn Term =
  | TmVar String
  | TmAbs {var: String, body: Term, ..} -- Note the "..", meaning "and more"
  | TmApp (Term, Term)
end

lang TypedLambdaSyntax = LambdaSyntax
  syn Type =
  | TyArrow (Type, Type)

  syn Term =
  | TmAbs {var: String, body: Term, type: Type, ..}
end
```

We cannot do the same thing by just adding another field to a tuple; since tuples are ordered, we would be in trouble if two different languages extend the same tuple with different fields. With records, there is no ambiguity regarding "where" in a record the value of a field is.

We also cannot do without the `..` above, since that would allow us to write the term `TmAbs {var = "x", body = TmVar "x"}` in an interpreter in `LambdaSyntax`, which would not be compatible with the expected type in `TypedLambdaSyntax` (since the `type` field is missing). With the `..` syntax, we can only create a new `TmAbs` by changing fields in the record carried by an existing `TmAbs` (which will then, possibly implicitly, carry all the fields needed by all extensions). Note that when we `use` a language, we "close" the terms for extension, and the constructors behave as if the `..` was not there.

PR #34, opened 2020-01-20
Add make install for local installation

dlunde:


johnwikman:
Isn't it more appropriate to use `$HOME/.local/bin` as specified in the [freedesktop standard](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html#Home%20Directory)?

dlunde:
That's fine with me!

johnwikman:
Just a quick thought, could it also be an idea to also install the standard library into `$HOME/.local/lib/mcore/stdlib` and in [boot.ml](https://github.com/miking-lang/miking/blob/develop/src/boot/boot.ml#L31) try to use this directory if `MCORE_STDLIB` is not defined?

dlunde:
That sounds like a good default, I'll add it when I have time.

br4sco:
Wouldn't it be better to omit the interactive flag `-i` from from `cp`? It messes up Emacs compilation mode for example and other tools might have similar issues.

dlunde:
Then we would need to use `-f` instead in order to overwrite old installed versions, which feels somewhat insecure. But I guess this is fine, at least for now when we are working with this internally. Is overwriting old stuff the usual approach when running make install in general?

dlunde:
I've now added your suggestions (install stdlib and remove -i flag)

johnwikman:
LGTM. Works fine on my computer.

PR #36, opened 2020-02-03
Add pretty printer for MExpr AST

johnwikman:
A pretty printer for the MExpr AST nodes. This is still work in progress as a lot of the code is based on the MLang parser from November/December last year.

Uncomment the three print statements at the bottom of `stdlib/mexpr/pprint.mc` to see the result of the pretty printer.


david-broman:
When I run it, I get the error 

FILE "mexpr/pprint.mc" 219:16-219:32 ERROR: Unknown variable 'getPatStringCode'

Do you get the same problem John?

johnwikman:
It should work now.

Issue #32, opened 2019-12-06
Rearranging MExpr implemented in MLang


david-broman:
I have moved the files and split them into `mexpr/eval.mc`,  `mexpr/ast.mc` , and `mexpr/mexpr.mc` (in the `stdlib`). I have also cleaned up and removed old project files (`preboot`) and tests associated to preboot.

 I would like to do some more rearrangements:

- Move semantic functions `tryMatch` to `eval.mc`. 

- Remove fragments such as `UnitPat` and move the `tryMatch` function to `UnitEval`, and move the `syn Pat`, in this case `UnitPat` to the syntax fragment for unit, in this case fragment `Unit`.

- Rename pattern constructors to `PatUnit`, `PatInt` etc. instead of `UnitPat`. By doing so, we have the same formatting as in the other constructors, such as `TmApp`. This is also how we have done it in the OCaml impl. That is, to always have the kind of constructor as the prefix.

- Following the previous point, I would suggest to change to `TyUnit` from `UnitType` in the same way. Right now, we have a mix.

- Right now we have `FunEval` for ast `Fun`, but `ArithEval` for `Int`. I would suggest that we use the same pattern here, i.e. `IntEval` for `Int`.

- I am also thinking of going back to what we discussed before, i.e., to call the syntax fragments `IntAst` instead of just `Int`. I think it makes even more sense now when we have separated it into a specific file that is called `ast.mc`. 

This RFC is specifically target to @elegios and @EliasC since you have been working on this, but everyone is of course welcome to comment. I would appreciate quick answers, so that I can proceed.

elegios:
Seems reasonable to me.

EliasC:
In general I agree. Some comments:

> Move semantic functions tryMatch to eval.ml.

~Do you mean `mexpr/eval.mc`? Or are you rearranging stuff in the OCaml implementation as well?~
**saw the edit now**

> Move [pattern syntax] to the syntax fragment [...]

I can see a strength in being able to reason about (e.g.) the syntax of integers without automatically introducing integers as patterns. On the one hand, this is what we want for MCore. On the other hand, this prevents reusing that fragment in other languages that might have more limited pattern matching.

> I am also thinking of going back to what we discussed before, i.e., to call the syntax fragments `IntAst` instead of just `Int`

I agree that this makes sense!

david-broman:
@elegios OK, good.
@EliasC OK, good point with pattern syntax. I will keep `IntPat` for now.

[...]

PR #39, opened 2020-02-05
Separate AST specific and Eval specific code

### LINKED ISSUE #37: AST is still fairly Eval-influenced
Syntax in [ast.mc](https://github.com/miking-lang/miking/blob/develop/stdlib/mexpr/ast.mc) such as [CAddi2](https://github.com/miking-lang/miking/blob/develop/stdlib/mexpr/ast.mc#L81), [CSubi2](https://github.com/miking-lang/miking/blob/develop/stdlib/mexpr/ast.mc#L83), and [CMuli2](https://github.com/miking-lang/miking/blob/develop/stdlib/mexpr/ast.mc#L85) expect are defined to have an `Int` type as their parameter. When compiling, this will most likely be a non-evaluated expression instead of a direct MCore type. The `C<op>2` statements should probably be moved out to the Eval-files as they will most likely not be used anywhere else. It would also be nice to have the names `CAddi2`, `CSubi2`, etc. in the compilation process, but then be defined as `C<op>2 Expr` instead of having a fixed MCore type.

johnwikman:
The [comment by Viktor](#issuecomment-582445800) explains what was initially being changed. Since then some more changes have or will be incorporated into the PR:
- Replace all uses of tuples in the AST nodes with records. This is to allow for easier inclusion metadata such as line number and columns for each AST node.
- Add `TmRecord` into the AST. (The removal of `TmTuple` is left to a future PR.)

Closes #37


johnwikman:
@elegios

elegios:
As a summary, this moves:

- partially applied `Const` functions (e.g. `CAddi2`, `CSubi2`) to `eval.mc` since they only appear at runtime.
- `TmConFun` to `ast.mc` since they represent unapplied constructors and will thus appear before runtime.
- `tryMatch` semantic function to `eval.mc` since it is only relevant at runtime, similarly to `apply`.

Edit: I now realize that this is essentially taking some of the steps suggested in #32

PR #43, opened 2020-02-13
Replace the environment with a context and added Float arithmetic

br4sco:
Changed the semantic function `eval` to pass around a context containing the environment rather than the environment by itself. I also had to change the semantic function `apply` to pass the context since it calls `eval`.


david-broman:
Oscar, can you fix the conflicts (a result now after other merges).

br4sco:
Fixed

david-broman:
Seems to still be a merge conflict. Please make sure that rebase works. Thanks!

PR #46, opened 2020-02-20
Add lambda lifting in MCore

johnwikman:
Adds lambda lifting semantics to the MCore AST. Due to the lack of a type checker, partially applied functions are not handled.

Recreated from #44.

### DIFF
--- a/stdlib/mexpr/lamlift.mc
+++ b/stdlib/mexpr/lamlift.mc
@@ -1,0 +1,898 @@
+// Defines semantics for lambda lifting
+// Based on the technique from the 1985 paper.
+// (This will not handle partial application until a type-checker has been
+// implemented.)
+
+// This defines the lamlift semantics which, given a state tuple, propagates
+// the entire AST and lifts out any internal lambda expressions to the top-
+// level.
+
+// This also defines the internal replace semantics which replaces all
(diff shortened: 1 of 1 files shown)


david-broman:
Testerna går inte igenom. 

mexpr/lamlift.mc: 
FILE "mexpr/eval.mc" 93:20-94:67 ERROR: Cannot update the term. The term is not a record: [("addi",ConstAst_TmConst_194 {val = ArithIntAst_CAddi_199 ()}),("subi",ConstAst_TmConst_194 {val = ArithIntAst_CSubi_200 ()}),("and",ConstAst_TmConst_194 {val = BoolAst_CAnd_210 ()}),("or",ConstAst_TmConst_194 {val = BoolAst_COr_211 ()}),("not",ConstAst_TmConst_194 {val = BoolAst_CNot_209 ()}),("eqi",ConstAst_TmConst_194 {val = CmpAst_CEqi_214 ()}),("nth",ConstAst_TmConst_194 {val = SeqAst_CNth_218 ()})]
mexpr/mcore_parser.mc: ............................

david-broman:
Looks good. In the future, we need to think about how we include constant functions in a modular way (now they are added in a separate function).

PR #48, opened 2020-02-24
Added functions to seq.mc

br4sco:


elegios (approved):
Looks good. Wider/later questions:
- how many of these functions do we want to have in the standard library vs. somewhere else?
- do we do comparison by subtraction (as now) or returning enum values `LessThan`, `Equal`, `GreaterThan` (haskell style)?

PR #49, opened 2020-02-28
Implemented and tested semantic functions smap and sfold

br4sco:
So I have implemented the semantic functions `smap` and `sfold` as we discussed.

There are some attempts to combine them to maps/folds over a complete AST at the bottom of the tests.

I don't really see how one would use `bottomUp` to fold over an AST, to let's say collect all nodes in a list, if that is even possible. To me it seems more straightforward to define something like I did with `foldi`, but It might be a lack of understanding on my part.

### DIFF
--- a/stdlib/mexpr/ast.mc
+++ b/stdlib/mexpr/ast.mc
@@ -3,6 +3,12 @@ include "string.mc"
 lang VarAst
   syn Expr =
   | TmVar {ident : String}
+
+  sem smap_Expr_Expr (f : Expr -> A) =
+  | TmVar t -> TmVar t
+
+  sem sfold_Expr_Expr (f : A -> B -> A) (a : A) =
+  | TmVar t -> a
 end
 
 lang VarPat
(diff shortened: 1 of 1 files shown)


elegios (approved):
Looks good.

You can collect all nodes with `bottomUp`, but for efficiency you probably want to either use a paramorphism (better name is probably something like `bottomUpWithOriginal`) or implement it directly using `smap` and `sfold`.

A direct implementation (yielding all nodes in preorder I believe) would look something like:
```
recursive
  let allNodes = lam tm. sfold_Expr_Expr concat [tm] (smap_Expr_Expr allNodes tm)
end
```

elegios:
As a followup, to make it clearer how to use these things to write various forms of maps and folds it might be a good idea to write a "recursion cookbook" that gives examples of how to write a bunch of different styles of functions.

I'll bring this up on the next meeting.

br4sco:
Great! That would be helpful for me at least.

elegios:
I can't seem to edit the wiki for the project, so for the moment the recursion cookbook is available [here](https://roamresearch.com/#/app/Memory-Beta/page/M5R37ZAt2).

PR #51, opened 2020-02-28
Add MCore support for pretty printing strings

johnwikman:
Adds support in pprint.mc for pretty printing a CSeq that only contains TmConst {val = CChar _} as a string.


br4sco:
Looks good and does what it says. Is there any point in writing a test for this case?

br4sco (approved):

johnwikman:
At the moment it just tests that the pprintCode semantic doesn't fail. When the parser is in place, you would probably test it by the code below or something stricter.
```ml
utest parse (pprintCode 0 ast) with parse (pprintCode 0 (parse (pprintCode 0 ast))) in ()
```

PR #52, opened 2020-03-03
Mlang mangling, make language fragments share constructors

elegios:
Continuation of #41, changing it so that language fragments that include other fragments reuse the same constructors, instead of creating copies.

### DIFF
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -197,6 +197,9 @@ module Option = struct
   let bind f = function
     | Some x -> f x
     | None -> None
+  let value ~default = function
+    | Some x -> x
+    | None -> default
 end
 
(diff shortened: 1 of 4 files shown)


elegios:
This will not handle the case of two independent language fragments (say `A` and `B`) defining a constructor each of the same name (say `Con`), then a later fragment (say `C`) including both of them. I expect this should produce an error, but it will here arbitrarily reuse one of the previously defined constructors, i.e., one of the following tests will succeed, but not both:
```
let c1 = use A in Con {} in
let c2 = use B in Con {} in
let c3 = use C in Con {} in
utest c1 with c3 in
utest c2 with c3 in
()
```

I make the judgement that that is ok, since we expect that sort of thing to be caught by typechecking at a later stage.

PR #60, opened 2020-03-19
Include the filename and line number in error prints

johnwikman:
Adds the filename and line number of the error print to the printed message. Currently on the format `ERROR (<filename>:<line number>): <msg>`. At the moment I am having to spend a lot of time having to manually enter metadata into each error print to know where in the code it happened. Would be nice to just put in the message that describes what actually went wrong.

The format of the filename+line number metadata was what looked OK after a short moment of thinking. Comments and opinions are welcome.

### DIFF
--- a/src/boot/mexpr.ml
+++ b/src/boot/mexpr.ml
@@ -368,8 +368,12 @@ let delta eval env fi c v  =
         Sys.remove (Ustring.to_utf8 (tmlist2ustring fi lst)); TmConst(NoInfo,Cunit)
     | CdeleteFile,_ -> fail_constapp fi
 
-    | Cerror, TmSeq(fi,lst) ->
-       (uprint_endline ((us"ERROR: ") ^. (tmlist2ustring fi lst)); exit 1)
+    | Cerror, TmSeq(fiseq,lst) ->
+       (let prefix = match fi with
+                     | Info(filename,l1,_,_,_) ->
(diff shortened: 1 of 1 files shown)


br4sco:
This is great! I do not have a strong opinion on the error message format but perhaps we could conform to some standard (maybe you have already), to make tool integration easier. 

E. g.
```
sourcefile:lineno: message
```

https://www.gnu.org/prep/standards/html_node/Errors.html

johnwikman:
Thank you! Yes I agree, it is good to conform to standards as much as possible. Now the format of the prints will look as follows:
```
<sourcefile>:<lineno>: ERROR: <msg>
```

br4sco (approved):

br4sco:
Looking more into it, can't we use `raise_error` in `msg.ml` here as well? Otherwise the error message format for `error` will be different compared to error messages from the parsing and failed const applications.

br4sco (commented):
Perhaps call `raise_error` in `msg.ml` instead of having a separate error message handling for `error`

johnwikman:
Could be a good idea to use `raise_error`, but that means the compile-time errors will look the same as the runtime errors (assertions really). Is that something we want?

br4sco:
Good point! I was thinking that the error format could be the same, but that the message part of the error should indicated the type of error. We can discuss it tomorrow.

PR #61, opened 2020-03-23
Mexprfix

david-broman:
Refactored and added a number of things to MExpr in boot:

- Added matching of sequences in Mexpr.
- Refactored to have separate pattern wildcards.
- Removed CSeq from boot.
- Removed TmIf in boot. Replaced with match.

### DIFF
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -163,10 +163,16 @@ and name =
 | NameStr of ustring                              (* A normal pattern name *)
 | NameWildcard                                    (* Pattern wildcard *)
 
+(* Kind of sequence matching in patterns *)
+and seqMatchType =
+| SeqMatchPrefix of name
+| SeqMatchPostfix of name
+| SeqMatchTotal
 
 (* Patterns *)
 and pat =
 | PatNamed of info * name                         (* Named, capturing wildcard *)
+| PatSeq   of info * pat list * seqMatchType      (* Sequence pattern *)
 | PatTuple of info * pat list                     (* Tuple pattern *)
 | PatCon   of info * ustring * sym * pat          (* Constructor pattern *)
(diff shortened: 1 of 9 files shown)


src/boot/ast.ml
```diff
@@ -152,21 +148,39 @@ and tm =
...
+| TmFix     of info                                                 (* Fix point *)
+
+
 
 and label =
 | LabIdx of int                                   (* Tuple index *)
 | LabStr of ustring                               (* Record label *)
 
+(* Kind of pattern name *)
+and name =
```
elegios: I'd probably prefer pat_name or something like it, `name` seems a bit generic

src/boot/mexpr.ml
```diff
@@ -429,7 +428,15 @@ let rec debruijn env t =
      | VarTm(y)::ee -> if y =. x then n else find fi ee (n+1) x
      | [] -> raise_error fi ("Unknown variable '" ^ Ustring.to_utf8 x ^ "'")) in
   let rec dbPat env = function
-    | PatNamed(fi,x) -> (VarTm(x)::env, PatNamed(fi,x))
+    | PatNamed(_,NameStr(x)) as pat -> (VarTm(x)::env, pat)
+    | PatNamed(_,NameWildcard) as pat -> (VarTm(us"")::env, pat)
```
elegios: I suspect that it should be possible to not add anything to the environment here, as long as nothing is added in `tryMatch` either.

src/boot/ast.ml
david-broman: Sure, I changed to `patName`. We have decided to not to use snake case.

src/boot/mexpr.ml
david-broman: Thanks. I have updated to not bind the name when it is a wildcard.

PR #65, opened 2020-03-27
Make sure filepaths handled in includes are absolute and normalized

elegios:
This fixes the issue mentioned in #63. The issue was that the mechanism for not including a file twice was looking solely at its path, but that path could be either relative to the current working directory of `boot` or an absolute path. If both of these are included they would be treated as different files. This PR makes sure to prepend the current working directory early on, with the hope that that'll produce a single canonical path.

Note that this doesn't work in general, e.g., in the presence of symlinks and the like. There are library functions for producing a canonical path, but they would incur an extra dependency, and this works for now at least.

### DIFF
--- a/src/boot/boot.ml
+++ b/src/boot/boot.ml
@@ -94,7 +94,7 @@ let rec merge_includes root visited = function
   | Program(includes, tops, tm) ->
      let rec parse_include root = function
        | Include(info, path) as inc ->
-          let filename = Filename.concat root (Ustring.to_utf8 path) in
+          let filename = Filename.concat root (Ustring.to_utf8 path) |> Utils.normalize_path in
           if List.mem filename visited
           then raise_error info ("Cycle detected in included files: " ^ filename)
           else if List.mem filename !parsed_files
--- a/src/boot/boot.ml
+++ b/src/boot/boot.ml
@@ -127,6 +127,11 @@ let add_prelude = function
 (* Main function for evaluation a function. Performs lexing, parsing
    and evaluation. Does not perform any type checking *)
 let evalprog filename  =
+  (* Make sure the filename is an absolute path, otherwise the duplicate file detection won't work *)
+  let filename =
+    if Filename.is_relative filename
+    then Filename.concat (Sys.getcwd ()) filename
+    else filename in
   if !utest then printf "%s: " filename;
   utest_fail_local := 0;
   begin try
--- a/src/boot/dune-boot
+++ b/src/boot/dune-boot
@@ -5,4 +5,5 @@
 
 (executable
  (name boot)
+ (libraries str)
 )
--- a/src/boot/dune-boot-ext
+++ b/src/boot/dune-boot-ext
@@ -5,5 +5,5 @@
 
 (executable
  (name boot)
- (libraries sundialsml)
+ (libraries str sundialsml)
 )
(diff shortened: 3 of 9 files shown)


johnwikman:
Looks good, and solves the issue raised in #63 when running on my machine! For a future PR, it would perhaps be good to make the filename prints be relative to current working directory when running the test cases. At the moment the prints look like this for me:
```
/home/john/GitRepos/Forks/miking/test/mexpr/bool.mc: ........... OK
/home/john/GitRepos/Forks/miking/test/mexpr/types.mc: ................ OK
/home/john/GitRepos/Forks/miking/test/mexpr/int.mc: ................................ OK
/home/john/GitRepos/Forks/miking/test/mexpr/records.mc: ........ OK
/home/john/GitRepos/Forks/miking/test/mexpr/reclets.mc: ........ OK
/home/john/GitRepos/Forks/miking/test/mexpr/letlamif.mc: ............ OK
/home/john/GitRepos/Forks/miking/test/mexpr/seq.mc: ................................ OK
/home/john/GitRepos/Forks/miking/test/mexpr/ident.mc: .... OK
/home/john/GitRepos/Forks/miking/test/mexpr/tuples.mc: ...... OK
/home/john/GitRepos/Forks/miking/test/mexpr/match.mc: .......................................... OK
/home/john/GitRepos/Forks/miking/test/mexpr/effects.mc: ....... OK
/home/john/GitRepos/Forks/miking/test/mexpr/string.mc: ............... OK
/home/john/GitRepos/Forks/miking/test/mexpr/stringops.mc: ............. OK
/home/john/GitRepos/Forks/miking/test/mexpr/nestedpatterns.mc: ........ OK
/home/john/GitRepos/Forks/miking/test/mexpr/float.mc: ..................................................................... OK
/home/john/GitRepos/Forks/miking/test/mexpr/symbs.mc: .... OK
/home/john/GitRepos/Forks/miking/test/mexpr/fix.mc: .......... OK
Unit testing SUCCESSFUL after executing 297 tests.
```

elegios:
After looking into this some more it turns out that `Filename.concat` does no normalization, i.e., `..` in a path will introduce duplication of parsing pretty much immediately. Options now include:

- Incurring a dependency on a library that has path normalization:
  - https://ocaml.janestreet.com/ocaml-core/109.07.00/doc/core/Filename.html (this strongly assumes POSIX, i.e., no windows support)
  - https://github.com/solvuu/phat (this is a strongly typed approach, not centered around strings; relative/absolute and file/directory/link are reflected in the type GADT-style)
  - https://github.com/dbuenzli/fpath
- Implementing proper path normalization ourselves (with some limitations, say only normalizing to absolute paths with no `.` or `..`, not considering links and the like) for both windows and posix style paths
- Using a non-filepath dependent approach, e.g.
  - Having a set of previously produced ASTs, only adding an AST if it's not already in the set. I have a working implementation of this. It is worth mentioning that it can technically make us reuse things when we shouldn't, if two distinct files have code that represents the exact same AST. It's additionally a little bit heavier than a path based thing, since we may parse a file multiple times. I doubt that difference would matter that much though.
  - Hash filecontents, don't parse a file if we've seen the hash before. Similar caveat to the previous point.

I think @david-broman should probably have an opinion here. I would probably advocate for incurring a dependency.

david-broman:
As discussed at the last meeting, just let me know when you have added a test case (ping me on Slack).

david-broman:
Regarding your question Viktor: As I see it now, I would actually prefer if we could create our own way of doing path normalization (that works on windows, Linux, and MacOS) in OCaml, without external dependencies. If we do this in boot, we can then translate this code into MExpr when we need to do this again.

elegios:
Ok. I'll make a version of that and change this PR to it instead. Initially I will only do it syntactically, i.e., I will not consider symlinks and the like, just making paths absolute and with minimal amounts of `..` and `.`.

david-broman:
Sounds good.

PR #70, opened 2020-04-03
Replaced sequence implementation with finger trees 

br4sco:
Here we have changed the underlying implementation of `mexpr` sequences from lists to the finger trees implementation in [https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatFingerTree.html](https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatFingerTree.html) to improve the performance of sequence operations.


stdlib/seq.mc
```diff
@@ -1,9 +1,15 @@
 include "option.mc"
 
-let head = lam s. nth s 0
-let tail = lam s. slice s 1 (length s)
 let null = lam seq. eqi 0 (length seq)
 
+let slice = lam seq. lam n. lam m.
```
johnwikman: Maybe I am being too pendantic, but could it be an idea to rename the `n` and `m` arguments to `off` (or `offset`) and `cnt` (or `count`) respectively? It would aid readability since `n` and `m` does not tell me very much about what those arguments represent.

br4sco: Sure!

src/boot/ast.ml
```diff
@@ -172,7 +174,7 @@ and seqMatchType =
 (* Patterns *)
 and pat =
 | PatNamed of info * patName                      (* Named, capturing wildcard *)
-| PatSeq   of info * pat list * seqMatchType      (* Sequence pattern *)
+| PatSeq   of info * pat Mseq.t * seqMatchType    (* Sequence pattern *)
```
EliasC, line 177: Should this really be an `Mseq.t`? It seems like this is representing a piece of MExpr syntax and not a runtime value.

src/boot/mexpr.ml
```diff
@@ -141,6 +147,7 @@ let fail_constapp f v fi = raise_error fi ("Incorrect application. function: "
    The reason for this is that if-expressions return expressions
    and not values. *)
 let delta eval env fi c v  =
+    let index_out_of_bounds_in_seq_msg = "Out of bound access in sequence." in
```
EliasC: "Out of bound**s**"?

src/boot/mexpr.ml
```diff
@@ -481,29 +493,35 @@ let rec debruijn env t =
...
-
 let rec tryMatch env value pat =
   let go v p env = Option.bind (fun env -> tryMatch env v p) env in
-  let rec splitNth n = function [] -> ([],[]) | x::xs ->
-    if n > 0 then let (pre,post) = splitNth (n-1) xs in (x::pre,post) else ([],x::xs) in
+  let splitNth n s =
+    assert (n <= Mseq.length s);
+    if Mseq.length s == 0 then (Mseq.empty, Mseq.empty)
+    else if n < Mseq.length s then Mseq.split_at s n
+    else (s, Mseq.empty)
```
EliasC: Why are these special cases needed? Is our intended semantics of `split_at` different from the one in the library?

stdlib/seq.mc
```diff
@@ -1,8 +1,18 @@
 include "option.mc"
 
-let head = lam s. nth s 0
-let tail = lam s. slice s 1 (length s)
 let null = lam seq. eqi 0 (length seq)
+let head = lam seq. get (splitAt seq 1).0 0
+let tail = lam seq. (splitAt seq 1).1
+let last = lam seq. get (splitAt seq (subi (length seq) 1)).1 0
+let init = lam seq. (splitAt seq (subi (length seq) 1)).0
```
EliasC, line 7: Is `splitAt` clever enough to do the efficient thing here?

test/mexpr/seq.mc
```diff
@@ -10,43 +10,45 @@ utest [] with [] in
...
-utest slice ['a','b'] 1 10 with ['b'] in
-utest slice [1,3] 2 10 with [] in
+// snoc xs x
+utest snoc [1,2] 3 with [1,2,3] in
+utest snoc [] 1 with [1] in
+
+// splitAt seq n
+utest splitAt [1,2,3] 0 with ([],[1,2,3]) in
+utest splitAt [1,2,3] 1 with ([1],[2,3]) in
+utest splitAt [1,2,3] 2 with ([1,2],[3]) in
```
EliasC, line 45: I'm surprised that `splitAt [1,2,3] 3` does not equal `([1,2,3, [])` (now it throws an error)! I think this is the behaviour of the finger tree library.

EliasC (changes requested):
This mostly looks really good! I think the corner case of `splitAt l n` when `n` = `length l` should work though.

EliasC:
Another comment is that we should be able to remove the home-rolled `Option` modules and replace them by the one from Batteries, but I can make sure to remember this.

src/boot/mexpr.ml
br4sco: Yes, this local function `slitNth` has a different semantics. Less so if we change the semantics of `split_at` so that `split_at [1,2,3] 3` equals `([1,2,3],[])`. Still `splitNth s 0` equals `([],[])` which is not true for `split_at` and I don't think we want `split_at` to have this semantics. In any case, I agree that we should depart from the `split_at [1,2,3] 3` semantics of batteries and change the code a bit to remove `splitNth` altogether.

stdlib/seq.mc
br4sco: I believe so, see the implementation of `split_at` in `mseq.ml`

test/mexpr/seq.mc
br4sco: `split_at` in batteries has this semantics, but we might want to change this semantics for `splitAt`

src/boot/ast.ml
br4sco: The main reason for this data-structure is that we apply a `fold_right2` over a sequence of patterns and a sequence when matching over sequences in `mexpr.ml` line 518. If we keep these patterns in a list we would have to write a custom `fold_right2` that folds over a sequence and a list rather than two sequences.

br4sco: Idk what is best

src/boot/mexpr.ml
br4sco: Ok!

src/boot/mexpr.ml
br4sco: I decided to keep `splitNth` but change its name to reflect the difference in semantics.

test/mexpr/seq.mc
br4sco: Changed semantics of `splitAt`

src/boot/mexpr.ml
```diff
@@ -147,7 +147,7 @@ let fail_constapp f v fi = raise_error fi ("Incorrect application. function: "
    The reason for this is that if-expressions return expressions
    and not values. *)
 let delta eval env fi c v  =
-    let index_out_of_bounds_in_seq_msg = "Out of bound access in sequence." in
+    let out_of_bounds_in_seq_msg = "Out of bounds." in
```
EliasC: Sorry, I was unclear! My comment was that it should be "Out of bounds" and not "out of bound". I think the message should be "Out of bounds access in sequence". Just "out of bounds" is a bit unspecific.

br4sco: Fixed

src/boot/ast.ml
EliasC: That sounds reasonable!

EliasC (approved):
I'm happy with this now! Well done!

src/boot/ast.ml
br4sco: I thought about it some more, I think we can change it to a list and then just convert it to a sequence when doing a matching. Since the number of patterns will only be "a handful", speed should not be a factor.

PR #72, opened 2020-04-07
Improved pretty printing in boot

dlunde:
I've added additional pretty printing functionality to boot, mostly to make it easier for myself to learn how the mlang transformations work. Maybe others can benefit from this as well.


src/boot/boot.ml
```diff
@@ -211,27 +219,49 @@ let testprog lst =
 
 (* Run program *)
 let runprog name lst =
+    (* TODO prog_argv is never used anywhere *)
```
elegios, line 222: Looking at the code in `mexpr.ml` it appears to grab `argv` separately, maybe this is left from some earlier thought on how it should be done?

(not particularly relevant to this PR, but maybe worth considering later)

src/boot/boot.ml
```diff
@@ -211,27 +219,49 @@ let testprog lst =
...
 
 (* Main function. Checks arguments and reads file names *)
 let main =
 
-  (* Check command  *)
-  (match Array.to_list Sys.argv |> List.tl with
+  (* A list of command line arguments *)
+  let speclist = [
+
+    (* First character in description string must be a space for alignment! *)
```
elegios, line 236: Is this something that `Arg` dictates?

src/boot/pprint.ml
```diff
@@ -1,244 +1,492 @@
-
-(*
-   Miking is licensed under the MIT license.
-   Copyright (C) David Broman. See file LICENSE.txt
-*)
+(** Pretty-printing for data types in Miking.
+ *
+ *  The main functions that should be used outside of this module is
```
elegios, line 3: Maybe we should consider making a `.mli`? Maybe overkill for `boot`?

src/boot/pprint.ml
```diff
@@ -1,244 +1,492 @@
...
+    print_const c
+
+(** Converting environments to strings.
+ *  TODO Messy with optional arguments passing. Alternatives? *)
+let string_of_env ?debruijn ?indent ?max_indent ?margin ?max_boxes ?prefix e =
+  str_formatter_print ?debruijn ?indent ?max_indent ?margin ?max_boxes ?prefix
+    print_env e
+
+(** Old printing function for terms.
+ *  @deprecated Use string_of_tm instead. *)
```
elegios: Question for meeting: do we want to deprecate and then remove, or just remove and make necessary changes?

elegios (approved):

src/boot/boot.ml
dlunde: Yes!

src/boot/pprint.ml
dlunde: I thought about this as well

dlunde:
The comments from the previous meeting has now been addressed:
- Removed deprecated functions (this also resolved camel vs snake case)
- Return ustrings instead of strings

PR #73, opened 2020-04-07
Lambda lifting: Include globaldefs from the scanned reclets bodies

johnwikman:
Added test case that also checks that a lifted expression from a reclets body is correctly moved to the top level.

~~Also reversed the order of includes since the include path fix from a while ago (#65) does not appear to have solved the issue completely. This error can be reproduced on my machine by running `make` and `make install` in the repostiory root, restoring the include order in `lamlift.mc:38:39`, entering the `stdlib/mexpr` subdirectory, and running `miking test lamlift.mc`. The error is not reproduced on my machine by running `make test` in the repository root.~~

Edit: Restored the order of the includes due to the explanation below.

### DIFF
--- a/stdlib/mexpr/lamlift.mc
+++ b/stdlib/mexpr/lamlift.mc
@@ -501,7 +501,7 @@ lang RecLetsLamlift = VarLamlift + RecLetsAst + FunLamlift + ConstAst + UnitAst
 
         st_addVarToEnv oldname binding accstate
       in
-      let envstate = foldl envgen {state with id = liftedstate.id} liftedbindings in
+      let envstate = foldl envgen {{state with id = liftedstate.id} with globaldefs = liftedstate.globaldefs} liftedbindings in
 
       -- Replace all internal occurrences with the newly bound values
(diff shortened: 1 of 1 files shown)


elegios:
The issue here is actually a different one, some of the imports are being resolved to the installed stdlib, some to the stdlib in the repository. These are different files, thus we would most likely expect them to be included multiple times.

In this particular case `eval.mc` has the include `mexpr/ast.mc`, which does not exist relative to the current location, but it does exist relative to the stdlib root.

Not entirely sure what we would like to do here. One option is to make it explicit whether we want to include something relative to the current directory or the stdlib (`"in-stdlib"` vs `./somewhere-relative`, or C's `#include <lib>` vs `#include "local"`). It may also be the case that it doesn't make sense to run tests on the stdlib with an installed executable (since it will include its own stdlib as well).

In any case I suspect that testing the stdlib needs some special treatment, one way or another.

As a conclusion though, the ordering of includes here will only matter when running `boot` with another copy of the stdlib somewhere else, so it should be fine to have the desired ordering here.

PR #74, opened 2020-04-08
Add implementation of a directed graph

lingmar:
This PR adds an implementation of a directed graph data structure to stdlib. Please see the comments in the top of the digraph.mc file for an overview of how a graph is represented.

Some points to discuss:
- Internal representation of the graph: right now, it uses an adjacency map, which maps a vertex to a list of outgoing edges from that vertex. As we don't have an efficient map data structure such as hash map yet, the map is a rather primitive associated list, where search and update takes linear time. The complexity can be improved if an adjacency _list_ was used instead (as we will soon have sequences with random access), but that would force vertices to be integers. Then, one might have to use a map anyway when using the graph, if one wants other types than integers to be vertices.
- The prefix 'digraph' creates rather long function names, should the prefix be e.g. 'dg' instead?
- Note that uniq in seq.mc runs in O(n^2) time, this can probably be improved.

### DIFF
--- a/stdlib/digraph.mc
+++ b/stdlib/digraph.mc
@@ -1,0 +1,169 @@
+include "option.mc"
+include "seq.mc"
+
+-- Represents a directed graph with labeled edges.
+
+-- Vertices and labels can be of any data types, as long as they have an
+-- equality function (given as arguments to the graph when created).
+
+-- An edge is represented as a triple (from, to, label), where the edge goes
+-- from "from" to "to", and "label" is the label of the edge. There can be
+-- several edges between two vertices.
+
+-- Vertices must be unique: there cannot be two vertices whose value of the
+-- equality function is true. This is maintained as invariant. Similarly, labels
+-- between any pair of vertices must be unique; this is also maintained as
+-- invariant.
+
+-- Presently, the graph is implemented using an adjacency map, which maps each
+-- vertex to a list of outgoing edges from that vertex.
+
+-- Map functions
+let mapLookup = lam key. lam eq. lam m.
+                match findAssoc (eq key) m with Some e
+                then e
+                else error "Element not found"
+
+recursive
+let mapUpdate = lam key. lam f. lam eq. lam m.
+                let k = (head m).0 in
+                let v = (head m).1 in
+                if eq key k then
+                  cons (k, f v) (tail m)
+                else
+                  cons (head m) (mapUpdate key f eq (tail m))
+end
+
+-- Returns an empty graph. Input: equality functions for vertices and labels.
+let digraphEmpty = lam eqv. lam eql. {adj = [], eqv = eqv, eql = eql}
+
+-- Access vertices and edges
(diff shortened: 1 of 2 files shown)


EliasC:
> The prefix 'digraph' creates rather long function names, should the prefix be e.g. 'dg' instead?

My feeling is that we should optimize for readability right now (i.e. pick the longer name) and then shorten it when we have namespaces/type classes/whatever to handle name clashes.

> Note that uniq in seq.mc runs in O(n^2) time, this can probably be improved.

I would guess that the only way to improve this is to use an auxiliary data structure with fast lookup to remember which elements have already been seen, and that brings you back to your first bullet: you need something like a hash map to get efficient lookup. You could go for something like a search tree if you had an ordering of your elements.

Another big efficiency thief right now is that `null` is implemented as `eqi 0 (length s)`, meaning it's linear in the length of the list 😬

EliasC:
> Another big efficiency thief right now is that `null` is implemented as `eqi 0 (length s)`

I just realized that this will be fixed by #70!

lingmar:
> I would guess that the only way to improve this is to use an auxiliary data structure with fast lookup to remember which elements have already been seen, and that brings you back to your first bullet: you need something like a hash map to get efficient lookup. You could go for something like a search tree if you had an ordering of your elements.

First sorting the list and then remove the duplicates should run in O(n*logn) at least.

EliasC:
> First sorting the list and then remove the duplicates should run in O(n*logn) at least.

True, but that would also require there to be an ordering (same as the search tree).

lingmar:
> True, but that would also require there to be an ordering (same as the search tree).

True. And requiring ordering for vertices would exclude symbols (CSymb) to be used.

stdlib/digraph.mc
```diff
@@ -0,0 +1,169 @@
...
+-- several edges between two vertices.
+
+-- Vertices must be unique: there cannot be two vertices whose value of the
+-- equality function is true. This is maintained as invariant. Similarly, labels
+-- between any pair of vertices must be unique; this is also maintained as
+-- invariant.
+
+-- Presently, the graph is implemented using an adjacency map, which maps each
+-- vertex to a list of outgoing edges from that vertex.
+
```
dlunde, line 20: Suggestion: for readability, I would explicitly add the type of the graph. That is, something similar to
`type Digraph a = { adj : [(a,[(a,b)])], eqv : a -> a -> Bool, eql : b -> b -> Bool }` (not sure if the boot interpreter supports type variables yet though).

dlunde (approved):
Looks good to me!

stdlib/digraph.mc
lingmar: Good idea. For now, added the type definition without the type parameters, i.e. just `type Digraph = { adj : [(a,[(a,b)])], eqv : a -> a -> Bool, eql : b -> b -> Bool }`

PR #82, opened 2020-04-22
Loosen seqpat restriction

elegios:
When originally implemented we placed restrictions on concat patterns: one side must be a fixed length sequence, and the other must be wildcard. This was done to prevent writing the pattern `_ ++ [1, 2, 3] ++ _` which would match any sequence that has `[1, 2, 3]` as a subsequence. This would not be obvious how to implement and it would be too fancy a thing to have in a core language.

However, this also prevents `[1, 2] ++ mid ++ [3, 4]` which is *not* particularly fancy. This PR thus loosens the restriction to the following: one side must be a fixed length sequence pattern, and the other may be any pattern.

Additionally there was a parser bug where `[1, 2] ++ _` would parse the last part as the name `_` and bind a value to it, instead of seeing it as a wildcard that doesn't bind anything. This PR fixes that as an extra bonus. It also fixes a bug that was invisible due to this parser bug: the de bruijn translation did not add anything to the environment for a non-binding wildcard pattern, but the matching function `tryMatch` did. This would make every name inside the `then` branch of a match refer to the wrong value (assuming the name was bound outside the match).

### DIFF
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -168,8 +168,8 @@ and patName =
 
 (* Kind of sequence matching in patterns *)
 and seqMatchType =
-| SeqMatchPrefix of patName
-| SeqMatchPostfix of patName
+| SeqMatchPrefix of pat
+| SeqMatchPostfix of pat
 | SeqMatchTotal
 
 (* Patterns *)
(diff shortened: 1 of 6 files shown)


EliasC:
Regarding the parser bug, it seems a bit strange to me that `let _ = 42 in _` is well-formed, but `match 42 with _ then _ else 0` is not. Should `_` ever be considered a name? (doesn't have to be fixed in this PR)

PR #83, opened 2020-04-23
Sequence based vector, matrix and set libraries 

br4sco:
These libraries implements set, vector and matrix operations over sequences.

### DIFF
--- a/stdlib/matrix.mc
+++ b/stdlib/matrix.mc
@@ -1,0 +1,111 @@
+-- A simple library that defines matrix operations over sequences of sequences.
+-- Matrices are represented as sequences of rows, represented as sequences.
+
+include "vec.mc"
+
+-- True if mtx is a matrix, otherwise false.
+let matrixIsMatrix = lam mtx.
+  and (not (null mtx))
+      (all (lam row. and (eqi (length (head mtx)) (length row))
+                         (gti (length row) 0))
+           mtx)
+
+-- Makes matrix of dimensions d = (m, n) with elements c.
+let matrixConst = lam d. lam c. makeSeq d.0 (makeSeq d.1 c)
+
+-- Get element (i, j) from matrix mtx.
+let matrixGet = lam mtx. lam i. lam j. get (get mtx i) j
+
+-- Set element (i, j) of matrix mtx to v
+let matrixSet = lam mtx. lam i. lam j. lam v. set mtx i (set (get mtx i) j v)
+
+-- Apply binary operator bop elementwise over the matrices mtx1 and mtx2.
+-- Results in an error if mtx1 and mtx2 are of different dimensions or empty.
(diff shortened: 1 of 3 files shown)


johnwikman:
This is more of a comment regarding matrix representation in general than about the implementation. When I perform GPU optimized matrix operations in MCore I rely on the fact that the matrix is represented as flat sequence of elements. Then I adjust the indexing when accessing the individual elements. In a way, the type looks like this for me:
```
type Matrix a = {rows : Int, cols : Int, v : [a]}
```
In the case of the sequence in sequence representation, if we were to perform matrix multiplication with 4000x4000 matrices on a GPU we would have to either flatten them first or perform ~12000 individual memcpys (2 input + 1 output matrix). On the other hand it would be more efficient to perform operations such as swapping rows by using the sequence in sequence representation.

br4sco:
> This is more of a comment regarding matrix representation in general than about the implementation. When I perform GPU optimized matrix operations in MCore I rely on the fact that the matrix is represented as flat sequence of elements. Then I adjust the indexing when accessing the individual elements. In a way, the type looks like this for me:
> 
> ```
> type Matrix a = {rows : Int, cols : Int, v : [a]}
> ```
> 
> In the case of the sequence in sequence representation, if we were to perform matrix multiplication with 4000x4000 matrices on a GPU we would have to either flatten them first or perform ~12000 individual memcpys (2 input + 1 output matrix). On the other hand it would be more efficient to perform operations such as swapping rows by using the sequence in sequence representation.

I see. I could hide the matrix implementation from the interface, which is  a good thing in general. I didn't bother now since this is anyways temporary, but if this library is something we both could use I don't mind changing the implementation.

br4sco:
Thinking a little about I think I will remove the vector library and just define a matrix library to keep this as slim as possible.

johnwikman:
> > This is more of a comment regarding matrix representation in general than about the implementation. When I perform GPU optimized matrix operations in MCore I rely on the fact that the matrix is represented as flat sequence of elements. Then I adjust the indexing when accessing the individual elements. In a way, the type looks like this for me:
> > ```
> > type Matrix a = {rows : Int, cols : Int, v : [a]}
> > ```
> > 
> > 
> > In the case of the sequence in sequence representation, if we were to perform matrix multiplication with 4000x4000 matrices on a GPU we would have to either flatten them first or perform ~12000 individual memcpys (2 input + 1 output matrix). On the other hand it would be more efficient to perform operations such as swapping rows by using the sequence in sequence representation.
> 
> I see. I could hide the matrix implementation from the interface, which is a good thing in general. I didn't bother now since this is anyways temporary, but if this library is something we both could use I don't mind changing the implementation.

I only needed a few matrix functions to run some benchmarks, so do not worry about me! I was more concerned about the future non-temporary implementation, but maybe it is too early to discuss that now.

lingmar:
Looks good to me!

PR #87, opened 2020-05-03
Changed from debruijn indicies to symbol representation of variable and pattern names in boot.

david-broman:
Updated the boot interpreter to use symbols instead of debruijn indices for representing names (variable names, pattern names etc.). The rationale is two fold: (1) it is easier to implement OR-patterns, and (2) this is the pattern we will use in the MLang implementation because it simplifies program transformations.

### DIFF
--- a/src/boot/mexpr.ml
+++ b/src/boot/mexpr.ml
@@ -14,7 +14,7 @@ open Printf
    correspond constants *)
 let builtin =
   let f c = TmConst(NoInfo,c) in
-  [("unit",f(Cunit));
+  ([("unit",f(Cunit));
    ("not",f(Cnot));("and",f(Cand(None)));("or",f(Cor(None)));
    ("addi",f(Caddi(None)));("subi",f(Csubi(None)));("muli",f(Cmuli(None)));
    ("divi",f(Cdivi(None)));("modi",f(Cmodi(None)));("negi",f(Cnegi));
@@ -49,7 +49,15 @@ let builtin =
    ("eqs", f(Ceqs(None))); ("gensym", f(Cgensym))
   ]
   (* Append external functions TODO: Should not be part of core language *)
-  @ Ext.externals
+  @ Ext.externals)
+  |> List.map (fun (x,t) -> (x,gensym(),t))
+
+(* Mapping name to symbol *)
+let builtin_name2sym = List.map (fun (x,s,_) -> (us x,s)) builtin
+
+(* Mapping sym to term *)
+let builtin_sym2term = List.map (fun (_,s,t) -> (s,t)) builtin
+
 
 (* Returns the number of expected arguments of a constant *)
 let arity = function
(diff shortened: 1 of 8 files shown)


src/boot/mexpr.ml
```diff
@@ -438,28 +448,28 @@ let rec val_equal v1 v2 =
...
-  let rec find fi env n x =
-    (match env with
-     | VarTm(y)::ee -> if y =. x then n else find fi ee (n+1) x
-     | [] -> raise_error fi ("Unknown variable '" ^ Ustring.to_utf8 x ^ "'")) in
+(* Add symbol associations between lambdas, patterns, and variables *)
+let rec symbolize env t =
+  let findsym fi x kind env = try List.assoc x env
+    with Not_found -> raise_error fi ("Unknown " ^ kind ^ " '" ^ Ustring.to_utf8 x ^ "'")
+  in
   let rec dbPat env = function
```
elegios: The name `dbPat` stems from de bruijn indices, should perhaps be renamed now that we're removing them (I'll probably do that during my changes if it's not done here).

src/boot/mexpr.ml
```diff
@@ -438,28 +448,28 @@ let rec val_equal v1 v2 =
...
+         | SeqMatchPostfix(NameStr(x,_)) -> let s = gensym() in ((x,s)::env, SeqMatchPostfix(NameStr(x,s)))
+         | SeqMatchPrefix(NameWildcard) | SeqMatchPostfix(NameWildcard) | SeqMatchTotal -> (env,seqMP)) in
+       (env,PatSeq(fi,ps,seqMP'))
     | PatTuple(fi,ps) -> (* NOTE: this causes patterns to introduce names right-to-left, which will cause errors if a later pattern binds a name that is seen as a constructor in a pattern to the left *)
        let go p (env,ps) = let (env,p) = dbPat env p in (env,p::ps) in
        let (env,ps) = List.fold_right go ps (env,[])
        in (env,PatTuple(fi,ps))
     | PatCon(fi,cx,_,p) ->
-       let cxId = find fi env 0 cx in
+       let cxId = findsym fi cx "constructor" env in
```
elegios, line 472: This means that constructors and normal variables still live in the same namespace, which I believe isn't what we want in the long run (it isn't quite the case in the interpreter in mlang). Not sure if we should fix that now or later.

src/boot/mlang.ml
```diff
@@ -159,16 +159,16 @@ let translate_cases f target cases =
   let translate_case case inner =
     match case with
     | (ConPattern (fi, k, x), handler) ->
-      TmMatch (fi, target, PatCon(fi, k, noidx, PatNamed(fi, NameStr(x))), handler, inner)
+      TmMatch (fi, target, PatCon(fi, k, noidx, PatNamed(fi, NameStr(x,0))), handler, inner)
```
elegios: I feel like `0` here should be `noidx`, no? Probably won't matter since `symbolize` is run after, but still.

elegios (approved):

src/boot/mexpr.ml
david-broman: I was planning to rework some parts on constructors in another patch.

src/boot/mexpr.ml
david-broman: OK, good. Changed to sPat.

src/boot/mlang.ml
david-broman: Changed to `nosym` insteadof `noidx`and updated on the necessary places in `mlang.ml`.

PR #89, opened 2020-05-11
Mexpr simp

david-broman:
Removed TmTuple and projections (both for records and tuples) in Boot. Changed so that unit now is an empty record. 

Note: The PR works for `make test`, but not for `make externals-test`. I think it is better that Oscar makes the refactoring for the external dependencies.


src/boot/mexpr.ml
```diff
@@ -398,11 +393,10 @@ let delta eval env fi c v  =
         in uprint_endline (prefix ^. us"ERROR: " ^. (tmseq2ustring fiseq lst)); exit 1)
     | Cerror,_ -> fail_constapp fi
     | CdebugShow,t ->
-       uprint_endline ((us"EXPR: ") ^. (ustring_of_tm t));
-       TmConst(NoInfo,Cunit)
+       uprint_endline ((us"EXPR: ") ^. (ustring_of_tm t)); tmUnit
 
     | CSymb(_),_ -> fail_constapp fi
-    | Cgensym, TmConst(fi, Cunit) -> TmConst(fi, CSymb(gen_symid()))
+    | Cgensym, TmRecord(fi,x) when x = Record.empty -> TmConst(fi, CSymb(gen_symid()))
```
elegios: `Record.is_empty` probably

src/boot/parser.mly
```diff
@@ -270,10 +272,15 @@ left:
...
-  | LPAREN RPAREN        { TmConst($1.i, Cunit) }
+  | atom DOT proj_label
+      { let fi = mkinfo (tm_info $1) (fst $3) in
+        let id = unique_ident in
+        TmMatch(fi,$1,PatRecord(fi,Record.singleton (snd $3) (PatNamed(fi,NameStr(id,nosym)))),
+                                TmVar(fi,id,nosym), TmNever(fi)) }
+  | LPAREN seq RPAREN
+      { if List.length $2 = 1 then List.hd $2
+        else tuple2record (mkinfo $1.i $3.i) $2 }
+  | LPAREN RPAREN        { TmRecord($1.i, Record.empty) }
```
elegios, line 283: This means we can write unit both as `()` and `unit` (and `{}`)

src/boot/pprint.ml
```diff
@@ -326,20 +320,25 @@ and print_tm' fmt t = match t with
...
-      | LabStr s -> string_of_ustring s in
-    fprintf fmt "%a.%s" print_tm (Atom, t) l
+    if r = Record.empty then fprintf fmt "()"
+    else
+      let parsed_record_to_tuple =
+        List.fold_left (fun (a,k) (i,tm) ->
+          match a with
+          | Some _ when (try (int_of_ustring i) != k with _ -> true) -> (None,0)
+          | Some lst -> (Some(tm::lst),k+1)
+          | None -> (None,0)) (Some [], 0) (Record.bindings r)
```
elegios: These parentheses feel a bit confusing

elegios (approved):

src/boot/mexpr.ml
david-broman: Oops. Fixed. Thanks.

src/boot/parser.mly
david-broman: Yes, we can write `()` or `{}` because they are the same thing. We have for historical reasons also been able to write `unit`. I would vote for remove `unit`, but let's discuss this at the Miking meeting. Several files use `unit` today.

src/boot/pprint.ml
david-broman: Note sure which parentheses you mean, but the last two are arguments to the `fold_left` function (you cannot see the fold in the code snippet above). It does change the meaning of the code, but I separated the fold function into a separate named function. This might be easier to read...

br4sco:
See https://github.com/david-broman/miking/pull/1 for a fix of `ext.ml`

PR #90, opened 2020-05-14
Change constructor handling in boot and MExpr

david-broman:
This PR changes the way constructors are handled in MExpr. Everything is in the implementation, and the basic use of MExpr from a syntax point of view has not changed. The implementation is so far only in boot. Before, constructors created new symbols on the fly, during evaluation. Now, a unique symbol is created statically in the `TmConaef` term. As part of the `symbolize` transformation, a new term `TmConapp` is created from a combination of a variable and a function application. That is, syntactically constructor applications are still seen as function applications, but a transformation step before evaluation makes the translation to `TmConapp`. This also means that constructors are not first class, you need to apply the constructor directly and you cannot pass it as a value. If you would like to get this effect, you can always wrap the constructor in a lambda.

As a consequence, variables and constructors live in the same name space and there is no limitation of the names (both constructors and lambda variables can have lower and upper case identifiers). Hence, I have removed #var and #con and replaced it with #ident. We can for instance write `#ident"foo"` instead of `foo`. We can also use any non-standard string as an identifier, e.g `#ident"7*/1@ a1A" is also an identifier.

### DIFF
--- a/src/boot/mexpr.ml
+++ b/src/boot/mexpr.ml
@@ -464,44 +464,44 @@ let rec symbolize (env : (ident * sym) list) (t : tm) =
     match List.assoc_opt x patEnv with
     | Some s -> (patEnv, s)
     | None -> let s = gensym() in ((x,s)::patEnv, s) in
-  let rec s_pat_sequence env patEnv pats =
+  let rec s_pat_sequence patEnv pats =
     Mseq.fold_right
-      (fun p (patEnv, ps) -> let (patEnv, p) = sPat env patEnv p in (patEnv, Mseq.cons p ps))
+      (fun p (patEnv, ps) -> let (patEnv, p) = sPat patEnv p in (patEnv, Mseq.cons p ps))
       pats
       (patEnv, Mseq.empty)
-  and sPat (env : (ident *int) list) (patEnv : (ident * int) list) = function
+  and sPat (patEnv : (ident * int) list) = function
     | PatNamed(fi,NameStr(x,_)) -> let (patEnv, s) = add_name (IdVar(sid_of_ustring x)) patEnv
                                    in (patEnv, PatNamed(fi,NameStr(x,s)))
     | PatNamed(_,NameWildcard) as pat -> (patEnv, pat)
     | PatSeqTot(fi, pats) ->
-       let (patEnv, pats) = s_pat_sequence env patEnv pats
+       let (patEnv, pats) = s_pat_sequence patEnv pats
        in (patEnv, PatSeqTot(fi, pats))
     | PatSeqEdg(fi, l, x, r) ->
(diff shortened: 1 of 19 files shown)


src/boot/mexpr.ml
```diff
@@ -434,87 +434,99 @@ let rec val_equal v1 v2 =
...
+    | None -> let s = gensym() in ((x,(IdentVar,s))::patEnv, s) in
+  let rec s_pat_sequence env patEnv pats =
     Mseq.fold_right
-      (fun p (env, ps) -> let (env, p) = sPat env p in (env, Mseq.cons p ps))
+      (fun p (patEnv, ps) -> let (patEnv, p) = sPat env patEnv p in (patEnv, Mseq.cons p ps))
       pats
-      (env, Mseq.empty)
-  and sPat patEnv = function
+      (patEnv, Mseq.empty)
+  and sPat env patEnv = function
```
elegios: Is `env` ever changed in this function? I believe it is only changed when calling `symbolize` recursively, i.e., it will be constant throughout the body here, and doesn't need to be passed as an argument.

src/boot/mlang.ml
```diff
@@ -200,7 +200,8 @@ let resolve_id {normals; _} ident =
   | Some(ident') -> ident'
   | None -> empty_mangle ident
 
-(* TODO(vipa): this function is here since the current implementation has variables and constructors in the same namespace, it should be replaced by correct uses of resolve_id and resolve_con *)
+(* TODO(vipa): this function is here since the current implementation has variables and constructors in the same namespace, it should be replaced by correct uses of resolve_id and resolve_con
+   I am not sure what this is doing now, and if we can remove it. *)
```
elegios: As mentioned during the meeting, this function does lookup with the assumption that variables and constructors live in the same namespace. `resolve_id` and `resolve_con` do lookup assuming they live in different namespaces. With the current approach we do nothing and keep `resolve_id_or_con`, but with what we talked about during the meeting each `resolve_id_or_con` and `delete_id_and_con` needs to be replaced with the corresponding specific `id` or `con` version. I can do this replacement as a PR to this PR after that change has been made.

david-broman: OK, good. I have now pushed an update that separates the name space between constructors and variables. Tests for MExpr passes, but not for MLang (most likely because of the above). Can you take a look and then create a patch to this PR? Thanks.

src/boot/mexpr.ml
david-broman: True. Fixed!

david-broman:
@br4sco Some variables start with capital letter, e.g. IDA_SUCCESS which is not allowed anymore (variables need to start with _ or lower-case letter). Can you make a patch to this PR in a way that you want it? All tests work now, except the external.

PR #91, opened 2020-05-14
Nested patterns in mlang

elegios:
This PR changes patterns in `sem`antic functions in mlang to be the normal patterns present in mexpr. For pre-existing semantic functions this should have no externally visible change, but new ones gain additional powers.

When calling a semantic function the most specific pattern that matches the argument will dictate the arm that gets executed. Note that this means that the order of the arms in the code does not matter, ordering is determined solely by specificity. Specificity is defined as subset on the set of values a pattern can match. For example, `(true, true)` is more specific than `(true, _)` since the set of matchable values is a subset.

This also creates a new failure case for merging semantic functions: if there are two patterns that overlap, but neither is more specific than the other, then the composition fails. At present there is no way to handle this failure (e.g. by manually specifying which branch to take, or adding a new pattern that matches the intersection of the overlapping patterns) but we have thoughts on how to do this in the future.

The code that performs this analysis does not have great time complexity (this is expected, checking for redundancy in pattern matches is NP-complete). However, cases we run into in practice should still be fast, since the patterns are small, but please report slowdowns.

Additionally, and unrelated, it also removes `unit` from the initial environment of values, and `PatUnit` from patterns.

### DIFF
--- a/src/boot/mlang.ml
+++ b/src/boot/mlang.ml
@@ -8,140 +8,183 @@
 open Ast
 open Msg
 open Ustring.Op
+open Pprint
 
-(*******************
- * Auxiliary stuff *
- *******************)
-let constr_compare decl1 decl2 =
-  match decl1, decl2 with
-  | CDecl(_, c1, _), CDecl(_, c2, _) -> Ustring.compare c1 c2
+open Patterns
 
-let param_eq p1 p2 =
-  match p1, p2 with
-  | Param(_,_,ty1), Param(_,_,ty2) -> ty1 = ty2
+let accum_map (f: 'acc -> 'a -> 'acc * 'b) (acc: 'acc) (l: 'a list): 'acc * 'b list =
+  let rec recur acc = function
+    | [] -> (acc, [])
+    | x::xs ->
+       let (acc, x') = f acc x in
+       let (acc, xs') = recur acc xs in
+       (acc, x'::xs')
+  in recur acc l
 
-(*********************
- * Compare patterns  *
- *********************)
-
-let pattern_compare p1 p2 = match p1, p2 with
-  | VarPattern _,         VarPattern _ -> 0
-  | VarPattern _,         ConPattern _ -> 1
-  | ConPattern _,         VarPattern _ -> -1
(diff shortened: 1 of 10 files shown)


src/boot/msg.ml
```diff
@@ -41,6 +41,11 @@ type message = id * severity * info * arguments
 exception Error of message
 
 
+let eq_info a b = match a, b with
+  | NoInfo, NoInfo -> true
+  | Info(f1, l11, c11, l21, c21), Info(f2, l12, c12, l22, c22) ->
+     l11 = l12 && c11 = c12 && l21 == l22 && c21 == c22 && f1 =. f2
```
EliasC: Is there a motivation for comparing the starting coordinates with `=` and the latter with `==`?

src/boot/mlang.ml
```diff
@@ -8,140 +8,171 @@
...
+  let inters =
+    Record.bindings inters
+    |> List.map (fun (name, inter_data) -> sort_inter name inter_data) in
+  Lang(info, name, includes, List.rev_append syns inters)
+
+let flatten_lang (prev_langs: lang_data Record.t): top -> lang_data Record.t * top = function
+  | (TopLet _ | TopRecLet _ | TopCon _) as top -> (prev_langs, top)
+  | TopLang(Lang(info, name, includes, _) as lang) ->
+     let self_data = compute_lang_data lang in
+     let included_data = List.map (lookup_lang info prev_langs) includes in
```
EliasC, line 170: This isn't really a list of data, but a list of langs, right?

EliasC (approved):
This looks OK to me. I did not get into the nitty gritty details of `patterns.ml`.

src/boot/msg.ml
elegios: No, it's a typo. Thanks.

src/boot/mlang.ml
elegios: It's a `lang_data list`. I don't particularly like the name of this type, but I didn't come up with anything better. It's supposed to be a representation of the results of the calculations I need for a language for my analyses here, so essentially caching (though I prefer to think of it as a more accurate representation, e.g., the cases are put in a DAG instead of just in a list, since the order in the list doesn't actually matter).

PR #92, opened 2020-05-14
Find strongly connected component.

br4sco:
In this PR I implement Tarjan's algorithm for finding strongly connected components in directed graphs. This is the paper: [Depth-First Search and Linear Graph Algorithms, Tarjan 72](https://doi.org/10.1137/0201010).

To implement the algorithm without mutability I use maps. Therefor, a simple and inefficient sequence based map library is also included in this PR.

The complexity of the original mutable algorithm is `O(|E|,|V|)`. If maps have `O(log(n))` insertion and lookup, this non-mutable version should be `O(log(|V|)|V|,|E|)`. However, since the maps are implemented as sequences this implementation should be `O(|V|^3,|E|)`.

### DIFF
--- a/stdlib/digraph.mc
+++ b/stdlib/digraph.mc
@@ -213,7 +213,7 @@ utest digraphCountVertices g6 with 0 in
 
 let compsEq = setEqual (setEqual eqi) in
 
-utest  compsEq (digraphStrongConnects empty) [] with true in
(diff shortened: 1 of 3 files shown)


lingmar:
Nice that you did the separate map library -- the graph implementation uses a map internally, so we could refactor it into using the map library instead.

PR #104, opened 2020-06-23
Decision points

lingmar:
This PR adds implementations related to decision points. The main file is decision-points.mc, rest is helper code.

### DIFF
--- a/stdlib/eq-paths.mc
+++ b/stdlib/eq-paths.mc
@@ -25,7 +25,6 @@ let eqPaths = lam g. lam v. lam d.
 
   recursive let traverse = lam v. lam path. lam visited. lam d.
     let toEdges = digraphEdgesTo v g in
-    --let nonSelfLoops = filter (lam e. if isSelfLoop g.eqv e then false else true) toEdges in
     let nonSelfLoops = filter (lam e. not (isSelfLoop g.eqv e)) toEdges in
 
     if isVisited g.eqv visited v then
@@ -51,12 +50,10 @@ let addEdges = lam g. lam es. foldl (lam acc. lam e. digraphAddEdge e.0 e.1 e.2 
 
 -- To compare paths
 recursive let eqpath = lam eq. lam p1. lam p2.
-  if and (null p1) (null p2) then
-    true
-  else if or (null p1) (null p2) then
-    false
-  else
-    and (eq (head p1) (head p2)) (eqpath eq (tail p1) (tail p2)) in
+  match (p1, p2) with ([], []) then true else
+  match (p1, p2) with ([], _) | (_, []) then false
+  else and (eq (head p1) (head p2)) (eqpath eq (tail p1) (tail p2))
+in
 
 -- To compare edges
 let eqedge = lam e1. lam e2.
@@ -65,10 +62,10 @@ in
 
 let samePaths = lam eq. lam seq1. lam seq2.
   and
-  -- equal number of paths
-  (eqi (length seq1) (length seq2))
-  -- each path in seq1 exists in seq2
-  (all (lam p. match (find (eqpath eq p) seq2) with Some _ then true else false) seq1)
+    -- equal number of paths
+    (eqi (length seq1) (length seq2))
+    -- each path in seq1 exists in seq2
+    (all (lam p. optionIsSome (find (eqpath eq p) seq2)) seq1)
 in
 
 -- Create some labels
--- a/stdlib/eq-paths.mc
+++ b/stdlib/eq-paths.mc
@@ -14,9 +14,9 @@ include "string.mc"
 -- TODO: prove true/false: an eq. path is never a sub-path of another eq-path
 
 
--- Complexity: O(V+E), as we explore each node and its neighbours at most
--- once (compare to DFS). This assumes digraphEdgesTo, isVisited and concat are
--- constant operations.
+-- Complexity: O(|V|^2), as for each node, we potentially visit every other
+-- node. This assumes digraphEdgesTo, isVisited and concat are constant
+-- operations.
 let eqPaths = lam g. lam v. lam d.
   let isVisited = lam eq. lam path. lam v.
     optionIsSome (find (eq v) path) in
--- a/stdlib/eq-paths.mc
+++ b/stdlib/eq-paths.mc
@@ -11,6 +11,12 @@ include "string.mc"
 -- It's really a graph algorithm, but very specific, and the test cases
 -- takes lots of space, so it's currently in a separate file from digraph.mc.
 
+-- TODO: prove true/false: an eq. path is never a sub-path of another eq-path
+
+
+-- Complexity: O(V+E), as we explore each node and its neighbours at most
+-- once (compare to DFS). This assumes digraphEdgesTo, isVisited and concat are
+-- constant operations.
 let eqPaths = lam g. lam v. lam d.
   let isVisited = lam eq. lam path. lam v.
     optionIsSome (find (eq v) path) in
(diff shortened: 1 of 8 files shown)


stdlib/eq-paths.mc
```diff
@@ -0,0 +1,363 @@
...
+
+let eqPaths = lam g. lam v. lam d.
+  let isVisited = lam eq. lam path. lam v.
+    optionIsSome (find (eq v) path) in
+
+  let isSelfLoop = lam eq. lam e. eq e.0 e.1 in
+
+  recursive let traverse = lam v. lam path. lam visited. lam d.
+    let toEdges = digraphEdgesTo v g in
+    --let nonSelfLoops = filter (lam e. if isSelfLoop g.eqv e then false else true) toEdges in
```
elegios: I guess this can be removed?

stdlib/eq-paths.mc
```diff
@@ -0,0 +1,363 @@
...
+
+mexpr
+-- To construct test graphs
+let empty = digraphEmpty eqi eqstr in
+let fromList = lam vs. foldl (lam g. lam v. digraphAddVertex v g) empty vs in
+let addEdges = lam g. lam es. foldl (lam acc. lam e. digraphAddEdge e.0 e.1 e.2 acc) g es in
+
+-- To compare paths
+recursive let eqpath = lam eq. lam p1. lam p2.
+  if and (null p1) (null p2) then
```
elegios: This could be done with matches

stdlib/eq-paths.mc
```diff
@@ -0,0 +1,363 @@
...
+let eqedge = lam e1. lam e2.
+  eqstr e1 e2
+in
+
+let samePaths = lam eq. lam seq1. lam seq2.
+  and
+  -- equal number of paths
+  (eqi (length seq1) (length seq2))
+  -- each path in seq1 exists in seq2
+  (all (lam p. match (find (eqpath eq p) seq2) with Some _ then true else false) seq1)
```
elegios: `optionIsSome` I guess?

stdlib/eq-paths.mc
```diff
@@ -0,0 +1,363 @@
...
+-- points.
+
+-- Input: a graph, a start node v and a depth d
+-- Output: the set of equivalence paths ending in v. The lengths of the
+-- paths are at most d.
+
+-- It's really a graph algorithm, but very specific, and the test cases
+-- takes lots of space, so it's currently in a separate file from digraph.mc.
+
+let eqPaths = lam g. lam v. lam d.
```
elegios, line 20: I find it interesting that direct and indirect recursion are treated differently here, the latter puts a cap on how far back we can go when discovering a path, while the former is simply ignored and passed over (i.e., no cap). Omitting the indirect cycle without introducing a cap seems more consistent, but it makes the representative path rather different when you increase `d` enough to cover the entire cycle.

stdlib/mexpr/decision-points.mc
```diff
@@ -0,0 +1,913 @@
...
+  sem pprintCode (indent : Int) =
+  | TmHole h -> strJoin "" ["Hole (", getTypeStringCode indent h.tp, ", ",
+                                      pprintCode indent h.startGuess, ", ",
+                                      pprintCode indent h.depth, ")"]
+end
+
+-- Temporary workaround: uniquely labeled decision points
+lang LHoleAst = HoleAst
+  syn Expr =
+  | LTmHole {tp : Type,
```
elegios, line 33: This should probably be a product extension, once those land

stdlib/mexpr/decision-points.mc
```diff
@@ -0,0 +1,913 @@
...
+    sfold_Expr_Expr digraphUnion cgraph (smap_Expr_Expr (findEdges cgraph prev) tm)
+end
+
+
+lang ContextAwareHoles = Ast2CallGraph + LHoleAst + IntAst + SymbAst
+  -- Transform the program
+  sem transform =
+  | tm ->
+    -- Define hard coded constants
+    let callCtxVar = "callCtx" in
```
elegios, line 182: I assume that these mean that this transformation will at present potentially break if any of these identifiers are used otherwise? (probably fine for the moment)

stdlib/mexpr/decision-points.mc
```diff
@@ -0,0 +1,913 @@
...
+    let handleLetExpr = lam le.
+      if optionIsSome (find (eqstr le.ident) fs) then
+        match le.body with TmLam lm then
+          let newBody = lam_ "callCtx"  (None ())
+                             (TmLam {lm with body = transform2 fs le.ident lm.body})
+          in {le with body = newBody}
+        else error "Internal error: this letexpr should have a TmLam in its body"
+      else le
+    in
+    -- FIXME: Really do it for all bindings, not just the first
```
elegios: Might be worth throwing an error if there is more than one binding?

elegios (approved):

stdlib/mexpr/decision-points.mc
lingmar: Sorry, this was fixed this morning by a commit I forgot to push!

stdlib/mexpr/decision-points.mc
lingmar: Correct. I will add a comment about it.

stdlib/mexpr/decision-points.mc
lingmar: Agree.

stdlib/eq-paths.mc
lingmar: Thanks, fixed!

stdlib/eq-paths.mc
lingmar: Thanks, fixed!

stdlib/eq-paths.mc
lingmar: Thanks, fixed!

stdlib/eq-paths.mc
lingmar: [I tried to paste ascii versions of the dot graphs below but Github didn't render them readable]

You are right that it's inconsistent in a sense, as the following graph g1 with direct recursion:
digraph {
    1->2[label=l1];
    2->2[label=l2];
    2->4[label=l4];
}
would give eqpaths g1 4 4 = {[l1, l4]} (source 4, depth 4)
while this similar graph g2 with indirect recusion 
digraph {
    1->2[label=l1];
    2->3[label=l2];
    3->2[label=l3]
    2->4[label=l4];
}
would give eqpaths g2 4 4 = {[l3, l4], [l1,l4]}. 
To make it consistent perhaps the result should be just [l1,l4] in the latter case as well.

The source of the inconsistency above was to be able to distinguish between "in a mutual recursion" and "not in a mutual recursion" for nodes in a mutual recursion. That is, if node 2 above is the source, we wanted:
eqpaths g2 2 1 = {[l1],[l3]}
eqpaths g2 2 2 = {[l1],[l3]} 
But now I don't feel sure if this makes sense. Perhaps always cutting the cycle if it is completed is the best as it is the most consistent way. In that way, we would have:
eqpaths g2 2 1 = {[l1],[l3]}
eqpaths g2 2 2 = {[l1]}

lingmar: g1
g1
g2
g2

lingmar: Sorry, I got confused and missed some paths in the examples above. It has been updated now.

elegios: [I'm guessing you could put the ascii graphs in a code block to render appropriately, but I'm not sure]

I really have no idea which way it should be. Maybe it's ok that the number of equivalence classes can decrease (which I would otherwise consider as surprising) as the depth gets greater, as we discover new mutual recursions? Either way that's probably a question for later.

PR #107, opened 2020-06-30
Minor updates to README.md, including added info about how to run utest.

david-broman:


README.md
```diff
@@ -78,7 +78,7 @@ MCore consists of two parts:
 
 ## MExpr
 
-One design objective of MExpr is to make the concrete syntax very close to the abstract syntax of the language. That is, no syntactic sugar is introduced to the concrete MCore syntax. The MExpr language is not intended to be a general purpose programming language, but a core language to which other languages translate into.
+One design objective of MExpr is to make the concrete syntax very close to the abstract syntax of the language. That is, no syntactic sugar is introduced to the concrete MCore syntax. The MExpr language is not intended to be a general purpose programming language, but as a core language to which other languages translate into.
```
dlunde: "but a core language" sounds better than "but as a core language"

README.md
```diff
@@ -92,7 +92,14 @@ When writing MCore programs, it is typically done by writing explicit unit tests
 utest addi 1 2 with 3 in
 ()
 ```
-checkes that the addition of `1` and `2` is in fact `3`. Typically when you develop MCore programs, you do not use the `print` function. Instead, you write unit tests directly, and then leave the units tests as is directly after your function. By doing so, you test your code, write regression tests, and document the informal semantics of your program directly. We strongly encourage you to develop your MCore programs this way.
+checkes that the addition of `1` and `2` is in fact `3`. To run the tests in an `.mc` file, run the `boot` command with argument `test`:
```
dlunde: "checkes" -> "checks"

dlunde (commented):
Looked through changes.

README.md
david-broman: Well, I disagree. Anyway, I have now reformulated the sentence by breaking it up into two sentences instead.

README.md
david-broman: Fixed, thanks!

PR #108, opened 2020-06-30
Refactor option.mc

aathn:
This pull request refactors `option.mc`.

- Fix parameter orders: `optionMap` and `optionFilter` now take the function first, and `optionMapOr` and co. take the default value first. This enables smoother composition of these functions and conforms with convention in other languages.
- Rename `optionFlatten` to `optionJoin` and `optionFlatMap` to `optionBind` for consistency with the rest of the stdlib.
- Implement functions in terms of each other.
- Add utility functions `optionGetOrElse` and `optionGetOr` (with `optionGetOrElse f o` corresponding to `optionMapOrElse f identity o`).

These changes may be breaking for users of the library.

### DIFF
--- a/stdlib/map.mc
+++ b/stdlib/map.mc
@@ -14,9 +14,8 @@ let mapLookupOpt = lam eq. lam k. lam seq. findAssoc (eq k) seq
 -- Value binded to k in seq, if no such binding in seq then an error occurs.
 -- Equality defined by eq.
 let mapLookup = lam eq. lam k. lam seq.
-  optionMapOrElse (mapLookupOpt eq k seq)
-                  (lam _. error "Element not found")
-                  (lam x. x)
+  optionGetOrElse (lam _. error "Element not found")
+                  (mapLookupOpt eq k seq)
 
 -- True if k is bound in seq, else false.
 -- Equality defined by eq.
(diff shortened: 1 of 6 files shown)


stdlib/map.mc
```diff
@@ -14,9 +14,8 @@ let mapLookupOpt = lam eq. lam k. lam seq. findAssoc (eq k) seq
 -- Value binded to k in seq, if no such binding in seq then an error occurs.
 -- Equality defined by eq.
 let mapLookup = lam eq. lam k. lam seq.
-  optionMapOrElse (mapLookupOpt eq k seq)
-                  (lam _. error "Element not found")
-                  (lam x. x)
+  optionGetOrElse (lam _. error "Element not found")
```
johnwikman, line 17: The changes looks good to me! (Also verified that the tests succeed after a merge with develop.)

More of a philosophical discussion: Could it be an idea to adopt a standard of how to handle error prints? At the moment (and previously) the error just prints out that something went wrong, but no details on what actually went wrong. In this case we could for example print out which key that was missing from the map:
```ocaml
let mapLookup = lam eq. lam k. lam seq.
  optionGetOrElse (lam _. error (strJoin " " ["Element bound to key \"", k, "\" was not found"]))
                  (mapLookupOpt eq k seq)
```

Or could we perhaps adopt a formatted error print? Such as:
```ocaml
let mapLookup = lam eq. lam k. lam seq.
  optionGetOrElse (lam _. errorFmt "Element bound to key \"$0\" was not found" [k])
                  (mapLookupOpt eq k seq)
```

PR #113, opened 2020-07-01
Feature: REPL

aathn:
This pull request adds a simple REPL to the bootstrap interpreter.

### DIFF
--- a/src/boot/mexpr.ml
+++ b/src/boot/mexpr.ml
@@ -523,7 +523,12 @@ let rec symbolize (env : (ident * sym) list) (t : tm) =
     -> TmUtest(fi,symbolize env t1,symbolize env t2,symbolize env tnext)
   | TmNever(_) -> t
 
-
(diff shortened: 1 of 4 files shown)


src/boot/mexpr.ml
```diff
@@ -523,7 +523,23 @@ let rec symbolize (env : (ident * sym) list) (t : tm) =
     -> TmUtest(fi,symbolize env t1,symbolize env t2,symbolize env tnext)
   | TmNever(_) -> t
 
-
+(* Same as symbolize, but records all toplevel definitions and returns them
```
david-broman, line 526: Function `symbolize_toplevel` should perhaps be moved to `mlang.ml` since it is more related to MLang than MExpr.

david-broman (commented):
I think it looks great! 

The next step is to improve the REPL user experience (remove the need for `;;`, add history etc.).

PR #118, opened 2020-07-09
Python intrinsics

aathn:
This PR adds Python intrinsics as an optional feature to the bootstrap interpreter, allowing Python functions to be called from MCore code. The Python intrinsics are structured in the same way as the sundials externals, allowing to compile without them.

In addition, a singleton tuple syntax is added (to enable conveniently passing single arguments to Python functions) and the makefile is updated to allow selecting which externals to compile with support for.

Note that this PR changes the command for compiling and testing with sundials to `MI_ENABLE_SUNDIALS=1 make test`.


theoahfeldt:
I just noticed this PR reverts the name change to mi. We will fix this ASAP.

README.md
```diff
@@ -795,15 +799,59 @@ opam install sundialsml
...
+
+The following example shows how to use the intrinsics to to plot an MCore
+sequence using `matplotlib` (assuming it's installed) after sorting it.
+
+```
+let builtins = pyimport "builtins"
+let plt = pyimport "matplotlib.pyplot"
+
+let x = [5.,4.5,4.,1.,1.5]
+let y = pycall builtins "sorted" (x,)
```
elegios, line 842: There might be a point to mention that you can't get values back into mcore. Not sure though, since I got the impression something like it would be added pretty soon.

make
```diff
@@ -42,44 +79,30 @@ runtests() {
...
+    fi
+    if [[ -n $MI_ENABLE_SUNDIALS ]]; then
+        runtests_ext
+    fi
+    )
 }
 
 case $1 in
-    # with external functions integration
-    externals-test)
```
elegios, line 59: Mentioning this for completeness since it was mentioned during the meeting, it'd be good to add a `make test-all` or similar that includes all supported externals and runs all tests.

src/boot/parser.mly
```diff
@@ -289,6 +289,8 @@ atom:
   | LPAREN seq RPAREN
       { if List.length $2 = 1 then List.hd $2
         else tuple2record (mkinfo $1.i $3.i) $2 }
+  | LPAREN mexpr COMMA RPAREN
```
elegios, line 292: I don't see a corresponding change for `pprint`, how are singleton tuples printed there?

src/boot/ast.ml
```diff
@@ -109,6 +109,7 @@ and const =
 | Ceqs of int option
 (* External functions TODO: Should not be part of core language *)
 | CExt of tm Extast.ext
+| CPy of Pyast.py
```
elegios: It might be nice to structure both externals in the same way, more or less to the point that both could have identical `.mli` files. Not sure it's worth it to actually create such files, but making all externally used names the same between them is probably nice for consistency.

src/boot/py/pyffi.ml
```diff
@@ -0,0 +1,68 @@
...
+  | TmConst(_,CBool(v)) -> Py.Bool.of_bool v
+  | TmConst(_,CInt(v)) -> Py.Int.of_int v
+  | TmConst(_,CFloat(v)) -> Py.Float.of_float v
+  | TmConst(_,CPy(PyObject(v))) -> v
+  | TmSeq(_,s) ->
+    begin
+      try tmseq2ustring fi s |> Ustring.to_utf8 |> Py.String.of_string
+      with _ -> Mseq.to_list s |> Py.List.of_list_map (val_to_python fi)
+    end
+  (* | TmRecord(_,r1) ->
```
elegios: I think you can just remove these two. The alternative might be to create dicts (or whatever python calls them, I forget), but maybe we add that once we have a use for it?

elegios (approved):

README.md
aathn: Done!

make
aathn: This is added now.

src/boot/parser.mly
aathn: Good catch! We updated the pretty printer now.

src/boot/ast.ml
aathn: Fixed.

src/boot/py/pyffi.ml
aathn: Removed for now.

PR #120, opened 2020-07-16
Jupyter kernel

aathn:
This is a large pull request adding a fully featured Jupyter kernel based on the boot interpreter. The kernel is completely optional and is described in `KERNEL_README.md`. The kernel supports a large amount of interoperability between Python and MCore, such as executing Python code directly and calling functions defined in Python cells from MCore cells. It is also able to display plots made by `matplotlib` inline in notebooks. The kernel includes autocompletion and we also bundle a notebook extension for MCore syntax highlighting.

In addition, the pull request adds the following features, which are related to the new kernel and its features.
- Conversion from python values back to MCore values using a new builtin `pyconvert`
- Autocompletion and persistent history in the REPL

### DIFF
--- a/KERNEL_README.md
+++ b/KERNEL_README.md
@@ -54,16 +54,43 @@ and select the `MCore` kernel when creating a new notebook.
 *Note that `$HOME/.local/bin` must be included in your PATH. This should be
 done by default on most Linux distributions.*
 
+## Now you're playing with Python
+
+The MCore kernel also allows executing Python code and interacting with it from
+MCore. Use the special directive `%%python` at the top of a cell to evaluate
+Python code.
+
+You can call the functions you have defined in Python cells in normal MCore
+cells by importing and using the module `__main__`. For example, consider the
+following two cells:
+
+```python
+%%python
+def foo(x):
+  print("foo"+x)
+```
+
+```ocaml
+let x = "bar" in
+let m = pyimport "__main__" in
+pycall m "foo" (x,)
+```
+
+The final `pycall` will print `foobar` as expected.
+
 ## Plotting graphs
 
-It is possible to plot graphs with MCore using the Python library `matplotlib`.
+It is possible to plot graphs using the Python library `matplotlib`.
 The Jupyter kernel offers integration with `matplotlib` to display plots inline.
 Make sure you have `matplotlib` installed (if not, you can install it using
 `pip`). Now, when you use `matplotlib`'s plot functions in a notebook cell, the
-plots will be displayed as part of the cell's output. For example, you can try running the following in a cell:
+plots will be displayed as part of the cell's output.
+For example, you can try running the following in a cell:
 
 ```ocaml
 let plt = pyimport "matplotlib.pyplot"
 let x = [1,2,4,8]
 let _ = pycall plt "plot" (x,)
 ```
+
+Of course, the same goes for plots made using Python cells.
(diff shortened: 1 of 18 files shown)


capitanu (commented):
Tested it on my system and skimmed through the code. Works really well!

lingmar:
Works fine for me too (using macOs)! Is there a way of configuring which version/installation/environment of Python the kernel uses? Perhaps that is a question on how I configure jupyter.

aathn:
The kernel uses the [`pyml`](https://github.com/thierry-martinez/pyml) Python bindings to execute Python code. `pyml` uses the `libpython` shared library under the hood, which it finds it at runtime by trying a number of different methods and selecting the first that works; this is what determines the version of Python used. I believe the first thing it tries is essentially to use `which python3` to find a bin folder and look in `bin-folder/../lib/`, which should result in the same version as that of the `python3` interpreter available in `PATH` in most cases.

I'm not entirely sure of the details though, especially for `pyvenv`. Also, we've configured `pyml` to always use Python 3, so the kernel does not work with Python 2; this is mainly because Python 3 has a `builtins` module which allows us to treat builtins as any other functions, whereas in Python 2 the builtins would require special treatment.

KERNEL_README.md
```diff
@@ -0,0 +1,156 @@
...
+## Getting started
+
+Before using the Jupyter kernel you need to install some dependencies.
+The Jupyter kernel requires all the base dependencies of the boot interpreter
+(`dune`, `batteries`, and `linenoise`),
+and the `pyml` Python bindings for OCaml. For information on how to install
+these, see [README.md](./README.md).
+
+Next, you will need to have Jupyter itself installed.
+To install Jupyter using using `pip`, run the following command:
```
andnil5: Typo "using using"

KERNEL_README.md
```diff
@@ -0,0 +1,156 @@
...
+## Jupyter basics
+
+[Jupyter](https://jupyter.org/) provides an ecosystem for writing, documenting
+and visualizing code in an interactive way. The _Jupyter Notebook_ is a
+literate programming environment for this purpose. Notebooks can contain text,
+executable code, display images and more. Notebooks provide support for many languages
+by using language-specific _kernels_, which take care of executing the user's
+code and producing output for the notebook to display.
+
+To get a quick introducion on how to use Jupyter Notebooks in general, check out
```
andnil5: typo "introducion"

andnil5 (approved):
Works fine for me too!

moaand:
I looked through it and it works well for me too!

david-broman:
Note that this PR has conflicts. Please fix and rebase!

david-broman:
I do not think that the file `MCoreJupyter.ipynb` should be in the root directory. I would suggest to move the file into a new folder `examples/jupyter/`.

david-broman:
I am not sure about having the Jupyter notebook  to include the readme.md file content. It will unfortunately make it really hard to keep it consistent. I think that this nodebook should be updated and just focus on some of the core examples in MCore and then highlight all aspects of interacting with Python etc.

david-broman:
In the notebook, it says "Most basic values can be converted between Python and MCore types;". I think it would be good if you demonstrate all values that work, as examples in the notebook. Also, explain, what does not work (if there is anything more than classes)
.

aathn:
Closing this, as discussed in meeting. Instead, two new pull requests will be made, one to this repo and one to the `miking-jupyter` repo.

PR #121, opened 2020-07-19
Top level utests

aathn:
This adds MLang top level utests, and moves some of the utests in the standard library to be next to the function they are testing.
Note that using top level utests is not quite as convenient as mexpr utests, as you cannot reuse a value across several utests without defining it in the top level.

### DIFF
--- a/stdlib/string.mc
+++ b/stdlib/string.mc
@@ -33,6 +33,11 @@ let string2int = lam s.
   then negi (string2int_rechelper (tail s))
   else string2int_rechelper s
 
+utest string2int "5" with 5
+utest string2int "25" with 25
+utest string2int "314159" with 314159
+utest string2int "-314159" with (negi 314159)
+
 let int2string = lam n.
   recursive
   let int2string_rechelper = lam n.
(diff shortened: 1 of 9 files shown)


stdlib/string.mc
```diff
@@ -159,50 +197,7 @@ recursive
...
-utest strJoin "--" ["water", "tea", "coffee"] with "water--tea--coffee" in
-utest strJoin "--" [] with "" in
-utest strJoin "--" ["coffee"] with "coffee" in
-utest strJoin "water" ["coffee", "tea"] with "coffeewatertea" in
-
-()
+utest strJoin "--" ["water", "tea", "coffee"] with "water--tea--coffee"
+utest strJoin "--" [] with ""
+utest strJoin "--" ["coffee"] with "coffee"
+utest strJoin "water" ["coffee", "tea"] with "coffeewatertea"
```
andnil5, line 203: Like! A possible downside with the changes to the stdlib files is that it adds a kind of inconsistency of utest usage in stdlib (top level/not top level).

david-broman:
Looks good!

PR #122, opened 2020-07-19
Constructors as functions

aathn:
This pull request makes data constructors behave like first-class functions. This enables code such as 

```ocaml
map Foo [1,2,3]
```

where before, you had to use

```ocaml
map (lam x. Foo x) [1,2,3]
```

I don't know whether this change is wanted, but this is something which has been annoying me, and I don't see any particular downsides in changing it.

### DIFF
--- a/src/boot/ast.ml
+++ b/src/boot/ast.ml
@@ -143,7 +143,7 @@ and tm =
 | TmRecord  of info * tm Record.t                                   (* Record *)
 | TmRecordUpdate of info * tm * ustring * tm                        (* Record update *)
 | TmCondef  of info * ustring * sym * ty * tm                       (* Constructor definition *)
-| TmConapp  of info * ustring * sym * tm                            (* Constructor application *)
+| TmConapp  of info * ustring * sym * tm option                     (* Constructor application *)
 | TmMatch   of info * tm * pat * tm * tm                            (* Match data *)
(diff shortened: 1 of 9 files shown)


andnil5:
Make sense to me, like!

david-broman:
We have discussed this a lot and there are several reasons for not having first-class data constructors in MExpr. First, note that MExpr is an intermediate language and not intended to be a language that people write large programs in. Instead, new languages should be created on top of MExpr. These languages can have first-class data constructors and this can easily be implemented as syntactic sugar on top of MExpr by just wrapping the data constructor in a lambda. 

The main reason for not including it directly in MExpr (as you have suggested in this PR) is for performance reasons. Adding it to MExpr is easy to do in an interpreter: just make a check in the eval function when performing an application. However, if this is translated directly into compiled code, data constructors and lambdas need to be tagged, to check which kind of application that should be applied. This will give severe performance penalties. It might be possible to perform static analysis to revert back, but it becomes unnecessarily complicated. Hence, we prefer to keep MExpr simple to enable efficient compilation, and add the first-class property (which is really nice in a high-level programming language) as syntactic sugar in high-level languages built on top of MExpr. For instance, Ragnar is planned to have first-class data constructors.

aathn:
I see, interesting! I'll close this PR, then.

PR #129, opened 2020-08-05
Add HashMap to stdlib

johnwikman:
Adds a simple hashmap to the standard library. The measured time for the tests at the bottom of `hashmap.mc` on my machine was ~~>60s~~ ~60s for the `map.mc` association list and <5s for the hashmap. Not as much of a speedup as it probably should be, but the hashmap is at least noticably faster than the association list for when the map is large.

Update: Output from isolated timing measurements (Top is hashmap, bottom is assoc list)
```
john@obsidian:~/GitRepos/Forks/miking (hashmap)$ time mi test stdlib/hashmap.mc 
/home/john/GitRepos/Forks/miking/stdlib/hashmap.mc: ........................................................................................................................................................................................................................................................................................................................................................................................................................................................ OK
Unit testing SUCCESSFUL after executing 440 tests.


real	0m1.281s
user	0m1.272s
sys	0m0.007s
john@obsidian:~/GitRepos/Forks/miking (hashmap)$ time mi test stdlib/hashmap.mc 
/home/john/GitRepos/Forks/miking/stdlib/hashmap.mc: ....................................................................................................................................................................................................................................................................................................................................................................................................................................................... OK
Unit testing SUCCESSFUL after executing 439 tests.


real	0m58.332s
user	0m58.264s
sys	0m0.000s
```

### DIFF
--- a/stdlib/hashmap.mc
+++ b/stdlib/hashmap.mc
@@ -10,8 +10,6 @@ include "math.mc"
 include "option.mc"
 include "string.mc"
 
-include "map.mc" -- REMOVE BEFORE MERGE (only used for comparison)
-
 let hashmapDefaultBucketCount = 100
 
 -- The base type of a HashMap object.
@@ -223,37 +221,5 @@ in
 let m = removeall 0 m in
 
 utest m.nelems with 0 in
-
-
-
--- BEGIN TEMPORARY TIMING TEST (Remove before merge)
---let lookupOpt = mapLookupOpt eqstr in
---let lookup = mapLookup eqstr in
---let insert = mapInsert eqstr in
---let update = mapUpdate eqstr in
---let mem = mapMem eqstr in
---
---recursive let populate = lam m. lam i.
---  if geqi i n then
---    m
---  else
---    let key = cons 'a' (int2string i) in
---    populate (insert key i m)
---             (addi i 1)
---in
---let m = populate [] 0 in
---
---utest length m with n in
---
---recursive let checkmem = lam i.
---  if geqi i n then
---    ()
---  else
---    let key = cons 'a' (int2string i) in
---    utest lookupOpt key m with Some (i) in
---    checkmem (addi i 1)
---in
---let _ = checkmem 0 in
--- END TEMPORARY TIMING TEST
 
 ()
--- a/stdlib/hashmap.mc
+++ b/stdlib/hashmap.mc
@@ -193,6 +193,7 @@ recursive let populate = lam hm. lam i.
     hm
   else
     let key = cons 'a' (int2string i) in
+    utest hashmapLookupOpt key hm with None () in
     populate (hashmapInsert key i hm)
              (addi i 1)
 in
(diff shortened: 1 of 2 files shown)


stdlib/hashmap.mc
```diff
@@ -0,0 +1,225 @@
...
+
+
+-- Returns an empty hashmap with a default number of buckets.
+--   eqArg: A function that specifies equalities between keys in the hashmap.
+--   hashfnArg: A function for computing the hash of a key value.
+let hashmapEmpty = lam eqArg. lam hashfnArg.
+  {buckets = makeSeq hashmapDefaultBucketCount [],
+   nelems = 0,
+   eq = eqArg,
+   hashfn = hashfnArg}
```
aathn: This is a matter of preference, but I would probably drop the `Arg` suffix in these parameters. I would prefer just having `eq` and `hashfn` even if it collides with the record field names.

stdlib/hashmap.mc
```diff
@@ -0,0 +1,225 @@
...
+  let idx = modi (absi hash) (length hm.buckets) in
+  let bucket = get hm.buckets idx in
+  let newEntry = {hash = hash, key = key, value = value} in
+  recursive let inserter = lam seq.
+    if null seq then
+      [newEntry]
+    else
+      let entry = head seq in
+      if neqi hash entry.hash then
+        cons entry (inserter (tail seq))
```
aathn, line 66: This case doesn't impact the result of the code, is there a reason for having it? I'm guessing it's for performance?

My comment applies to `hashmapRemove` and `hashmapLookupOpt` as well, which also have a similar structure.

stdlib/hashmap.mc
```diff
@@ -0,0 +1,225 @@
...
+-- Inserts a value into the hashmap
+--   key: The key to bind the value to.
+--   value: The value to be inserted.
+--   hm: The hashmap to insert the value into.
+-- [NOTE]
+--   The insertion uses a recursion which is not tail-recursive.
+let hashmapInsert = lam key. lam value. lam hm.
+  let hash = hm.hashfn key in
+  let idx = modi (absi hash) (length hm.buckets) in
+  let bucket = get hm.buckets idx in
```
aathn: How about a `hashmapGetBucket : HashMap -> Hash -> Bucket` for performing this? I think it's good to isolate low-level calculations to smaller functions, and it would reduce some code duplication across `hashmapRemove` and `hashmapLookupOpt` as well.

aathn (commented):
Looks good I think!

stdlib/hashmap.mc
johnwikman: Yes. This if statement is introduced under the assumption that the equality function is slow and that the hash function is good. Not sure about how this affects the cumulative execution time of the hashmap, but this will hopefully keep the variance down.

stdlib/hashmap.mc
johnwikman: I fully agree with this, and it was my original intention to lift these out as common definitions. However, I am not keen on exposing the inner workings of the hashmap to outside files, so I would like to keep these operations private.

This is probably a good discussion for the meeting.

stdlib/hashmap.mc
johnwikman: Lets discuss this at the meeting! I have no strong opinion on this.

PR #137, opened 2020-08-11
Add random number generation intrinsics

lingmar:
This PR adds an intrinsic function `randIntU` to boot, such that ~`randInt b`~ `randIntU a b` returns a random number between ~0~`a` (inclusive) and `b` (exclusive). 

Edit: the intrinsic `randSetSeed s` seeds the random generator, so that the same seed always generates the same sequence of numbers. If `randSetSeed` is not called, `randIntU` sets a seed automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ containing `utest` expressions. Please see the following files:
 
 * [Symbol intrinsics](test/mexpr/symbs.mc)
 
+* [Random number generation intrinsics](test/mexpr/random.mc)
+
 Besides the intrinsic functions, the prelude includes a number of functions defined in the MCore standard library, which can be found in the folder [stdlib](stdlib/). The main file [prelude.mc](stdlib/prelude.mc) is automatically included in all `.mc` files. Note also that the prelude file includes other files, e.g., [seq.mc](stdlib/seq.mc) and [option.mc](stdlib/option.mc). For the details of these prelude functions, please see the above files.
 
 
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -97,6 +97,7 @@ and const =
 | Creverse
 (* MCore intrinsic: random numbers *)
 | CrandIntU of int option
+| CrandSetSeed
 (* MCore debug and I/O intrinsics *)
 | Cprint
 | Cdprint
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -96,7 +96,7 @@ and const =
 | CsplitAt of (tm Mseq.t) option
 | Creverse
 (* MCore intrinsic: random numbers *)
-| CrandInt
+| CrandIntU
 (* MCore debug and I/O intrinsics *)
 | Cprint
 | Cdprint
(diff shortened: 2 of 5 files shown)


johnwikman:
Looks good to me! A couple of design comments from my end:

- What distribution is this generating from? I assume that it is uniformly from the interval `[0,b)`, but it would be nice to include a hint about the distribution in the name itself. Something like `randIntU` or `urandInt`? Although many languages just use the word `rand` in their name for generating uniformly random integers, we probably want to specify the distribution when generating random floats (gaussian vs. uniformly random). In that case we probably want something like `randFloatU` and `randFloatG`. In my opinion, it would be nice to have a clear and consistent naming scheme between types for random number generation.
- A personal preference on my part is to have explicit bounds on random number generation. Such that if I want to generate a number in the interval `[a,b)` I would write `randInt a b` instead of the single argument approach `addi a (randInt (subi b a))`. Just a preference on my part.
- It would be nice to have a seed function (e.g. `randSetSeed : Int -> ()`), such that it is possible to get reproducible results when using random number generation.

lingmar:
> Looks good to me! A couple of design comments from my end:
> 
>     * What distribution is this generating from? I assume that it is uniformly from the interval `[0,b)`, but it would be nice to include a hint about the distribution in the name itself. Something like `randIntU` or `urandInt`? Although many languages just use the word `rand` in their name for generating uniformly random integers, we probably want to specify the distribution when generating random floats (gaussian vs. uniformly random). In that case we probably want something like `randFloatU` and `randFloatG`. In my opinion, it would be nice to have a clear and consistent naming scheme between types for random number generation.
Agree, I will go with `randIntU` (in that way, all random functions can start with rand).
> 
>     * A personal preference on my part is to have explicit bounds on random number generation. Such that if I want to generate a number in the interval `[a,b)` I would write `randInt a b` instead of the single argument approach `addi a (randInt (subi b a))`. Just a preference on my part.
Agree, I will make it take two arguments.
> 
>     * It would be nice to have a seed function (e.g. `randSetSeed : Int -> ()`), such that it is possible to get reproducible results when using random number generation.
I was considering this but was originally against it, partly to add as few intrinsics as possible, and partly to keep the interface as purely functional as possible, as setting the seed is a side effect, as it globally affects the behaviour of the rest of the program.  However, I agree it's desirable to be able to get reproducible results, so I'll add it.  I'll make it optional though so that the program self-initiates the seed in case it has not been set by the user.

johnwikman:
> I was considering this but was originally against it, partly to add as few intrinsics as possible, and partly to keep the interface as purely functional as possible, as setting the seed is a side effect, as it globally affects the behaviour of the rest of the program. However, I agree it's desirable to be able to get reproducible results, so I'll add it. I'll make it optional though so that the program self-initiates the seed in case it has not been set by the user.

Well, the current solution is not completely functional either as the same invocation does not necessarily return the same result? A purely functional random function would in this case probably be `randInt : Int -> RngState -> (RngState, Int)`, though that would be very inconvenient to use. But I agree that it is preferable to keep the number of intrinsics down, and I am also unsure about our stance on which side effects to tolerate.

lingmar:
> 
> Well, the current solution is not completely functional either as the same invocation does not necessarily return the same result? A purely functional random function would in this case probably be `randInt : Int -> RngState -> (RngState, Int)`, though that would be very inconvenient to use. But I agree that it is preferable to keep the number of intrinsics down, and I am also unsure about our stance on which side effects to tolerate.

We can discuss this during the meeting. I added a seeding intrinsic for now.

PR #139, opened 2020-08-12
Remove bash style syntax from make script

johnwikman:
Removes the bash style syntax on the if-statements for checking whether to run the sundials and/or python tests.

### DIFF
--- a/make
+++ b/make
@@ -55,10 +55,10 @@ runtests() {
     cd ..
     export MCORE_STDLIB='@@@'
     build/mi test stdlib)
-    if [[ -n $MI_TEST_PYTHON ]]; then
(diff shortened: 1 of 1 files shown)


aathn:
Thanks for fixing this, my miss from PR #133.

We could just change the shebang to `#!/usr/bin/env bash` if we want to avoid further problems like this.

johnwikman:
I think that it was good that we changed it from /bin/bash to /bin/sh. We get one less dependency and we didn't really use any bash exclusive functionality in our make script. IMO we should try to reduce the number of dependencies until it becomes an inconvenience to avoid them.

david-broman:
@johnwikman I agree. The new PR still works on my macOS Catalina.

PR #141, opened 2020-08-13
Add time measurements intrinsics

lingmar:
This PR adds the intrinsic functions `wallTimeMs : Unit -> Int` and `sleepMs : Int -> Unit` (the latter added mainly to test the former), for measuring elapsed time and pausing the execution, respectively.

The elapsed wall-clock time in ms is calculated as the difference between two calls to `wallTimeMs`:
```
let t1 = wallTimeMs () in
...
let t2 = wallTimeMs () in
-- elapsed time in ms
subi t2 t1
```

Dependencies: This PR adds an additional dependency package: [Mtime](https://opam.ocaml.org/packages/mtime/). I chose to use this package even though there are two other options that don't require additional dependencies because:
- `Unix.gettimeofday`: exact precision is not specified, [the documentation](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html) says only "resolution better than 1 second". Mtime, on the other hand, guarantees nano-second precision.
- `Sys.time`: returns the CPU time. We could use this if we prefer to measure CPU time instead of wall-clock time (e.g. `procTimeMs` could replace, or be added alongside with, `wallTimeMs`).

Precision: I chose ms precision even though Mtime has ns precision, as I find ns to be a bit overkill for timing, and I'm not sure whether ns precision is common in other target platforms. However, perhaps we can discuss which precision we prefer, e.g. micro-seconds can be an option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ Before you can use the Miking system, you need to install
 [OPAM](https://opam.ocaml.org/) package manager.
 
 
-After the installation, you need to install the `opam` packages `dune`, `batteries`, `linenoise`, and `mtime` by running the following:
+After the installation, you need to install the `opam` packages `dune`, `batteries`, and `linenoise` by running the following:
 
 ```
-opam install dune batteries linenoise mtime
+opam install dune batteries linenoise
 ```
 
 To compile and run the test suite, execute:
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ containing `utest` expressions. Please see the following files:
 
 * [Random number generation intrinsics](test/mexpr/random.mc)
 
+* [Time intrinsics](test/mexpr/time.mc)
+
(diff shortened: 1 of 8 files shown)


johnwikman:
Found this issue on the mtime GitHub site regarding Windows support: https://github.com/dbuenzli/mtime/issues/2

(Not tested the PR myself yet.)

johnwikman:
Also, it appears that [Unix.gettimeofday](https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/unix.ml#L442) is microsecond precision: https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/gettimeofday.c

Though as you pointed out, the documentation provides no guarantee that this must be the case in the future.

lingmar:
> Found this issue on the mtime GitHub site regarding Windows support: [dbuenzli/mtime#2](https://github.com/dbuenzli/mtime/issues/2)
> 
> (Not tested the PR myself yet.)

Thanks! To be honest I don't completely understand the issue from reading the thread, and if they already fixed it or not.

PR #143, opened 2020-08-14
Added a few more functions to seq.mc

david-broman:
Note: in the future, I think we should add more documentation to all functions in the standard library. See this PR as an example.

### DIFF
--- a/stdlib/seq.mc
+++ b/stdlib/seq.mc
@@ -145,6 +145,7 @@ let partition = (lam p. lam seq.
 utest partition (lam x. gti x 3) [4,5,78,1] with ([4,5,78],[1])
 utest partition (lam x. gti x 0) [4,5,78,1] with ([4,5,78,1],[])
 
+
 let findAssoc = lam p. lam seq.
   match find (lam tup. p tup.0) seq with Some res
   then Some res.1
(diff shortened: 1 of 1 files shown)


johnwikman:
I have not tested this, but it looks very similar to the functionality found in [stdlib/map.mc](https://github.com/miking-lang/miking/blob/develop/stdlib/map.mc). Feels like we should probably only have one or the other.

Also regarding the naming, would it perhaps be better to use _assoc_ as the prefix rather than the suffix? Such that we instead would have `assocAdd`, `assocFind`, etc.

david-broman:
@johnwikman Good point. I have actually never used the map.mc library (who added it?) and it is true that there is duplication, which it should not be. I did not really reflect on that library. 

I think we need to have a longer discussion about how we organize these libraries and their names. So far, it has been growing a bit too organic, without a high-level discussion and strategy. Also, I think that there should in general be better documentation in the standard libraries, compared to what it is there today. 

This is a good discussion point for the next meeting.

lingmar:
The map library was added in #92 for the digraph library, and based on grepping this seems like the only place it is used. It would probably make sense to change so that digraph uses the hashmap library instead.

david-broman:
We decided to change: in `map.mc` use function names such as `mapAdd` and in `hashmap.mc` 
 use names such as  `hashmapAdd`. We should remove `assoc` in `seq.mc`.

PR #148, opened 2020-08-20
Refactoring of map interfaces

lingmar:


lingmar:
Still a draft as some updates remain to be done for digraph.

stdlib/assoc.mc
```diff
@@ -0,0 +1,140 @@
...
+-- returned.
+let assocLookupOpt : AssocTraits -> k -> AssocMap -> OptionV =
+  lam traits. lam k. lam m.
+    optionMapOr (None ())
+                (lam t. Some t.1)
+                (find (lam t. traits.eq k t.0) m)
+
+-- 'assocLookup traits k m' returns the value of key 'k' in 'm'. Throws an error
+-- if 'k' does not exist in 'm'.
+let assocLookup : AssocTraits -> k -> AssocMap -> v =
```
elegios: I don't remember if we've discussed this at some point in the past, but I would prefer if things that can error have exceptional names, rather than the ones that are total. I'll bring it up at the meeting.

stdlib/hashmap.mc
```diff
@@ -25,17 +29,13 @@ let _hashmapDefaultBucketCount = 100
 let _hashmapBucketIdx = lam hash. lam hm. modi (absi hash) (length hm.buckets)
 
 
--- Returns an empty hashmap with a default number of buckets.
---   eq: A function that specifies equalities between keys in the hashmap.
---   hashfn: A function for computing the hash of a key value.
-let hashmapEmpty = lam eq. lam hashfn.
+-- 'hashmapEmpty ()' returns an empty hashmap with a default number of buckets.
+let hashmapEmpty : Unit -> HashMap = lam _.
```
johnwikman: I do not understand this change. Why does this have to be evaluated at runtime?

stdlib/assoc.mc
lingmar: Solved during meeting. Option type version is called lookup and the other lookupOrElse.

stdlib/hashmap.mc
lingmar: You are right, I fixed this.

PR #149, opened 2020-08-20
Update interpreter to more closely match boot

dlunde:
This pull request changes the interpreter to more closely match boot, and is as of now a work in progress. The following general changes will be included:

- General cleanup in `ast.mc`, `eval.mc`, `ast-builder.mc`, and `pprint.mc`.
- Unit constant and pattern removed (now expressed as empty records).
- If expressions removed (now expressed using `match`).
- Constant sequences (CSeq) removed.
- Tuples and tuple projections removed (now expressed with records).
- Record projections removed (now expressed using `match`).
- Data constructors no longer first-class (equivalent to #90 for boot)
- Added record patterns to `match`, so that projections can be expressed.
- Fix bug in eval for `TmRecordUpdate`.

The above has some consequences for `decision-points.mc`, `lamlift.mc`, `mcore-parser.mc`, and `meta.mc`.
- I can fix trivial things that break in `decision-points.mc` and `lamlift.mc` myself, but I might require some assistance from @lingmar @johnwikman for non-trivial changes. I'll notify you if this is the case.
- Should we deprecate/remove `mcore-parser.mc` and `meta.mc`? It looks like a lot of work to fix these after my other changes, and if I'm not mistaken, we will not use these files moving forward in any case.

In a future PR, I will also change to using symbols for variables (equivalent to #87 for boot).

Another thing that is not added in this PR is various missing patterns (e.g., SeqTot, SeqEdg, And, Or, Not). @elegios Maybe you could add this in a subsequent PR?


elegios:
> Another thing that is not added in this PR is various missing patterns (e.g., SeqTot, SeqEdg, And, Or, Not). @elegios Maybe you could add this in a subsequent PR?

I should be able to do that, though I think I'd like to wait until the symbol update has happened. Does that work, and in that case, could you ping me when that's ready?

dlunde:
@elegios Sure, I'll ping you

johnwikman:
> The above has some consequences for `decision-points.mc`, `lamlift.mc`, `mcore-parser.mc`, and `meta.mc`.
> 
>    - I can fix trivial things that break in `decision-points.mc` and `lamlift.mc` myself, but I might require some assistance from @lingmar @johnwikman for non-trivial changes. I'll notify you if this is the case.

The whole `lamlift.mc` should be refactored once identifier lookup is replaced by symbol lookup. Most of the lambda lifting code deals with somewhat complicated name mangling, which has led to a bug where typeless copies of an argument is created (I have yet to locate the source of this). If you integrate symbols into your PR then I would be happy to refactor `lamlift.mc`.

dlunde:
The PR is now ready for review!

@johnwikman Alright, I'll let you know when that is ready. For now, I say we simply let the lamlift testing (when running `make test`) fail, and merge anyway.

johnwikman:
> @johnwikman Alright, I'll let you know when that is ready. For now, I say we simply let the lamlift testing (when running `make test`) fail, and merge anyway.

How much has the AST changed in how it operates? If it is just a matter of renaming then you could probably just fix the naming errors present in `lamlift.mc`.

dlunde:
It has changed quite a lot (see full list in first comment). I also decomposed some language fragments (in particular those that combined patterns and terms) and reorganized `ast.mc`, `eval.mc`, `ast-builder.mc`, and `pprint.mc`. It would be nice if you could similarly reorganize `lamlift.mc`.

johnwikman:
> It has changed quite a lot (see full list in first comment). I also decomposed some language fragments (in particular those that combined patterns and terms) and reorganized `ast.mc`, `eval.mc`, `ast-builder.mc`, and `pprint.mc`. It would be nice if you could similarly reorganize `lamlift.mc`.

Will give it a go! You should hopefully see a PR on your fork soon

dlunde:
You can also wait until symbols are ready (will be in my next PR), and do it all in one go. Probably saves you some work!

johnwikman:
Though it is not nice to merge code that does not pass the tests. Should we perhaps comment out the whole contents of `lamlift.mc` instead? To be "dealt with later" :slightly_smiling_face:

dlunde:
Sure, I can do that for cleanliness :)

david-broman:
Yes, we should not merge anything to `develop` that do not pass all tests. If you would like to merge this PR before `lamlift.mc` is refactored, then we should remove lambda lift now. John can then add a new PR with the updated lambda lift, when you have time. I do not think that anything is dependent on lambda lift right now, right? 
(We would of course like to have it back as soon as possible).

david-broman:
Oops, just realized that you did this in between :-)

stdlib/mexpr/pprint.mc
```diff
@@ -313,8 +313,9 @@ lang IntPatPrettyPrint = IntPat
   | PInt t -> int2string t.val
 end
 
-lang CharPatPrettyPring = CharPat
-  -- TODO
+lang CharPatPrettyPrint = CharPat
+  sem getPatStringCode (indent : Int) =
+  | PChar t -> [head "'", t.val, head "'"]
```
johnwikman: You can use `'\''` instead of `head "'"`.

dlunde: Thanks, fixed (also for `CharPrettyPrint`).

PR #151, opened 2020-08-20
Refactoring of the parser combinator library

EliasC:
This commit includes a number of improvements to the parser
combinator library:

- The unit tests have been moved to the top level
- Some missing documentation was added
- An example parser for the untyped lambda calculus with numbers,  let bindings and if expressions was added to make it easier to understand for future users.


stdlib/parser.mc
```diff
@@ -8,23 +8,45 @@ let debug = lam s. if debug_flag then printLn s else ()
 -- The Parser monad -----------------------------
 
 type Pos = (String, Int, Int)
```
johnwikman: Could this perhaps be changed to use a record instead?

EliasC: That makes sense! If I recall correctly, this was written before there were records in the language 🙂

EliasC: Fixed now! Thanks!

ernstwi:
Not changed in this PR, but should [this line](https://github.com/miking-lang/miking/blob/29319664994aa77ef14ed5b07075b51c4106e0f1/stdlib/parser.mc#L59):

```
type Parser = State -> String -> ParseResult
```

instead be:

```
type Parser = State -> ParseResult
```

Since `State` holds the input string?

EliasC:
@ernstwi: Well-spotted! An effect of not having a type checker 🙂 
Fixed now!

PR #153, opened 2020-08-23
[Suggestion] sym2hash intrinsic for converting a symbol to an integer

johnwikman:
This (suggestion) PR adds the ~`sym2int`~`sym2hash` intrinsic that converts a symbol to an integer. I marked this PR as a suggestion as I am unsure if this is even something that we want to have in the language, or just want in the underpinnings of the compiler. The only guarantee of this operation would be:
```
eqs s1 s2  =>  eqi (sym2hash s1) (sym2hash s2)
```
While the opposite direction might not necessarily be true, the language should make an effort for that to be true. This makes it easier for symbols to be used in data structures that relies on hashing, while not making too strict demands on symbols generated in multiple threads (for when that is implemented).

I can also see a number of cons with having this intrinsic available:
 1. _Promotes misuse._ Unexperienced users of MCore might start converting all symbols they enounter to integers for convenience, leading to bugs further along the process.
 2. _Half-hearted._ While symbols are globally unique, the integer representation of it is not. If you would like to print a sequence of symbols to the terminal, the risk is that two different symbols are printed with the same string. Instead of allowing a symbol to be represented as an integer, it could instead be converted to a globally unique string. (How this would work I do not know.)
 3. _Deprecated from conception._ While symbols probably should be hashable, that is something that the compiler should take care of using unified collection types. This is just a temporary hack to allow more efficient lookups until UC types are in place.

At least I hope that this PR can clarify what we want from symbols and what we do not want from symbols :slightly_smiling_face:

UPDATE: Changed name from `sym2int` to `sym2hash`.

### DIFF
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -116,6 +116,7 @@ and const =
 | CSymb of int
 | Cgensym
 | Ceqs of int option
+| CSym2int
 (* External functions TODO: Should not be part of core language *)
 | CExt of tm Extast.ext
 | CPy of tm Pyast.ext
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -116,7 +116,7 @@ and const =
 | CSymb of int
 | Cgensym
 | Ceqs of int option
-| CSym2int
+| CSym2hash
 (* External functions TODO: Should not be part of core language *)
 | CExt of tm Extast.ext
 | CPy of tm Pyast.ext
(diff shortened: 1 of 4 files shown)


david-broman:
Very good and interesting thoughts. Having such an non-injective function with the property above makes sense to me. As you stated in item 3, I am not sure that it should be exposed in the end if UC types are built in, but if they are not, this is the kind of function you need (with certain properties). Since I do not think that this PR is urgent (right John?) I would suggest that we wait with this PR until you are present at one of the meetings.

johnwikman:
Yes, this PR is not urgent. The only slight worry I have with item 3 is that UC types might not cover all data structures that a symbol is useful in. I do not know what these might be, and maybe it would be a case of overengineering if we consider these unknown data structures.

Anyway, I am happy to wait with this PR until I can explain it during a meeting.

dlunde:
Thanks for the summary John. I have also implemented `sym2int` for pretty printing in the interpreter (see #158).

PR #156, opened 2020-08-26
Parametrized type signatures

aathn:
This PR adds syntactic support for parametrized type definitions and type signatures. At the moment, any type can be applied to any other type.

The PR also updates the type signatures of `assoc.mc` and `option.mc` to be parametrized.

### DIFF
--- a/src/boot/lib/parser.mly
+++ b/src/boot/lib/parser.mly
@@ -35,8 +35,8 @@
 %token <Ustring.ustring Ast.tokendata> LABEL_IDENT
 %token <Ustring.ustring Ast.tokendata> UC_IDENT  /* An identifier that starts with an upper-case letter */
 %token <Ustring.ustring Ast.tokendata> LC_IDENT  /* An identifier that starts with "_" or a lower-case letter */
-%token <Ustring.ustring Ast.tokendata> STRING_LIT
-%token <Ustring.ustring Ast.tokendata> CHAR_LIT
+%token <Ustring.ustring Ast.tokendata> STRING
+%token <Ustring.ustring Ast.tokendata> CHAR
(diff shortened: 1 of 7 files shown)


dlunde:
Looks good!

src/boot/lib/parser.mly
```diff
@@ -430,11 +436,16 @@ ty_op:
...
+  | ty_left
       { $1 }
-  | ty_atom ARROW ty
+  | ty_left ARROW ty
       { TyArrow($1,$3) }
 
+ty_left:
+  | ty_atom
+    { $1 }
+  | ty_left ty_atom
```
johnwikman, line 447: Should we really accept anything on the left-hand side here? For example, this makes `(Int -> Bool) a` a valid type according to the parser.

Also, since this is a syntax change, it would be good to take this opportunity to discuss how we communicate changes about the syntax. The type `Option (Option a)` softly breaks my syntax highlighter where it claims that the right parenthesis should not exist. My initial suggestion would be to give us who maintain syntax highlighters a short period of time, lets say a week, to sort this out before people starting to integrate these changes into the main code base. Until then, this syntax could be part of an example file under `test/` that maintainers can use to verify their changes. Though this is just a suggestion of what would work for me.

aathn: We considered disallowing syntax like `(Int -> Bool) a`, `[a] b`, or `Int a`, but in the end we decided against it. It makes the parser code more complicated, and there doesn't seem to be any major benefit at the current stage. Sure, `(Int -> Bool) a` is syntactically correct, but the types are only for documentation anyways. Also, I'm guessing this would be a thing for the type checker to check (analogous to how `5 10` is syntactically correct but would be caught as an error by the type checker)?

PR #157, opened 2020-08-27
Allow polymorphic cost in local search and bug fix in eq-paths

lingmar:
This PR allows cost to be polymorphic in local search. It also fixes some errors in the type annotations in local search and a bug in eq-paths.


david-broman:
Linnea, can you ping me when you are done (I saw int he notes that you should make an update)

PR #158, opened 2020-08-28
Symbolize for the interpeter

dlunde:
- **Changed identifiers from strings to names (i.e., tuples of type (String,Symbol)) and implemented symbolization of interpreter programs.**
- Added sym2int (duplicate of #153) in order to pretty print symbols in `pprint.mc`.
- Cleanup of unit tests, in particular in `eval.mc`.
- Added mapAccumR and mapAccumL to `seq.mc`.
- Added a few new functions (and unit tests) to `assoc.mc`: `assocMap` and `assocMapAccum`.

@lingmar: You seem to be relying quite a bit on the fact that identifiers are strings in `decision-points.mc`, so I was not able to (quickly) fix the things that broke on my own. Can you take a look at this in a subsequent PR?
I commented out `decision-points.mc` for now (as I did previously with `lamlift.mc`)
@johnwikman: As we have discussed, you can (same as for Linnea) start refactoring `lamlift.mc` when this PR gets merged.
@elegios: Same as above, but for missing patterns in match.

Lastly, I would like to mention that many of the test cases involving checking equality of (interpreter) programs broke after moving to symbols, since `symbolize` is not referentially transparent. I suggest that we implement a new equality function for interpreter programs which handles this.


stdlib/mexpr/ast.mc
```diff
@@ -259,9 +255,12 @@ end
 -- PATTERNS --
 --------------
 
+type PatName
+con PName     : Name -> PatName
+con PWildcard : ()   -> PatName
```
johnwikman, line 259: Should this perhaps be MLang syntax instead? (That is part of `VarPat`)

stdlib/mexpr/ast.mc
```diff
@@ -274,12 +273,12 @@ end
 
 lang RecordPat
   syn Pat =
-  | PRecord {bindings : AssocMap} -- AssocMap String Pat
+  | PRecord {bindings : AssocMapStringPat}
```
johnwikman: As #156 is merged, could we perhaps change this to `AssocMap String Pat` straight away?

stdlib/mexpr/ast.mc
```diff
@@ -396,6 +395,448 @@ lang MExprAst =
   TupleTypeAst + RecordTypeAst + DataTypeAst + ArithTypeAst + BoolTypeAst +
   AppTypeAst
 
+-----------------------
+-- BUILDER FUNCTIONS --
+-----------------------
```
johnwikman, line 399: I can't remember all the previous discussions regarding the ast-builder, but I recall that it was a conscious decision to keep the `ast-builder.mc` functionality separated from the `ast.mc` functionality. The only reason I can think about at the moment is that including `ast.mc` should only include that AST and nothing else.

stdlib/mexpr/pprint.mc
```diff
@@ -1,43 +1,60 @@
...
 -- Set spacing on increment
-let incr = lam indent. addi indent 4
+let incr = lam indent. addi indent 2
+
+let symbolDelim = "'"
+
+-- 'sym2string sym' returns a string corresponding to the symbol 'sym'. This
+-- exposes the underlying symbol representation, but is necessary for pretty
+-- printing.
+let sym2string = lam s. int2string (sym2int s)
```
johnwikman: Should this perhaps be part of a more general `symbol.mc` library instead? Sounds like something that could be useful for other debugging purposes.

stdlib/mexpr/ast.mc
elegios: Once we get a typesystem this choice will semi-significantly affect how they can be used. The last we got to (as I recall, @EliasC can probably clarify/remind) was essentially that types outside mlang are closed while those inside are extensible. This means that as they are now it would be sufficient to pattern match on `PVar { ident = PName _}` and `PVar { ident = PWildcard _}`, while if they were defined in mlang there would have to be an additional `PVar { ident = _ }` pattern.

(Note that it would be fine if the semantic function consumed a `PatName` directly, it's the nesting that means that we want to ensure exhaustiveness in this way)

stdlib/mexpr/cps.mc
```diff
@@ -1,83 +1,62 @@
...
-utest cpsM [] (var_ "x") with
-      var_ "x"
-in
+let id = _s (ulam_ "x" (var_ "x")) in
 
-utest cpsM [] id with
-      ulam_ "x" (ulam_ "k" (app_ (var_ "k") (var_ "x")))
-in
+-- TODO This is currently not working since the symbols differ between the LHS
+-- and RHS. Implement equality check between terms that takes care of symbols?
```
elegios, line 57: This is essentially alpha-equivalence. Definitely something we want, but it might pose a larger question of wanting a way to change what `utest` uses for equivalence.

As a temporary workaround in lieu of that one could have a function that just walks the tree and removes the symbols (replaces them with `nosym`).

stdlib/seq.mc
```diff
@@ -87,6 +87,32 @@ utest zipWith (zipWith addi) [[1,2], [], [10, 10, 10]] [[3,4,5], [1,2], [2, 3]]
...
 
+-- Accumulating maps
+let mapAccumL : (a -> b -> (a, c)) -> a -> [b] -> (a, [c]) =
+  lam f. lam acc. lam seq.
+    let res =
+      foldl
+        (lam tacc. lam x.
+           match f tacc.0 x with (acc, y) then (acc, cons y tacc.1) else never)
+        (acc, []) seq in
+    (res.0, reverse res.1)
```
aathn: You could use `snoc` instead of `cons` to avoid reversing at the end

Also, a potentially interesting observation is that this function is equivalent to `mapM` for `State a`: 
`(b -> State a c) -> [b] -> State a [c]`

stdlib/seq.mc
```diff
@@ -87,6 +87,32 @@ utest zipWith (zipWith addi) [[1,2], [], [10, 10, 10]] [[3,4,5], [1,2], [2, 3]]
...
+
+let mapAccumR : (a -> b -> (a, c)) -> a -> [b] -> (a, [c]) =
+  lam f. lam acc. lam seq.
+    foldr
+      (lam x. lam tacc.
+         match f tacc.0 x with (acc, y) then (acc, cons y tacc.1) else never)
+       (acc, []) seq
```
aathn, line 104: Maybe it would be nice to have the argument order of `f` reversed here for consistency with `foldr`? Not sure though, as that convention is somewhat confusing

elegios (approved):

stdlib/mexpr/pprint.mc
```diff
@@ -1,43 +1,60 @@
...
-      strJoin "" ["#var\"", name, "\""]
+let conString = lam name.
+  match name with (str,sym) then
+    let str =
+      if eqi (length str) 0 then
+        "#con\"\""
+      else if is_upper_alpha (head str) then
+        str
+      else
+        strJoin "" ["#con\"", str, "\""] in
```
aathn: For reasons mainly of elegance (but also performance I guess), I would prefer if `join` was used instead of `strJoin ""`

stdlib/mexpr/ast.mc
dlunde: I put it outside here since I assumed this is a type that would not get extended. Let's discuss at the next meeting!

stdlib/mexpr/ast.mc
dlunde: Definitely, will fix when I have time.

stdlib/mexpr/ast.mc
dlunde: Yes, I also vaguely remember such discussions. Let's have it again at the next meeting :)

stdlib/mexpr/pprint.mc
dlunde: Sure, I can do that.

stdlib/mexpr/cps.mc
dlunde: Removing the symbols is not correct though, since it is actually the symbols that define the variable now, rather than the strings (two names visible in the same scope can have the same string identifier, but different symbols). The easiest workaround (I believe) is to keep track of symbol mappings (LHS symbol -> RHS symbol) in an environment and check that the variable/constructor occurrences match while traversing the terms during equality check.

stdlib/seq.mc
dlunde: Doesn't `snoc` operate in linear time? Or maybe it is constant because of finger trees?

lingmar:
> @lingmar: You seem to be relying quite a bit on the fact that identifiers are strings in `decision-points.mc`, so I was not able to (quickly) fix the things that broke on my own. Can you take a look at this in a subsequent PR?
> I commented out `decision-points.mc` for now (as I did previously with `lamlift.mc`)

Sure!

stdlib/seq.mc
dlunde: Yes, I also thought that was standard. Data.List in Haskell has the same signature for both, however, so I followed that.

stdlib/mexpr/pprint.mc
dlunde: I'll take a look, I just copied similar code that I found in `pprint.mc`.

stdlib/seq.mc
aathn: `snoc` has the same performance as `cons` for finger trees: https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatFingerTree.S.html

stdlib/seq.mc
aathn: Ah, I didn't know that these are in Haskell's standard lib. Then it's probably best to follow that.

stdlib/seq.mc
dlunde: Implementation now changed to use snoc

stdlib/mexpr/pprint.mc
dlunde: Now fixed (everywhere, not just in `pprint.mc`). Thanks!

stdlib/mexpr/ast.mc
dlunde: Now fixed

stdlib/mexpr/ast.mc
dlunde: The merge is now reverted for this PR

stdlib/mexpr/cps.mc
david-broman: I think that the right way is that we add an alpha equivalence function that compares if two terms are equal. This should not be a primitive in MCore, but rather a function, part of the Mcore implementation in Mlang. Then, the utest would just check if a boolean is true or not. The checking would be done as Daniel suggests using an environment.

stdlib/mexpr/symbolize.mc
```diff
@@ -0,0 +1,334 @@
...
+  | TmApp {lhs = lhs, rhs = rhs} ->
+    TmApp {lhs = symbolize env lhs, rhs = symbolize env rhs}
+end
+
+lang FunSym = FunAst + VarSym + AppSym
+  sem symbolize (env : Env) =
+  | TmLam {ident = (str, _), tpe = tpe, body = body} ->
+    let sym = gensym () in
+    let env = insertId (IdVar str) sym env in
+    TmLam {ident = (str,sym), tpe = tpe, body = symbolize env body}
```
david-broman: Why are you manually creating string symbol tuples here? The whole point of having the name.mc library is to hide the tuple abstraction and only use these functions in name.mc.

stdlib/mexpr/symbolize.mc
```diff
@@ -0,0 +1,334 @@
...
+    TmRecordUpdate {rec = symbolize env rec, key = key,
+                    value = symbolize env value}
+end
+
+lang LetSym = LetAst
+  sem symbolize (env : Env) =
+  | TmLet {ident = (str, _),  body = body, inexpr = inexpr} ->
+    let body = symbolize env body in
+    let sym = gensym () in
+    let env = insertId (IdVar str) sym env in
```
david-broman: Same here. You should use name.mc instead.

stdlib/symbol.mc
```diff
@@ -0,0 +1,9 @@
+-- Functions related to the built-in symbol functionality (gensym, eqs, and
+-- sym2int)
+
+include "string.mc"
+
+-- 'sym2string sym' returns a string corresponding to the symbol 'sym'. This
+-- exposes the underlying symbol representation, but can be useful in many
+-- cases.
+let sym2string = lam s. int2string (sym2int s)
```
david-broman: I am not sure we should have a primitive sym2int. Potentially, we should have a sym2hash and another sym2string, where the former has the property that John described in the other PR, and where the latter must guarantee that the string is globally unique.

stdlib/mexpr/pprint.mc
```diff
@@ -1,51 +1,65 @@
...
+  match name with (str,sym) then
+    let str =
+      if eqi (length str) 0 then
+        "#con\"\""
+      else if is_upper_alpha (head str) then
+        str
+      else
+        join ["#con\"", str, "\""] in
+    let sym = if nameHasSym name then sym2string sym else "" in
+    strJoin symbolDelim [str, sym]
```
david-broman: We need to discuss how to pretty print this. The main idea of the name module was not to expose the actual symbol when pretty printing. Instead, the actual string should be used. But, the pretty printer must keep an environment which makes sure that the string is locally unique, potentially adding a number to the name, which makes it unique. By doing so, we get a much more readable pretty printer that always creates correct program code (that can be compiled).

Hence, for pretty printing names of terms, you do not need sym2string. Note the special case if you pretty print open terms (no lambda or match constract that is catching the variable name). In such a case, we should probably use the real name, but append/append it with some special string. E.g., if we have a name "x" we could pretty print it is "open_x" or something similar.

stdlib/mexpr/symbolize.mc
dlunde: True, I should not do this. I'll fix this (and the other occurrences).

stdlib/mexpr/pprint.mc
dlunde: Improving the pretty printer in this way is a non-trivial task, which is why I did not attempt it in this PR. I agree that we should do this in a future PR (in which we can also remove/change sym2int and sym2string).

I think it is best to merge this PR now, however, so that other people can start incorporating the new symbol stuff in their implementations. I can take a look at improving the pretty printing afterwards (it is actually already in my TODO list).

stdlib/mexpr/symbolize.mc
dlunde: Fixed

stdlib/mexpr/symbolize.mc
dlunde: Fixed

PR #159, opened 2020-08-31
Prepend '_' to record keys for tuples in AST builder

lingmar:
`ast-builder.mc` represents tuples as records, where the keys are the indices of the tuples, i.e.

`(1,2) -> {0 = 1, 1 = 2}`

For debugging I sometimes run a pretty-printed AST as an MCore program, but pretty-printing these records results in invalid MCore as integers are not valid identifiers. This PR prepends '_' to the keys to fix this, i.e.

`(1,2) -> {_0 = 1, _1 = 2}`

### DIFF
--- a/stdlib/mexpr/ast-builder.mc
+++ b/stdlib/mexpr/ast-builder.mc
@@ -2,6 +2,10 @@
 
 include "ast.mc"
 
+-- Private definitions
+let _tupleIdx2RecordKey = lam i.
+  concat "_" (int2string i)
+
 -- Methods of binding an expression into a chain of lets --
 recursive let bind_ = use MExprAst in
   lam letexpr. lam expr.
@@ -95,7 +99,7 @@ let prec_ = use MExprAst in
 
 let ptuple_ = use MExprAst in
   lam ps.
-  prec_ (mapi (lam i. lam p. (int2string i,p)) ps)
+  prec_ (mapi (lam i. lam p. (_tupleIdx2RecordKey i, p)) ps)
 
 -- Types --
 let tyarrow_ = use MExprAst in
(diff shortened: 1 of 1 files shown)


johnwikman:
Shouldn't the pretty-printer output `{#label"0" = 1, #label"1" = 2}` instead? Alternatively, the pretty-printer could instead look at the record and output a tuple if the labels are a sequence of integers starting at 0.

lingmar:
I looked at the pretty-printer and saw that we use `#var"..."` and `#con"..."` for invalid variable- and constructor names (I wasn't aware of this), so perhaps your first suggestion would be better for consistency. The second suggestion wouldn't comply with how tuple projections are done in `ast-builder` (via a record projection).

johnwikman:
Shouldn't `(1,2)` and `{#label"0" = 1, #label"1" = 2}` parse to the same AST node? Feels like I might have missed some detail regarding the second suggestion

lingmar:
Yes, my bad :) So both suggestions should work.

Anyway, I guess we then want to pretty-print matches with record patterns as e.g. `match t with {#label"0" = 1} then ...` to get pretty-printed tuple projections to parse properly.

dlunde:
Couldn't we also just pretty print tuples as tuples? That is, `{#label"0" = 1, #label"1" = 2}` pretty prints as `(1,2)`.

Edit: Just saw that John has already suggested this, sorry

lingmar:
We decided at the meeting to go with option 2 (although perhaps splitted into 2 PRs; one PR for option 1 and then another to make tuples pretty print as tuples), but I'll wait until #158 is merged before doing this.

Issue #136, opened 2020-08-11
Short-circuit evaluation


aathn:
Currently, MCore does not have short-circuit evaluation for the boolean operators `and` and `or`, which is standard in most programming languages.

For instance, consider the following piece of code.

```ocaml
and false (let _ = printLn "hej" in true)
```

The code evaluates to `false` as expected, but it also prints `hej`, since both arguments to `and` are evaluated regardless of the fact that the first argument evaluates to `false`.

Short-circuiting would be nice to have, but would require treating `and` and `or` differently from the other intrinsics. Is it worth it, what do you think?

johnwikman:
I suspect that we brought this up during a meeting after you made this issue, but never referenced this issue. IIRC the conclusion was that we do not want to have unneccesary special cases that deviates from the standard way of evaluation. As we are still just an IR, a higher-level language could implement `&&` as
```ml
a && b := if a then b else false
```

aathn:
On the contrary, I opened this issue after that meeting, at the request of David, because the conclusion was not that clear-cut. (Or maybe I missed some very recent discussion?)

johnwikman:
You did probably not miss a discussion, most likely I who forgot the conclusions of that meeting :slightly_smiling_face:

I am personally a bit torn on this. Because what is the purpose of having `and` and `or` instrinsics if they do not provide short-circuit evaluation? On the other hand, why introduce special cases in an IR for application when the desired behavior can be implemented using a short `if ... then ... else true/false`. What is there to gain for having the `and` and `or` intrinsics except for shorter (and perhaps cleaner) code?

johnwikman:
I think that my conclusion would be to remove `and` and `or` as intrinsics. Instead implement them in `prelude.mc` as regular functions.

elegios:
We discussed this at the meeting today and essentially came to the same conclusion as @johnwikman above: make `and` and `or` into functions in the prelude instead of intrinsics.

The potential drawback of this is recognizing compilation opportunities; most target languages support short-circuiting and/or builtins for `and` and `or`. The backends compiling to these targets would thus have to redisover these opportunities, and it would be nice to share this work between them.

The thought presented at the meeting was that we can have the following structure:
- We have mexpr proper, which is the small language we generally talk about. It has few intrinsics, and is the language most transformations and operations operate upon.
- We have a set of extension languages on top that rediscover these constructs. For example, one fragment could be something like this:

```
lang ShortCircuitAnd = MExpr
  syn Expr =
[...]

PR #160, opened 2020-09-01
Remove and,or as intrinsics

johnwikman:
Removes `and` and `or` as intrinsics and instead implements them as regular functions in the standard library. This relates to the discussion about short-circuiting from issue #136. Also adds the `xor` and `xnor` functions for completeness.


elegios:
Probably worth making the change to `ast.mc` at the same time, I think it has a `CAnd` and `COr`.

Issue #95, opened 2020-05-19
Bikeshedding: Comment syntax


EliasC:
Sometimes the least important decisions require the most discussion. With that in mind, let's talk syntax! We have had some informal discussions regarding the syntax for comments, and the point of this space is to continue the discussion in an archived fashion (and hopefully reach a conclusion). 

We currently support two kinds of line comments:
- Haskell-style: `-- This is a comment`
- C-style: `// This is a comment`

For starters, we should decide on a single style (which doesn't have to be one of these two). We should also decide on a syntax for block comments, and there should be a clear connection between the two. I remember three suggestions being brought up:

- A riff on the Haskell-style:
  ```
  -- This is a line comment
  --*
    This is a
    block comment
  *--
  ```
- A variation with identical open/close symbols:
  ```
  -- This is a line comment
  *--
    This is a
    block comment
  *--
  ```
- C-style all the way:
  ```
  // This is a line comment
  /*
    This is a
    block comment
  */
  ```

The floor is open, both for discussing these alternatives and for suggesting new ones!

br4sco:
A variation on the first one would be:
```
-- This is a line comment
-*
  This is a
  block comment
*-
```

EliasC:
@br4sco: Maybe that was even the original suggestion, come to think of it!

johnwikman:
I like @br4sco's suggestion for block comments. I am not too demanding on the syntax here, but it would be good to have the ability to nest block comments.

Edit: Another thing to perhaps add to this discussion is documentation and if we would like to generate online reference manuals similar to how Doxygen does it:
```
-*
 * Sorts a sequence of elements using heap sort.
 *   lam lt: Performs the less-than operation (<) on elements in the input array.
 *   lam s:  The sequence to be sorted.
 *-
let heapsort = lam lt : (a -> a -> Bool). lam s : [a].
  ...
```

david-broman:
I think I prefer Oscar's suggestion, or the one with two dashes
```
--*
text 
*--
```
I agree with John, it is important to support nested block comments.

br4sco:
I don't know if there are parsing considerations to this but how about using a markdown code block style block comment?

A variation on the first one would be:
```
-- This is a line comment
---
  This is a
  block comment
---
```

I guess you can increase the number of `-` to better distinguish the block comment from the line comment if you wish.

segeljakt:
You wont be able to meta-program the comments?!? 😆

EliasC:
[...]

PR #161, opened 2020-09-04
Remove support for C++ style line comments

johnwikman:
Removes support for C++ style line comments as discussed in #95.

This will probably break some MCore code in other repositories. But I assume that makes it more important to get this change in as soon as possible?


EliasC:
Looks straightforward to me!

PR #162, opened 2020-09-09
WIP: Add argument parser library to stdlib

johnwikman:
Adds an argument parser to stdlib. This parser is intended to
 * handle command line options (with and without parameters),
 * handle positional arguments with different priorities and orderings (not yet implemented), and
 * output the argparser configuration as a usage string.

About half of the features are missing at the moment, but opening up this PR prior to completion to allow comments on the design. E.g. the heavy reliance on errors, allowing compound parameter options such as `-O44` or `-Ddebug`, etc.


elegios:
This is less of a comment on the current design, and more of a comment on a different style I've found quite pleasant to work with, for comparison.

[Optparse-applicative](https://hackage.haskell.org/package/optparse-applicative) in Haskell models an argument parser as an applicative functor, similar to how parser combinator libraries do it (applicative instead of monadic since options can be given in any order and monad implies strict ordering, and applicative also makes it possible to construct a usage message). A `Parser a` is thus something that produces an `a` instead of something that modifies an `a`.

Much of the interface is given through standard typeclasses (which is an advantage there but wouldn't be for us, since we don't have typeclasses, at least not yet): Functor, Applicative, and Alternative, meaning that common operators such as `optional`, `many`, and `some` work. The interface is nicely compositional, and it's quite easy to express some decently complicated relationships, such as "you need at most one of these switches" or "these groups of switches are mutually exclusive". Some examples from the documentation:

```haskell
-- A parser that counts the number of "-t" arguments on the command line
length <$> many (flag' () (short 't'))

-- A parser that requires exactly one of "--on" or "--off" to be given
flag' True (long "on") <|> flag' False (long "off")
```

I becomes even nicer with `ApplicativeDo` (which also wouldn't benefit us since we don't have it, but eh), here's some of the configuration for [syncon-parser](https://github.com/miking-lang/syncon/blob/master/syncon-parser/exe/Main.hs) (with many options skipped since they're structurally quite similar):

```haskell
pbtCmd = do
  precKind <- Opt.flag LD.PermissivePrecedence LD.RestrictivePrecedence
    $ Opt.long "restrictive-precedence"
    <> Opt.help "Restrict low-precedence unary operators to retain unambiguity in the presence of a total precedence list. This changes the recognized syntactic language, but not the semantic language."
  numRuns <- Opt.option Opt.auto
    $ Opt.long "tests"
    <> Opt.metavar "N"
    <> Opt.value 10000
    <> Opt.help "The number of tests to run."
  showAmbDistr <- Opt.switch
    $ Opt.long "amb-distr"
    <> Opt.help "Show the percentage of tested programs that are ambiguous"
  dynAmbTimeout <- fmap (*1_000) $ Opt.option Opt.auto
    $ Opt.long "timeout"
    <> Opt.metavar "MS"
    <> Opt.help "Timeout before discarding a test case, in milliseconds. A negative value means 'wait forever'."
    <> Opt.value 4_000
  dynAmbKind <- dynOptFlag
  summaryLogFile <- optional $ Opt.option Opt.str
    $ Opt.long "summary-log-file"
    <> Opt.help "File to log a summary of this pbt run to. Result will be appended."
    <> Opt.metavar "FILE"
  files <- some $ Opt.argument Opt.str $
    Opt.metavar "FILES..."
    <> Opt.help "The '.syncon' files that define the language to be tested."
  pure $ -- a bunch of code that can use 'precKind', 'numRuns', etc.

dynOptFlag = fast <|> complete <|> race' <|> pure DynAmb.RaceDyn
  where
    fast = Opt.flag' DynAmb.FastDyn
      $ Opt.long "fast-dyn"
      <> Opt.help "Use the fast dynamic analysis that might not terminate (except for timeout) on unresolvable input"
    complete = Opt.flag' DynAmb.CompleteDyn
      $ Opt.long "complete-dyn"
      <> Opt.help "Use the complete dynamic analysis that will always terminate, but that might be very slow"
    race' = Opt.flag' DynAmb.RaceDyn
      $ Opt.long "race-dyn"
      <> Opt.help "Run both dynamic analyses and use the one that terminates first."
```

I use this by having the code after `pure $` have the type `IO ()` and being the action to perform. The code then looks a bit like a preamble that defines all the options, then code that just uses them, without having to go through any intermediate steps.



```

**Comment by johnwikman** [comment]
From today's meeting: Will change implementation to using lists of constructors instead of a predefined set of functions.

**Comment by dlunde** [comment]
It might also be worth having a look at OCamls Arg library: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Arg.html. It is quite nice to use, but also quite limited.

**Comment by lingmar** [comment]
Some things I would like to have:
- Multiple parameters get aggregated into a list, e.g. that `--foo arg1 arg2 arg3` gives `[arg1,arg2,arg3]`.
- The number of times an argument is given can change behaviour, e.g. increasing verbosity levels `-v`, `-vv` `-vvv`.
- Some way of specifying whether an argument is required or not and that an error is given if a required argument is missing.
- Ideally, a way of specifying relationships/dependencies between arguments like the examples Viktor mention.

**Comment by johnwikman** [comment]
>  * The number of times an argument is given can change behaviour, e.g. increasing verbosity levels `-v`, `-vv` `-vvv`.

Are you referring to only if they are located together like `-vvv -h -i`, or just if it is specified 3 times like `-vvh -i -v`? The latter is possible at the moment using `lam o. {o with verbosity = addi o.verbosity 1}` or something similar.

**Comment by lingmar** [comment]
> Are you referring to only if they are located together like `-vvv -h -i`, or just if it is specified 3 times like `-vvh -i -v`? The latter is possible at the moment using `lam o. {o with verbosity = addi verbosity 1}` or something similar.

As long as they are specified 3 times would be enough, not necessary to require them to be together I think!

**Comment by david-broman** [comment]
> It might also be worth having a look at OCamls Arg library: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Arg.html. It is quite nice to use, but also quite limited.

@dlunde Yes, but it is also based on references, which does not work in our case.

**Comment by david-broman** [comment]
@johnwikman Just for the record: as I mentioned at the Miking meeting, I would really prefer a simple approach where you specify all parameters in a list, instead of using a the more imperative version as is available in the current version. For instance, a constructor can be something like
ArgOption("p", "This is the text", lam x. ...)
etc. Keeping it less verbose is nice when you want to use the library.

**Comment by dlunde** [comment]
> > It might also be worth having a look at OCamls Arg library: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Arg.html. It is quite nice to use, but also quite limited.
> 
> @dlunde Yes, but it is also based on references, which does not work in our case.

Oh right, that's true

**Comment by johnwikman** [comment]
So I think that I am converging on something like this:
<details>
<summary><b>(Uncollapse this block to see the example)</b></summary>
<p>

```haskell
type ArgParserArg a
con ArgParserOption: [ArgParserModifier a] -> ArgParserArg a
con ArgParserPositional: [ArgParserModifier a] -> ArgParserArg a
con ArgParserMutuallyExclusive: [ArgParserArg a] -> ArgParserArg a
con ArgParserMutuallyExclusiveOptions: [ArgParserModifier a] -> ArgParserArg a

type TestArgs = {
  help: Bool,
  version: Bool,
  debugParser: Bool,
  optLevel: Int,
  defines: [String],
  multi: (String, String, String)
} in

let argspec = [
  ArgParserProgramName "test",
  ArgParserOption [
    ArgParserShort 'h',
    ArgParserLong "help",
    ArgParserDescription "Prints a help message and exits.",
    ArgParserApply (lam o. {o with help = true})
  ],
  ArgParserOption [
    ArgParserShort 'v',
    ArgParserLong "version",
    ArgParserDescription "Prints version and exits.",
    ArgParserApply (lam o. {o with version = true})
  ],
  ArgParserOption [
    ArgParserLong "debug-parser",
    ArgParserDescription "Shows debug prints during parsing.",
    ArgParserApply (lam o. {o with debugParser = true})
  ],
  ArgParserOption [
    ArgParserShort 'O',
    ArgParserLong "optimization-level",
    ArgParserMetavar "LEVEL",
    ArgParserDescription "Sets the optimization level.",
    ArgParserApplyVar (lam p. lam o. {o with optLevel = string2int p})
  ],
  ArgParserOption [
    ArgParserShort 'D',
    ArgParserLong "define",
    ArgParserMetavar "DEFINITION",
    ArgParserDescription "Add C preprocessor definition.",
    ArgParserApplyVar (lam p. lam o. {o with defines = snoc o.defines p})
  ],
  ArgParserOption [
    ArgParserLong "a-very-long-option-name",
    ArgParserMetavar "fooparam",
    ArgParserDescription "Shows debug prints during parsing.",
    ArgParserApplyVar (lam _. lam o. o)
  ],
  ArgParserOption [
    ArgParserLong "multi",
    ArgParserMetavars ["arg1", "arg2", "arg3"],
    ArgParserDescription "Takes multiple parameters.",
    ArgParserApplyVars (lam sl. lam o. {o with multi = (get sl 0, get sl 1, get sl 2)})
  ],
  ArgParserMutuallyExclusive [
    ArgParserOption [
      ArgParserLong "on",
      ArgParserDescription "Turns it on!",
      ArgParserApply (lam o. {o with isOn = true})
    ],
    ArgParserOption [
      ArgParserLong "off",
      ArgParserDescription "Turns it off!",
      ArgParserApply (lam o. {o with isOn = false})
    ]
  ],
  -- The above would be a wrapper for this, which allows for more complex behavior:
  ArgParserMutuallyExclusiveOptions [
    ArgParserLong "on",
    ArgParserLong "off"
  ]
] in

-- Is of type "Either TestArgs [String]" (or similar) where the right part is the collected error messages when something went wrong.
let result = argparserParse argspec defaults (tail argv) in
```

</p>
</details>

Then there would be many sanity checks in `argparserParse` to make sure that the options are well formed. Thoughts?

EDIT: Just to connect this to @david-broman's comment above, the idea was to have this setup as the "backend". Then you can have short-hand versions such as `ArgParserFlag ('h', "help", "Print help.", lam o. {o with help = true})` that gets translated into the more modular backend within `argparserParse`.

EliasC:
> Multiple parameters get aggregated into a list, e.g. that `--foo arg1 arg2 arg3 gives [arg1,arg2,arg3]`.

An issue with this is that you risk confusing program arguments with option arguments:
```
> ./myProg input.txt
...running...
> ./myProg --foo arg1 arg2 input.txt
No input file given
> ./myProg input.txt --foo arg1 arg2
...running...
```
Sometimes this is solved by having `--` mean "and the rest of the command line arguments are program arguments", but I don't know if we want to build this behaviour into the library since it is arguably more complicated:
```
> ./myProg --foo arg1 arg2 -- input.txt
...running...
```

One way of doing it right now is the same way (e.g.) gcc handles includes, by writing `--fooarg1:arg2:arg3` and having a handler that does something like `lam s. lam o. {o with fooList = split ":" s}` (where `split ":"` splits a string into the substrings separated by `":"`). It is not as nice as having them separated, but it avoids the issue above. Not sure which is the right way forward!

EliasC:
> So I think that I am converging on something like this...

I think this looks reasonable, especially with your edit about a short hand! Personally I wouldn't mind having only the short hand, as it lets you encode the required meta parameters in the type system instead of running sanity checks to see if there is at least one name, that there is a handler function, etc. Something like this:
```
type ArgParserArg opt
con ArgParserFlag : all opt . Char -> String -> String -> (opt -> opt) -> ArgParserArg opt
con ArgParserOpt : all opt . Char -> String -> String -> (String -> opt -> opt) -> ArgParserArg opt
```
Sure, it's slightly less flexible, but my feeling is that with only a few variants you will cover 99% of all the use cases and get something which is significantly simpler.

Some minor comments:
```let argspec = [
  ArgParserProgramName "test",
  ArgParserOption [
    ArgParserShort 'h',
    ArgParserLong "help",
    ArgParserDescription "Prints a help message and exits.",
    ArgParserApply (lam o. {o with help = true})
  ],
  ...
```
This should probably be a tuple of a string and a list of `ArgParserArg`, and not a single list (I'm guessing `ArgParserProgramName` is not of the same type as the rest of the elements of the list).

```
-- Is of type "Either TestArgs [String]" (or similar)...
```
The mnemonic for `Either` is usually "right is right", meaning the right argument is the expected result, and the left argument is the error case.

johnwikman:
> > So I think that I am converging on something like this...
> 
> I think this looks reasonable, especially with your edit about a short hand! Personally I wouldn't mind having only the short hand, as it lets you encode the required meta parameters in the type system instead of running sanity checks to see if there is at least one name, that there is a handler function, etc. Something like this:
> 
> ```
> type ArgParserArg opt
> con ArgParserFlag : all opt . Char -> String -> String -> (opt -> opt) -> ArgParserArg opt
> con ArgParserOpt : all opt . Char -> String -> String -> (String -> opt -> opt) -> ArgParserArg opt
> ```
> 
> Sure, it's slightly less flexible, but my feeling is that with only a few variants you will cover 99% of all the use cases and get something which is significantly simpler.

Then the solution would essentially be the same as it is right now, except that the argparser interface looks different. I am also thinking that there are a lot of special cases here that usually do not turn up so often, but that most applications probably contains at least one such special case. An example could be that `--fastcrc` is only valid if `--backend=amd64` has been set. Another could be that you want to limit the values of an option parameter to a small prefined set, such as limiting `--backend` to `[amd64, ocaml, llvm]`. As such, I am more converging on letting the user specify "option features" rather than being restricted to a set of predefined option types (as it was initially).

> 
> Some minor comments:
> 
> ```
>   ArgParserProgramName "test",
>   ArgParserOption [
>     ArgParserShort 'h',
>     ArgParserLong "help",
>     ArgParserDescription "Prints a help message and exits.",
>     ArgParserApply (lam o. {o with help = true})
>   ],
>   ...
> ```
> 
> This should probably be a tuple of a string and a list of `ArgParserArg`, and not a single list (I'm guessing `ArgParserProgramName` is not of the same type as the rest of the elements of the list).

The idea was that these would be the same types. `ArgParserArg` is perhaps not the most suitable name, but more `ArgParserConfigItem` or something similar. I am not trying to overengineer this, but I imagine that you could want additional modifiers to the behavior of the parser, such as making all option names case insensitive.

> ```
> -- Is of type "Either TestArgs [String]" (or similar)...
> ```
> 
> The mnemonic for `Either` is usually "right is right", meaning the right argument is the expected result, and the left argument is the error case.

That sounds ~more right~ better!

EliasC:
> I am also thinking that there are a lot of special cases here that usually do not turn up so often, but that most applications probably contains at least one such special case.

I think the real question is whether it is the job of the argument parsing library to handle these special cases. For your examples, you could easily write
```
match argparserParse argspec defaults (tail argv) with
  Right options ->
    if (and options.fastcrc (not (isAmd64 options.backend))) then
      error "Fastcrc can only be used with the Amd64 option"
    else if (elem options.backend validBackends) then
      error (concat "Unknown architecture " options.backend)
    else
      ...
  Left errors -> ...
```
If you want to be more abstract, you could add a post-processing step that transforms `options` from something consisting of parsed strings into a more specific datatype (and throws an error if the options are malformed). You're moving complexity to the user, but at the same time also adding specificity of error messages etc.

It all comes down to how complicated the library is to use in the simple, most common cases. If I am not affected by the support for special cases when I don't have special cases, I don't mind if there are additional possibilities to the library. 

> I am not trying to overengineer this, but I imagine that you could want additional modifiers to the behavior of the parser, such as making all option names case insensitive.

That makes sense, but there is also the issue of the complexity of the library, with more and more sanity checks (and the programmer needing to think about these explicitly). For example, why do we need the possibility to add multiple names to a program if we are never going to allow more than one?

elegios:
I believe that the applicative approach can handle most or all of these extra complications. Trying to use names we could actually use instead of Haskell's:

-----------------------------------
> Multiple parameters get aggregated into a list, e.g. that --foo arg1 arg2 arg3 gives [arg1,arg2,arg3].

This probably wouldn't work for the ambiguity reasons that Elias mentions, but could instead be expressed as:
```
many (ArgParseLongOption {long = "foo", help = "...", parser = (lam x. Right x)}) 
-- This could be used as "--foo arg1 --foo=arg2 --foo arg3"
-- parser would be a function String -> Either Error a, and the argparser would produce an a

-- As an aside,
many p = alt (some p) (pure [])
some p = ap (fmap cons p) (many p)
-- modulo laziness, this would work directly in Haskell, but would probably
-- need a little bit of extra thought in an eagerly evaluated language
```
---------------------------------------
> The number of times an argument is given can change behaviour, e.g. increasing verbosity levels -v, -vv -vvv
```
fmap length (many (ArgParseFlag {long = "verbose", short = "v", help = "..."}))
```
--------------------------------------
> Some way of specifying whether an argument is required or not and that an error is given if a required argument is missing.
```
ArgParseFlag {...}  -- Required
optional (ArgParseFlag {...})  -- Optional, becomes None if absent, otherwise Some x, where x is produced by the inner argparser
alt (ArgParseFlag {...}) (pure val)  -- Optional, becomes val if absent, otherwise x, where x is produced by the inner argparser

-- As an aside
optional p = alt (fmap (lam x. Some x) p) (pure (None ()))
```
-------------------------------------
> An example could be that --fastcrc is only valid if --backend=amd64 has been set.

This I suspect wouldn't work in in `optparse-applicative` since it's an applicative, not a monad, and the value `amd64` wouldn't exist until parse time. It could be worked around like so:
```
(ap
  (fmap
    (lam _. lam fastcrc. {fastcrc = fastcrc})
    (ArgParseLongOption {long = "backend-amd64", help = "..."}))
  (optional (ArgParseLongFlag {long = "fastcrc", help = "..."}))

-- This is the part that suffers the most for us, in Haskell this would be
-- (\_ fastcrc -> ...) <$> ArgParseOption {...} <*> optional (...)
-- where <$> = fmap, and <*> = ap.
```

It is also possible that we could something like selective applicative functors here to get the original shape, but I don't know that for sure, there are some questions on how we observe the condition.

It is of course also possible to use Elias' approach here, and separately check it afterwards, or have a way to inject checking in the applicative, e.g., `check : (a -> Either Error b) -> ArgParser a -> ArgParser b`.

--------------------------------------
It is I suppose worth mentioning that how the innermost thing looks (e.g., `ArgParseOption`, `ArgParseFlag`, etc.) is fairly open here, the way it works in Haskell is kind of emulating the builder pattern in Java or the like, except we're building the specification using Monoid `<>` instead of chained method calls. We could easily choose something else, depending on what we need and/or want.

It is also worth mentioning that I don't know how hard it is to implement this interface, only that it's really nice to use, both for simple options and for more complicated scenarios, so it's quite possible that this would be a library to write later.

johnwikman:
> If you want to be more abstract, you could add a post-processing step that transforms `options` from something consisting of parsed strings into a more specific datatype (and throws an error if the options are malformed). You're moving complexity to the user, but at the same time also adding specificity of error messages etc.

I was thinking of something similar to this, but that you allow an `ArgParserCond` as an argument to allow the error to be caught along with all the other errors:
```haskell
-- (condition function, error message)
con ArgParserCond: (a -> Bool) -> String -> ArgParserArg a
```

> That makes sense, but there is also the issue of the complexity of the library, with more and more sanity checks (and the programmer needing to think about these explicitly). For example, why do we need the possibility to add multiple names to a program if we are never going to allow more than one?

One thing that I had in the back of my mind was that this list could increase modularity, such that you do not have to define all properties of the argument parser in one place. But now when I think about it, this is probably unneccessary for anything that does not concern specific options. I suppose you could split it into two parts, one with the argparser _properties_ and another with the _arguments_.

My question regarding the discussion from the meeting yesterday is whether we want to change how the current code works, or just how the interface looks?

aathn:
I'll just tune in and say that I think the applicative approach proposed by Viktor looks very nice and flexible. The biggest downside which immediately stands out to me is that the syntax would be somewhat awkward since we don't have operators. However, I think with the right definitions it wouldn't be so bad. For instance, one could do something like the following to allow a user to build a 'list' of arguments, where `optCons = lam a. lam b. lam f. ap (b f) a` and `optNil = pure`:

```ocaml
let doStuff = lam opt1. lam opt2. lam opt3. ...

let myOpts =
  (optCons (parseFlag {long = "opt1", ...})
  (optCons (optMap length (many (parseFlag {long = "opt2", ...})))
  (optCons (optional (parseFlag {long = "opt3", ...}))
           optNil)))

let main = runParser (myOpts doStuff)
```

johnwikman:
> I'll just tune in and say that I think the applicative approach proposed by Viktor looks very nice and flexible. The biggest downside which immediately stands out to me is that the syntax would be somewhat awkward since we don't have operators. However, I think with the right definitions it wouldn't be so bad. For instance, one could do something like the following to allow a user to build a 'list' of arguments, where `optCons = lam a. lam b. lam f. ap (b f) a` and `optNil = pure`:
> 
> ```ocaml
> let doStuff = lam opt1. lam opt2. lam opt3. ...
> 
> let myOpts =
>   (optCons (parseFlag {long = "opt1", ...})
>   (optCons (optMap length (many (parseFlag {long = "opt2", ...})))
>   (optCons (optional (parseFlag {long = "opt3", ...}))
>            optNil)))
> 
> let main = runParser (myOpts doStuff)
> ```

I also think that Viktor's approach is nice, but not really applicable in our case. The main issue that I can see with the example you wrote is the parseFlag records, where the `...` is not possible today. My guess is that quite a lot that will go into that record, so without the ability to have default values it becomes fairly user unfriendly.

aathn:
> I also think that Viktor's approach is nice, but not really applicable in our case. The main issue that I can see with the example you wrote is the parseFlag records, where the ... is not possible today. My guess is that quite a lot that will go into that record, so without the ability to have default values it becomes fairly user unfriendly.

I don't think that should be a principal constraint. One simple way of handling it could be something like the following (there's probably some better way of doing it too):
```ocaml
type Opt = { long: String
           , short: String
           , help: String
           , ...
           }

let defaultOpt = { long = "", short = "", help = "", ... }

let mkOpt = lam f. mkParser (f defaultOpt) -- mkParser makes a parser based on the Opt record

let optLong = lam s. lam o. { o with long = s }
let optShort = lam s. lam o. { o with short = s }
let optHelp = lam s. lam o. { o with help = s }
...

let optPlus = compose

let myOpts =
  (optCons (mkOpt (optPlus (optLong "opt1")
                  (optPlus (optShort "o1")
                           (optHelp "My first option"))))
  (optCons (optMap length
                   (many (mkOpt (optLong "opt2"))))
  (optCons (optional (mkOpt (optLong "opt3")))
           optNil)))
```

johnwikman:
This is sort of what I suggested previously (and started working on), except that it looks like
```ml
let help = AP.Option [
    AP.Long "help",
    AP.Short 'h',
    AP.Description "Print a help message.",
    AP.Apply (lam o. {o with help = true})
]
```
Then you would have short-hand versions of common special cases, such as
```ml
let help = AP.Flag ('h', "help", "Print a help message", lam o. {o with help = true})
```
which then translates into the more modular backend shown above. Then you can define your functions above to target the `AP.Option` constructor.

aathn:
You're right that using a list of constructors is probably a nicer interface, and the `Opt` record I proposed above could certainly be replaced by something similar. However, the point is that by using the applicative approach, we gain composability and flexibility.
For instance, suppose I had two numeric options, `opt1` and `opt2`, which I wanted to add and use the result in my program. With the applicative approach I could use `liftA2 addi opt1 opt2` to immediately get a new parser which will return the sum of the two options, whereas with your approach I would essentially be required to aggregate the results into a record first and then perform my addition. Not that I couldn't do that, but I find the first way a lot more elegant.

PR #163, opened 2020-09-09
Pretty printing tuples

lingmar:
This PR updates the pretty printing for records so that:
- Keys x that are not valid identifiers are printed as #label"x".
- Records that represent tuples are printed as tuples, i.e. {#label"0" = 1, #label"1" = 2} -> (1,2).

### DIFF
--- a/stdlib/mexpr/pprint.mc
+++ b/stdlib/mexpr/pprint.mc
@@ -52,6 +52,10 @@ let conString = lam str.
 let varString = lam str.
   parserStr str "#var" (lam str. is_lower_alpha (head str))
 
+-- Label string parser translation for records
+let labelString = lam str.
+  parserStr str "#label" (lam str. is_lower_alpha (head str))
+
 let _ppLookupName = assocLookup {eq = nameEqSym}
 let _ppLookupStr = assocLookup {eq = eqstr}
 let _ppInsertName = assocInsert {eq = nameEqSym}
@@ -159,7 +163,7 @@ lang RecordPrettyPrint = RecordAst
         mapAccumL
           (lam env. lam r.
              match pprintCode innerIndent env r.1 with (env,str) then
-               (env, join [r.0, " =", newline innerIndent, str])
+               (env, join [labelString r.0, " =", newline innerIndent, str])
              else never)
           env t.bindings
       with (env,binds) then
@@ -170,7 +174,7 @@ lang RecordPrettyPrint = RecordAst
   | TmRecordUpdate t ->
     match pprintCode indent env t.rec with (env,rec) then
       match pprintCode indent env t.value with (env,value) then
-        (env,join ["{", rec, " with ", t.key, " = ", value, "}"])
+        (env,join ["{", rec, " with ", labelString t.key, " = ", value, "}"])
       else never
     else never
 end
@@ -422,7 +426,7 @@ lang RecordPatPrettyPrint = RecordPat
     match
       mapAccumL (lam env. lam r.
         match getPatStringCode indent env r.1 with (env,str) then
-          (env,join [r.0, " = ", str])
+          (env,join [labelString r.0, " = ", str])
         else never)
         env bindings
     with (env,binds) then
--- a/stdlib/mexpr/pprint.mc
+++ b/stdlib/mexpr/pprint.mc
@@ -189,13 +189,14 @@ lang RecordPrettyPrint = RecordAst
     else
       let innerIndent = incr (incr indent) in
       match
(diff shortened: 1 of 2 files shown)


stdlib/mexpr/pprint.mc
```diff
@@ -150,16 +154,41 @@ lang FunPrettyPrint = FunAst
...
+      None ()
+    else
+      let keys = assocKeys {eq=eqstr} t.bindings in
+      match all stringIsInt keys with false then None () else
+      let intKeys = map string2int keys in
+      let sortedKeys = sort subi intKeys in
+      -- Check if keys are a sequence 0..(n-1)
+      match and (eqi 0 (head sortedKeys))
+                (eqi (subi (length intKeys) 1) (last sortedKeys)) with true then
+        Some (map (lam key. assocLookupOrElse {eq=eqstr} (lam _. never)
```
elegios: I believe `never` is intended for cases that the compiler can see statically (through the type system/exhaustiveness checking most likely) will never happen, so this will probably be a compile time error once that is implemented.

stdlib/mexpr/pprint.mc
```diff
@@ -150,16 +154,41 @@ lang FunPrettyPrint = FunAst
...
+                                              (int2string key) t.bindings)
+                   sortedKeys)
```
lingmar: These lookups give quadratic complexity, while sorting the association map directly would give nlog(n), however, that would require `assocToList` or `assocSort` or similar, which I'm not sure that we want.

stdlib/mexpr/pprint.mc
```diff
@@ -150,16 +154,41 @@ lang FunPrettyPrint = FunAst
 end
 
 lang RecordPrettyPrint = RecordAst
+  sem record2tuple =
+  | TmRecord t ->
+    match t.bindings with [] then
+      None ()
+    else
+      let keys = assocKeys {eq=eqstr} t.bindings in
```
elegios: Does `assocKeys` guarantee that it produces a list with no duplicates? I'm guessing it depends on whether `AssocMap` was treated as an opaque datastructure and only `assoc*` functions were used, or if it was treated as on associative list.

stdlib/mexpr/pprint.mc
elegios: Given the most likely small sizes involved here I don't think it matters, but maybe worth leaving a comment noting the optimization opportunity, should it become relevant down the line?

stdlib/mexpr/pprint.mc
```diff
@@ -150,16 +154,41 @@ lang FunPrettyPrint = FunAst
 end
 
 lang RecordPrettyPrint = RecordAst
+  sem record2tuple =
```
dlunde: Do we really want `record2tuple` as a semantic function? I tend to only use semantic functions for functions that might be extended in some other context (this might be controversial, though), and `record2tuple` seems very specific.

Concretly, my suggestion is to change this to:
```
let record2tuple = lam tm.
  use RecordAst in
  match tm with TmRecord t then
    ...
  else error "Not a record"
```

dlunde (approved):
Looks good!

stdlib/mexpr/pprint.mc
lingmar: I assumed here that it's treated as an opaque datastructure, however, I don't know if we have a convention for that. In `pprintCode` for `TmRecord` we call `mapAccumL` directly on the associative list, so we are not consistent. Discuss during meeting perhaps?

stdlib/mexpr/pprint.mc
lingmar: That seems cleaner to me, I'll update!

stdlib/mexpr/pprint.mc
lingmar: Thanks, changed it into an `error`.

stdlib/mexpr/pprint.mc
lingmar: Now updated.

stdlib/mexpr/pprint.mc
dlunde: Take a look at `assocMapAccum` that I added previously. Maybe it makes sense to use it here?

lingmar: Thanks Daniel -- `assocMapAccum` did the trick of replacing `mapAccumL`, with the cost of an extra call to `assocValues` as `assocMapAccum` returns a new map. (I can revert this if we don't want this change.)

PR #164, opened 2020-09-10
Add ANF transformations

dlunde:
This PR adds ANF transformations, based on Flanagan et al. (1993). Only a subset of MExpr is currently supported (essentially the lambda calculus core), and the rest will be added in a future PR.


dlunde:
Resources:
- [The Essence of Compiling with Continuations](https://dl.acm.org/doi/10.1145/173262.155113)
- [A-Normalization: Why and How](http://matt.might.net/articles/a-normalization/)

br4sco:
Looks like a nice adaptation of the reference implementation.

PR #168, opened 2020-09-14
Add Either library

johnwikman:
Adds an Either library to stdlib. I know that @andersthune was supposed to start on this, but I need it for the refactoring that I am doing on #162. Feels better to put this as a separate PR instead of including it there.

This is more or less a rip-off from https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Either.html

Will add documentation to functions before the meeting on Tuesday.

### DIFF
--- a/stdlib/either.mc
+++ b/stdlib/either.mc
@@ -1,7 +1,7 @@
 -- Miking is licensed under the MIT license.
 -- Copyright (C) David Broman. See file LICENSE.txt
 --
--- The library defines the Either type and its two constructors: Left & Right.
+-- This library defines the Either type and its two constructors: Left & Right.
 
 include "option.mc"
 include "seq.mc"
--- a/stdlib/either.mc
+++ b/stdlib/either.mc
@@ -3,6 +3,7 @@
 --
 -- The library defines the Either type and its two constructors: Left & Right.
 
+include "option.mc"
 include "seq.mc"
 
 type Either a b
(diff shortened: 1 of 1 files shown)


aathn:
Nice! No problems on my part, I didn't have time to start working on this yet.

I would like to requst a few additional functions:
- `eitherBimap: (a -> c) -> (b -> d) -> Either a b -> Either c d`
- `eitherMap: (b -> c) -> Either a b -> Either a c`
- `eitherPure: b -> Either a b`
- `eitherAp: Either a (b -> c) -> Either a b -> Either a c`
- `eitherBind: Either a b -> (b -> Either a c) -> Either a c`

Having the functor, applicative and monad instances is up for discussion, but personally I feel that it would be nice to be able to use Either values as Result values, in which case these functions would be really nice to have

johnwikman:
>   * `eitherMap: (b -> c) -> Either a b -> Either a c`

Currently in the process in implementing separate `eitherLeftMap` and `eitherRightMap`. Is the shorthand something that we want?

>   * `eitherPure: b -> Either a b`

Couldn't this be just `Right v` instead of writing `eitherPure v`? I do not fully understand this. Is it because you want to be able to create statements such as `map eitherPure [1,2,3,4]` that gives you an `[Either a Int]` type?

I have no questions regarding the other functions, so I'll go ahead and implement them!

aathn:
Regarding `pure`, I guess it's not really necessary to have it as a separate function. The purpose would be both to be able to use it as a function, and to provide a semantic differentiation, where `pure` is semantically the monad identity whereas `Right` is just a constructor.

Regarding the names, I'm not sure. Having short names is always nice, but clarity may be more important. A user could always define some aliases if they wanted to use shorter names. On the other hand, it may be good to encourage consistent use of the right value as the 'result' by providing short-hands for the right versions

johnwikman:
In this case I would prefer to go with a pure Result approach instead:
```ml
type Result e a
con Error: e -> Result e a
con OK: a -> Result e a

let resultMap: (a -> b) -> Result e a -> Result e b
let resultMapErr: (e -> x) -> Result e a -> Result x a

let either2result: Either e a -> Result e a
let result2either: Result e a -> Either e a
```
Even though you can have the mnemonic that "Right is right", I still feel like it is best to keep Either unbiased to either case as it could be used for more than just encapsulating returned errors.

david-broman:
@johnwikman The unit tests do not pass for me when I tested this PR. Can you take a look? Thanks!

johnwikman:
@david-broman Should work now.

PR #169, opened 2020-09-15
Refactor decision points after symbolize change

lingmar:
What the title says, as well as a few bugfixes found along the way.

### DIFF
--- a/stdlib/mexpr/decision-points.mc
+++ b/stdlib/mexpr/decision-points.mc
@@ -20,6 +20,10 @@ let _eqn = lam n1. lam n2.
   else
     error "Name without symbol."
 
+let drecordproj_ = use MExprAst in
+  lam key. lam r.
+  nrecordproj_ (nameSym "x") key r
+
 lang HoleAst
   syn Expr =
   | TmHole {tp : Type,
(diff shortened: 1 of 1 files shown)


lingmar:
[This diff](https://github.com/miking-lang/miking/pull/169/commits/fb2f542bbf3d115f722df946c6974b4a1e51522e) is a bit more useful than the main one.

stdlib/mexpr/decision-points.mc
```diff
@@ -4,34 +4,56 @@ include "digraph.mc"
...
+      match pprintCode indent env1 h.depth with (env2, depthStr) then
+        (env2,
+          join ["Hole (",
+                strJoin ", " [getTypeStringCode indent h.tp, startStr, depthStr],
+                ")"])
+      else never
+    else never
 end
 
 -- Temporary workaround: uniquely labeled decision points
```
elegios, line 51: Is this due to some transformation needed pre-symbolize and thus not accurate any more, or is it referring to something else and still relevant?

lingmar: I need the hole and application AST nodes to have unique symbols, symbolize gives unique symbols to identifiers only (or did I misunderstand something?).

elegios: No, I think that's correct

david-broman: The test suite takes 14s for me when running with this PR, compared to around 4s before. @lingmar can you take a closer look at the test cases that you have added in this PR? I would prefer if the testing time does not increase significantly. Thanks!

lingmar: @david-broman I run basically the same tests on the same 17 ASTs as before. I did some detective work and currently  each AST takes ~0.6 s to test, while previously the time was ~0.2 s per AST. The majority of the runtime is spent in the `transform` function, i.e. it seems like `transform` just got ~3x slower! I am really surprised by this since the transformation is almost exactly like before. Part of the slowdown lies in that I have to run `symbolize` on the transformed AST (disabling this drops the runtime to ~0.4 s per AST). Another part, I suspect, comes from running `symbolize` on the test-ASTs (before the transform), however I couldn't easily test this hypothesis as I need them to be symbolized in order to run the test cases.

To decrease the runtime, perhaps I can do the transformation in a symbolize-preserving way so that I don't have to run `symbolize` on the transformed AST (tricky but possible), and/or remove some of the ASTs from the test cases.

EDIT: the `symbolize` on the test ASTs (before transform) does not seem to have a noticable impact on runtime (probably because they are quite small).

lingmar:
I have now updated so that the transformation no longer calls `symbolize`. In addition, the transformation tests are only run on a subset of the test ASTs, while the call graph tests are still run on all the ASTs. This makes sense as I designed most of the ASTs for testing the call graph construction.

On my machine the tests now run in about 2 s. @david-broman can you confirm if the runtime is low enough for you as well?

david-broman:
Good. The test time increased by 1.8 s for me, which I think is OK. I will merge!

PR #171, opened 2020-09-15
Update sundials interface reflecting changes in sundialsml.

br4sco:


david-broman:
Good. Can you describe which version that has changed (only Sundials ML)? Is another version of Sundials supported now? These things have implications on macOS installations.

br4sco:
So this is [sundialsml v4.1.0p0](https://github.com/inria-parkas/sundialsml/releases/tag/v4.1.0p0) which I have tested against [sundials 4.1.0](https://computing.llnl.gov/projects/sundials/sundials-software). We might want to hold off on merging this since the current sundialsml version in opam is 3.1.1. The reason I updated was that I had trouble getting the opam version working on my Fedora 32 system.

david-broman:
Should this PR perhaps be changed to a DRAFT PR?

br4sco:
> Should this PR perhaps be changed to a DRAFT PR?

I cannot, but I'll close it instead.

PR #172, opened 2020-09-15
Add optional using term to utest

larshum:
Adds an optional `using` argument to `utest`. This change does not affect tests using the original syntax. Example usage:

`utest 10 with 20 using lti in ()`

### DIFF
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -233,7 +233,7 @@ let rec map_tm f = function
     f (TmMatch(fi,map_tm f t1,p,map_tm f t2,map_tm f t3))
   | TmUse(fi,l,t1) -> f (TmUse(fi,l,map_tm f t1))
   | TmUtest(fi,t1,t2,tusing,tnext) ->
-    let tusing_mapped = Option.map (fun t -> map_tm f t) tusing in
+    let tusing_mapped = Option.map map_tm tusing in
     f (TmUtest(fi,map_tm f t1,map_tm f t2,tusing_mapped,map_tm f tnext))
   | TmNever(_) as t -> f t
 
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -130,7 +130,7 @@ and mlang   = Lang of info * ustring * ustring list * decl list
 and let_decl = Let of info * ustring * tm
 and rec_let_decl = RecLet of info * (info * ustring * tm) list
 and con_decl = Con of info * ustring * ty
-and utest_top = Utest of info * tm * tm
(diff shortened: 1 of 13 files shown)


david-broman:
Good. Can you add a number of test cases `utest` that is using the new feature as well? For instance in one of the files under `test/mexpr/`.

src/boot/lib/ast.ml
```diff
@@ -232,7 +232,9 @@ let rec map_tm f = function
   | TmMatch(fi,t1,p,t2,t3) ->
     f (TmMatch(fi,map_tm f t1,p,map_tm f t2,map_tm f t3))
   | TmUse(fi,l,t1) -> f (TmUse(fi,l,map_tm f t1))
-  | TmUtest(fi,t1,t2,tnext) -> f (TmUtest(fi,map_tm f t1,map_tm f t2,map_tm f tnext))
+  | TmUtest(fi,t1,t2,tusing,tnext) ->
+    let tusing_mapped = Option.map (fun t -> map_tm f t) tusing in
```
elegios: I suspect `(fun t -> map_tm f t)` can be written as `(map_tm f)`

src/boot/lib/mlang.ml
```diff
@@ -304,7 +304,7 @@ let rec desugar_tm nss env =
   | TmSeq(fi, tms) -> TmSeq(fi, Mseq.map (desugar_tm nss env) tms)
   | TmRecord(fi, r) -> TmRecord(fi, Record.map (desugar_tm nss env) r)
   | TmRecordUpdate(fi, a, lab, b) -> TmRecordUpdate(fi, desugar_tm nss env a, lab, desugar_tm nss env b)
-  | TmUtest(fi, a, b, body) -> TmUtest(fi, desugar_tm nss env a, desugar_tm nss env b, desugar_tm nss env body)
+  | TmUtest(fi, a, b, using, body) -> TmUtest(fi, desugar_tm nss env a, desugar_tm nss env b, using, desugar_tm nss env body)
```
elegios: I think this should be something like `Option.map (desugar_tm nss env) using` instead of just `using`, in case there's a `use` inside that part of the expression.

src/boot/lib/ast.ml
larshum: Haha good point.

src/boot/lib/mlang.ml
larshum: Ooh nice catch, I thought I had fixed all of those.

dlunde:
Reminder: remember to also update `utest` in the interpreter (`mexpr/ast.mc` and related files) in a future PR.

PR #175, opened 2020-09-16
Updated the syntax for multiline comments.

david-broman:
I have updated the syntax for multiline comments, to support syntax
```
*- 
  Text 
-*
```
Please see the examples in file `comments.mc`.

This PR also adds support for nested multiline comments in boot. 

Also, please check out the header syntax for copyright statement etc. If we can agree on this, I will update the headers in all files in this PR.

### DIFF
--- a/test/mexpr/match.mc
+++ b/test/mexpr/match.mc
@@ -1,7 +1,9 @@
--- Miking is licensed under the MIT license.
--- Copyright (C) David Broman. See file LICENSE.txt
---
--- Test integer primitives
+*-
+* Miking is licensed under the MIT license.
+* Copyright (C) David Broman. See file LICENSE.txt
+*
+* Test integer primitives
+-*
 
 mexpr
 
--- a/test/mexpr/match.mc
+++ b/test/mexpr/match.mc
@@ -1,9 +1,7 @@
-*-
-* Miking is licensed under the MIT license.
-* Copyright (C) David Broman. See file LICENSE.txt
-*
-* Test integer primitives
--*
+-- Miking is licensed under the MIT license.
+-- Copyright (C) David Broman. See file LICENSE.txt
+--
+-- Test integer primitives
 
 mexpr
 
--- a/src/boot/lib/lexer.mll
+++ b/src/boot/lib/lexer.mll
@@ -69,6 +69,7 @@ let filename = ref (us"")
 let rowno = ref 1
 let colno = ref 0
 let last_info = ref NoInfo
+let nesting_level = ref 0
 let utf8strlen s = Ustring.length (Ustring.from_utf8 s)
 let newrow() =
   incr rowno;
@@ -166,9 +167,11 @@ rule main = parse
       { colcount_fast s; main lexbuf }
   | line_comment
       { main lexbuf }
-  | "/*" as s
+  | "*-" as s
       { Buffer.reset string_buf ;
-	Buffer.add_string string_buf s; section_comment lexbuf;
+	Buffer.add_string string_buf s;
+        nesting_level := 0;
+        section_comment lexbuf;
 	count_utf8 (Buffer.contents string_buf);
 	main lexbuf}
   | tab
@@ -232,10 +235,19 @@ and parsestring = parse
 
 (* Section comment *)
 and section_comment = parse
-  | "*/" as s
-      { Buffer.add_string string_buf s }
+  | "-*" as s
+      { if !nesting_level = 0 then
+          Buffer.add_string string_buf s
+        else (
+          nesting_level := !nesting_level - 1;
+          section_comment lexbuf)
+      }
+  | "*-"
+      { nesting_level := !nesting_level + 1;
+        section_comment lexbuf
+      }
   | eof
-      { let s = Ustring.from_utf8 ("/*" ^ (Buffer.contents string_buf)) in
+      { let s = Ustring.from_utf8 ("*-" ^ (Buffer.contents string_buf)) in
 	raise (Lex_error (LEX_COMMENT_NOT_TERMINATED,ERROR,
 	 	 mkinfo_ustring s, [s])) }
   | _ as c
--- a/src/boot/lib/lexer.mll
+++ b/src/boot/lib/lexer.mll
@@ -167,7 +167,7 @@ rule main = parse
       { colcount_fast s; main lexbuf }
   | line_comment
       { main lexbuf }
-  | "*-" as s
+  | "/-" as s
       { Buffer.reset string_buf ;
 	Buffer.add_string string_buf s;
         nesting_level := 0;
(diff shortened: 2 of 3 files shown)


dlunde (approved):
Looks good, and I like the header syntax.

test/mexpr/match.mc
```diff
@@ -1,7 +1,9 @@
--- Miking is licensed under the MIT license.
--- Copyright (C) David Broman. See file LICENSE.txt
---
--- Test integer primitives
+*-
+* Miking is licensed under the MIT license.
+* Copyright (C) David Broman. See file LICENSE.txt
+*
+* Test integer primitives
+-*
```
johnwikman: Do we not want to indent the left stars by 1 space as well? Having
```
*-
*
*
*
-*
```
looks wierd to me since the "pattern" of stars going down a line doesn't hold.

dlunde: I prefer
```
*-
*
*
*
-*
```
over
```
 *-
 *
 *
 *
-*
```
for two reasons:
1. There is no wasted white space at the beginning of each row.
2. The last `*` nicely encapsulates the comment together with the first `-`. I don't mind that the last `*` is unaligned.

david-broman: I do not really like the indentation of one space (which is common in many C-style languages). 

If we would like symmetry and the starts to line up, the following is also possible :-)

```
*- *
* Miking is licensed under the MIT license.
* Copyright (C) David Broman. See file LICENSE.txt
* -*
```

david-broman: However, even if my last suggestion has symmetry, it is more complicated. I agree with Daniel's arguments, and I also prefer the original version:

```
*- 
* Miking is licensed under the MIT license.
* Copyright (C) David Broman. See file LICENSE.txt
-* 
```

johnwikman: I can agree that it is not very nice to have the one space indentation, but IMO it looks nicer than having a continuous column of stars be broken by what feels (to me) like an out of place dash character. I originally suggested the `*- block comment -*` syntax with the idea that the stars would align when spanning multiple lines, so I would prefer to either find a different style to format multiline block comments or to re-open #95. The suggestion with `*- *` -> `* -*` looks OK to me, but there is nothing to say that the left-most characters must be stars.

johnwikman: >   2. The last `*` nicely encapsulates the comment together with the first `-`. I don't mind that the last `*` is unaligned.

The idea with having the indentation at the top was that the function definition at the bottom would encapsulate the comment together with the opening symbol:
```
 *-
 * This is a function
 * comment for foo:
-*
let foo : String -> String -> Bool
```
In the sense that the comment flows down to the bottom instead of spilling over the top. But this does not work for standalone block comments.

johnwikman: Just changing the stars to dashes makes the block comment look better to me, but I think that there's probably a better solution than this.
```
*-
- This is a function
- comment for foo:
-*
let foo : String -> String -> Bool
```

david-broman: Now, I am actually wondering why we would use a multiline comment for this kind of header comments in the first place. The reason for having a multiline comment is to be able to comment out many lines of code without adding the comment characters on each line. However, this is actually exactly what we do in all the examples above, adding a special character on each line! Why are we then using multiline comments?

david-broman: To make it symmetric, we can just use single line comments. I think that this looks at least as nice as all the proposals above (actually nicer). For instance
```
-- Miking is licensed under the MIT license.
-- Copyright (C) David Broman. See file LICENSE.txt 
```

For a function comment, why are we using multiline comments in the first place? What is wrong with the following
```
-- This is a function
-- comment for foo:
let foo : String -> String -> Bool
```

I would actually sugget that we use `--` syntax for normal comments, and then have a special comment syntax e.g. `--|` for documentation comments similar to Doxygen. But, I would prefer a cleaner syntax than Doxygen. For instance:

```
--| Function foo creates something interesting and returns something more interesting.
--| bar: The interesting string that will come first.
--| x: Another interesting int that will be concatenated.
--| returns: The concatenated string.
let foo : String -> Int -> Bool =
  lam bar. lam x. 
    concat bar (int2string x)
```
Or, perhaps, just using `-|`

```
-| Function foo creates something interesting and returns something more interesting.
-| bar: The interesting string that will come first.
-| x: Another interesting int that will be concatenated.
-| returns: The concatenated string.
let foo : String -> Int -> Bool =
  lam bar. lam x. 
    concat bar (int2string x)
```

The compiler can then check that the documentation comment corresponds correctly to the actual function names (the lambdas) defined in the function.

Finally, since we do not need multiline comments for headers anymore, just for commenting out code, I would suggest a cleaner syntax:

```
  /- Some text that we want to comment out -/
```

david-broman: I have done some edits of the previous message. I'm not sure if you receive emails of the updates. I am not getting any email notifications on my own comments from GitHub, so I am not sure how it works.

johnwikman: I can also see single line comments as a viable option if done right. Rust does this wrong in my opinion. The three slashes for function documentation looks wonky to me.

Initially I like `--|` since it blockifies the comment and does not introduce any new comment style into the parser. Though when experimenting with comment styles I have felt that the bars do not hold up in the long run. But the examples you showed above looked good to me, so can probably work.

Seeing the new block comment style next to a line comment also makes me think that we should probably switch. For me on a Swedish keyboard layout it is easier to type `*- text -*` since the characters are closer on the keyboard, but on a US keyboard layout it is most likely the other way around.
```
/- New block comment style -/
-- Line comment

*- Old block comment style -*
-- Line comment
```

david-broman: A final alternative for the Doxygen style comments would be to just use three dashes `---` to mark that it concerns documentation, and not a normal comment.

```
--- Function foo creates something interesting and returns something more interesting.
--- bar: The interesting string that will come first.
--- x: Another interesting int that will be concatenated.
--- returns: The concatenated string.
let foo : String -> Int -> Bool =
  lam bar. lam x. 
    concat bar (int2string x)
```

johnwikman: The three dashes are a bit too subtle for me:
```
--- It can be difficult to
--- find the odd one out
--- when there is a bunch
--- of text to read at the
--  same time as trying to make
--- sure that you follow proper
--- formatting styles. Especially
--- when you are just skimming
--- thousands of lines of code.
```
Having `--*` could also be an alternative. Though it looks better in Sublimes tt font since it has a centered vertical alignment of `*`.
```
--* Function foo creates something interesting and returns something more interesting.
--* bar: The interesting string that will come first.
--* x: Another interesting int that will be concatenated.
--* returns: The concatenated string.
let foo : String -> Int -> String =
  lam bar. lam x. 
    concat bar (int2string x)
```

david-broman: The compiler can check that you do not mix `---` and `--` when it supposed to be a block comment. If you start with `---` it must be `---` until the function definition.

johnwikman: Another solution could be to have a header line and a footer line:
```
--*
-- This is function documentation
-- without any changes to the parser.
--*

Comment (david-broman) [reply test/mexpr/match.mc]:
```
-----------------------------------
-- This is function documentation
-- without any changes to the parser.
--
```

Comment (johnwikman) [reply test/mexpr/match.mc]: Return of the swords?
```
--/
-- This is function documentation
-- without any changes to the parser.
--\
```

Comment (david-broman) [reply test/mexpr/match.mc]: After a long discussion at the Miking meeting we decided to got for the syntax 

```
  /- some text -/
```
for a multiline comment. We also decided to not use multiline comments for documentation comments (in the style of Doxygen). We need further discussions for how we should design documentation comments. Elias will open a new issue about this.

PR #177, opened 2020-09-16
Finish remaining ANF fragments

dlunde:


lingmar:
I guess this is more of a question about ANF than about the implementation: it seems like the outermost application nodes don't get a unique identifier, like in the example below:
```
--- BEFORE ANF ---
addi
  (addi
    1
    2)
  1

--- AFTER ANF ---
let t =
  addi
    1
    2
in
addi
  t
  1
```
I would have expected the result to be
```
let t = addi 1 2 in
let t1 = addi t 1 in
t1
```
I.e., that the `addi t 1` also gets an identifier. Perhaps this is correct ANF (I haven't looked deeper into the definition), but I wonder because for me to use this I would need all the application nodes to have unique identifiers.

dlunde:
I believe this is only ever a problem for the tail position in the outermost expression. For example, if `addi t 1` would be used anywhere else, it would also be bound to a name.

A simple solution would be to add a "global" name in your environment that represents the outermost expression in tail position (e.g., `addi t 1`), or just simply do the transformation `addi t 1` -> `let t1 = addi t 1 in t1` once on the outermost expression after running ANF.

lingmar (approved):
Looks good to me!

dlunde:
> I believe this is only ever a problem for the tail position in the outermost expression. For example, if `addi t 1` would be used anywhere else, it would also be bound to a name.
> 
> A simple solution would be to add a "global" name in your environment that represents the outermost expression in tail position (e.g., `addi t 1`), or just simply do the transformation `addi t 1` -> `let t1 = addi t 1 in t1` once on the outermost expression after running ANF.

I've given this some more thought, and my comment above is most likely wrong. This PR currently implements the standard ANF transform, but I think we, in addition, need exactly what @lingmar pointed out for our static analyses. I'll take a look at this tomorrow (should be an easy fix).

dlunde:
Now fixed. All types of applications (regular, constructing records/sequences, using a match, etc.) will now also be bound in tail position. Example:
```
--- BEFORE ANF ---
1
  (2
    3)
  (4
    (5
      6))

--- AFTER ANF ---
let t =
  2
    3
in
let t1 =
  5
    6
in
let t2 =
  4
    t1
in
let t3 =
  1
    t
    t2
in
t3
```

PR #179, opened 2020-09-16
Split external functions into those that depend on external libraries and those who does not.

br4sco:
In this PR the external functions who does not depend on anything other than ocamls standard library are separated out from the interface to the sundials solver. These functions without additional dependencies are included in every build and are prefixed with `ext` to make them distinguishable from intrinsics in the core language. For convenience, these functions are aliased to prefix-less names in [./stdlib/ext/math.mc](./stdlib/ext/math.mc). Currently these functions are a handful of elementary functions over floats, which have also been lifted to dual numbers in [./stdlib/ad/dualnum-ext.mc](./stdlib/ad/dualnum-ext.mc).

As part of this PR there is also a bit of renaming to clarify the difference between the always included external functions and the sundials interface. Moreover, it includes a bit of code styling to make the code regarding the external functions more similar to the code in the rest of the project. In addition, code duplication of test fixtures in [./stdlib/ad](./stdlib/ad) have been reduced.


aathn:
After this PR, `exp` is available both as a regular builtin in `mexpr.ml` and as an external function `extExp` in `ext.ml`; I'm guessing it should be removed from one place or the other.

br4sco:
> After this PR, `exp` is available both as a regular builtin in `mexpr.ml` and as an external function `extExp` in `ext.ml`; I'm guessing it should be removed from one place or the other.

Aha, didn't realize it had been included as a builtin. I'm guessing it was never meant to be part of the core-language so I think it should be removed from `mexpr.ml` (If you do not have any objections @lingmar?).

PR #180, opened 2020-09-17
Update patterns in interpreter

elegios:
This is a followup to #158, adding support for the more fancy patterns we have throughout `pprint`, `smap`, `sfold`, `symbolize`, and `eval`.

I also did a few other things:
- split out tests for `smap` and `sfold` to deduplicate work between `ast.mc` and `ast-builder.mc`.
- two new helper functions for merging `AssocMap`s
- added `optionFoldlM` to `option.mc`


stdlib/mexpr/symbolize.mc
```diff
@@ -128,15 +129,20 @@ lang DataSym = DataAst
 end
 
 lang MatchSym = MatchAst
-  sem symbolizePat (env : Env) =
+  -- TODO(vipa): env is constant throughout symbolizePat, so it would be preferrable to pass it
+  -- in some other way, a reader monad or something. patEnv on the other hand changes, it would
+  -- be nice to pass via state monad or something.
+  -- env is the environment from the outside, plus the names added thus far in the pattern
+  -- patEnv is only the newly added names
+  sem symbolizePat (env : Env) (patEnv : Env) =
```
dlunde, line 138: Is it strictly needed have both `env` and `patEnv` in order to implement the new patterns? I find it cleaner to just pass in an environment which gets incrementally extended throughout `symbolize`, avoiding the need to merge the environments afterwards.

stdlib/option.mc
```diff
@@ -94,6 +94,21 @@ let optionMapM: (a -> Option b) -> [a] -> Option [b] = lam f. lam l.
 utest optionMapM (lam x. if gti x 2 then Some x else None ()) [3, 4, 5] with Some [3, 4, 5]
 utest optionMapM (lam x. if gti x 2 then Some x else None ()) [2, 3, 4] with None ()
 
+let optionFoldlM: (a -> b -> Option a) -> a -> [b] -> Option a = lam f.
+  recursive let recur = lam a. lam bs.
+    match bs with [b] ++ bs then
+      match f a b with Some a then
+        recur a bs
+      else None ()
+    else Some a
```
dlunde: Should probably be
```
else match bs with [] then
  Some a
else never
```
?

stdlib/mexpr/symbolize.mc
elegios: We need to introduce new symbols for names that are introduced in a pattern, but keep the same symbol if the same name is introduced multiple times. This shows up in `|` patterns, e.g., `"a" ++ rest | "b" ++ rest` should have the same symbol for both `rest` names. If we just pass along a single environment I don't think we can distinguish symbols introduced previously in the same pattern vs those introduced in a parent expression, e.g., `let a = 1 in match 2 with a then ...`.

stdlib/option.mc
elegios: This probably depends on what policy we want to have for exhaustive pattern matching. Writing the extra match and ending with a `never` would trigger our exhaustiveness analysis, but the semantics are otherwise the same. This is essentially equivalent with `match bs with | [b] ++ bs -> ... | _ -> ...`, while using `never` would instead be equivalent with `match bs with | [b] ++ bs -> ... | [] -> ...`. I would probably normally argue for the latter, I've just been lazy here.

Maybe worth deciding on a convention to follow generally?

stdlib/mexpr/symbolize.mc
dlunde: Ah, of course, the pattern names should introduce new symbols just as lambdas do. Then this is actually a bug in my first implementation!

stdlib/option.mc
dlunde: Probably a good idea to discuss at the next meeting.

dlunde (approved):

PR #183, opened 2020-09-20
Change snake_case function names to camelCase

aathn:
Some of these name changes might be controversial (for instance `gensym` -> `genSym`). I have tried to be as consistent as possible, but I may have missed something.

Edit: The changes to the conversion names are now undone, so the PR only updates snake-cased names to camel case.

### DIFF
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -76,12 +76,12 @@ and const =
 | Cfloorfi
 | Cceilfi
 | Croundfi
-| CInt2float
-| CString2float
+| Cint2float
+| Cstring2float
 (* MCore intrinsic: characters *)
 | CChar    of int
-| CChar2int
-| CInt2char
+| Cchar2int
+| Cint2char
 (* MCore intrinsic: sequences *)
 | CmakeSeq of int option
 | Clength
@@ -113,7 +113,7 @@ and const =
 | CSymb of int
 | Cgensym
 | Ceqsym of int option
-| CSym2hash
+| Csym2hash
 (* External functions TODO(?,?): Should not be part of core language *)
 | CExt of Extast.ext
 | CSd of Sdast.ext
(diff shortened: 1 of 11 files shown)


stdlib/char.mc
```diff
@@ -1,78 +1,78 @@
...
-let leqchar = lam c1. lam c2. leqi (char2int c1) (char2int c2)
-let geqchar = lam c1. lam c2. geqi (char2int c1) (char2int c2)
-let show_char = lam c. concat "'" (concat [c] "'")
+let eqchar = lam c1. lam c2. eqi (charToInt c1) (charToInt c2)
+let neqchar = lam c1. lam c2. neqi (charToInt c1) (charToInt c2)
+let ltchar = lam c1. lam c2. lti (charToInt c1) (charToInt c2)
+let gtchar = lam c1. lam c2. gti (charToInt c1) (charToInt c2)
+let leqchar = lam c1. lam c2. leqi (charToInt c1) (charToInt c2)
+let geqchar = lam c1. lam c2. geqi (charToInt c1) (charToInt c2)
+let showChar = lam c. concat "'" (concat [c] "'")
```
johnwikman, line 9: Should this perhaps be `charToString` instead? Feels more right to me since we have `intToString` instead of `showInt`.

src/boot/lib/ast.ml
```diff
@@ -112,9 +112,9 @@ and const =
 | Cexit
 (* MCore Symbols *)
 | CSymb of int
-| Cgensym
+| CgenSym
```
johnwikman: I we convert this to camel case, then I feel like it should probably be `symGen` instead.

johnwikman (commented):
In general, I think that this looks good. See my other comments for where I have questions about certain changes. I also prefer the new style of `typeToType` instead of the old conversion names.

Update: Changed my mind about `showChar`. That is probably fine as it is.

src/boot/lib/ast.ml
david-broman: Not sure about this one. Gensym is more of a word in PL and I would prefer to keep it as one word, i.e. `gensym`.

aathn: This change is now reverted, as discussed during the meeting

PR #188, opened 2020-09-28
Implemented ocaml pprinting of simple expressions, lambdas and lets.

br4sco:
This PR implement a language fragment and related pretty printer to generate `ocaml` code from a subset of `mexpr`. ~Currently names are not handled correctly in the pretty printer~. Currently simple expressions, abstractions, let expressions, and recursive expressions are supported. Tests covers expressions evaluating to `Int`, `Bool` and `Char` values.

Co-Authored-By: Linnea Ingmar <lingmar@kth.se>
Co-Authored-By: Lars Hummelgren <lasse.hummelgren@gmail.com>

### DIFF
--- a/stdlib/ocaml/generate.mc
+++ b/stdlib/ocaml/generate.mc
@@ -1,0 +1,17 @@
+include "mexpr/ast.mc"
+include "mexpr/ast-builder.mc"
+include "ocaml/ast.mc"
+include "ocaml/pprint.mc"
+
+lang OCamlGenerate = MExprAst + OCamlAst
+  sem generate =
+  | t -> smap_Expr_Expr generate t
+end
+
+lang OCamlTest = OCamlGenerate + OCamlPrettyPrint
+
+mexpr
+
+use OCamlTest in
+
+generate (addi_ (int_ 1) (int_ 2))
--- a/stdlib/ocaml/generate.mc
+++ b/stdlib/ocaml/generate.mc
@@ -2,6 +2,11 @@ include "mexpr/ast.mc"
 include "mexpr/ast-builder.mc"
 include "ocaml/ast.mc"
 include "ocaml/pprint.mc"
+include "mexpr/parser.mc"
+include "mexpr/symbolize.mc"
+include "mexpr/eval.mc"
+include "mexpr/eq.mc"
+
 
 let escapeFirstChar = lam c.
   if or (isLowerAlpha c) (eqChar c '_') then c
@@ -55,7 +60,8 @@ lang OCamlGenerate = MExprAst + OCamlAst
   | t -> smap_Expr_Expr generate t
 end
 
-lang OCamlTest = OCamlGenerate + OCamlPrettyPrint
+lang OCamlTest = OCamlGenerate + OCamlPrettyPrint + MExprSym + ConstEq + IntEq
+                 + BoolEq + CharEq
 
 mexpr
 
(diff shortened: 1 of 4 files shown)


stdlib/ocaml/pprint.mc
```diff
@@ -0,0 +1,124 @@
...
+  | CLtf _ -> "<"
+  | CChar {val = c} -> show_char c
+
+  sem pprint (indent : Int) =
+  | TmApp {lhs = l, rhs = r} ->
+    join ["(", pprint indent l, ")", newline (incr indent),
+          "(", pprint (incr indent) r, ")"]
+  | TmConst {val = c} -> pprintConst c
+  | TmLam {ident = id, body = b} ->
+    join ["fun ", nameGetStr id, " ->", newline (incr indent),
```
elegios: This (and other places using `nameGetStr` I suppose) assumes that our scoping rules are the same as OCaml's (which I think is true, but it might be worth being explicit about it, making sure we use unique names everywhere), but more importantly that our identifiers are the same lexically. We'll need to do some form of "escaping", since we allow any form of string at all.

johnwikman:
I am wondering about the semantic name "pprint" as it is used here. Because we are not pretty printing an OCaml AST, we are converting the MExpr AST into the OCaml code. Feels more like a light-weight codegen.

I suppose that this could also create conflict with the actual pretty printer.

br4sco:
> I am wondering about the semantic name "pprint" as it is used here. Because we are not pretty printing an OCaml AST, we are converting the MExpr AST into the OCaml code. Feels more like a light-weight codegen.
> 
> I suppose that this could also create conflict with the actual pretty printer.

br4sco:
Missclick ^^

br4sco:
> I am wondering about the semantic name "pprint" as it is used here. Because we are not pretty printing an OCaml AST, we are converting the MExpr AST into the OCaml code. Feels more like a light-weight codegen.
> 
> I suppose that this could also create conflict with the actual pretty printer.

I don't know, we could view the AST defined by OCamlAst as a language, then "pprint" wouldn't be such a bad name. We just won't implement a parser for this language. The conversion part would be to convert between the MExpr Ast and the OCaml Ast. For the current subset there's not much converting but there will be for other language fragments such as patterns and pattern matching.

br4sco:
However, I think we should rename the language fragment to _OCamlish_ as we discussed (or something similar) since it only encodes a subset of OCaml.

stdlib/ocaml/generate.mc
```diff
@@ -0,0 +1,170 @@
...
+
+let compareInt1 = eqi_ (int_ 1) (int_ 2) in
+utest compareInt1 with generate (symbolize compareInt1) using sameSemantics in
+
+let compareInt2 = lti_ (addi_ (int_ 1) (int_ 2)) (int_ 3) in
+utest compareInt2 with generate (symbolize compareInt2) using sameSemantics in
+
+let charLiteral = char_ 'c' in
+utest charLiteral with generate (symbolize charLiteral) using sameSemantics in
+
```
elegios, line 255: There should probably be tests using `sameSemantics` containing lambdas and lets (not as return values, but as part of the expressions), probably also using shadowing and the like.

stdlib/ocaml/generate.mc
```diff
@@ -0,0 +1,190 @@
...
+      in
+      match ocamlVal with TmConst {val = CChar _} then
+        eqExpr mexprVal ocamlVal
+      else error "Values mismatch"
+    else error "Unsupported constant"
+  else error "Unsupported value"
+in
+
+let addInt1 = addi_ (int_ 1) (int_ 2) in
+utest addInt1 with generate (symbolize addInt1) using sameSemantics in
```
elegios, line 192: Point worth noting based on this that will have (minor) implications on typechecking: `utest using` isn't necessarily reflexive, and further doesn't require the same type on the two things being compared.

elegios (changes requested):

stdlib/ocaml/generate.mc
br4sco: Added some more tests.

stdlib/ocaml/generate.mc
br4sco: I guess that is a future headache, or do you propose to change something?

elegios: Just a thing for the future

elegios (approved):

PR #189, opened 2020-09-29
Alpha equivalence for MExpr

dlunde:
An implementation of alpha equivalence for MExpr. As an example, this enables the following:
```
utest var_ "x" with var_ "y" using eqmexpr
```
with the new `using` syntax from #172.

It is particularly useful if you are working on some type of MExpr transformation for which you want to do unit testing.

### DIFF
--- a/stdlib/mexpr/ast.mc
+++ b/stdlib/mexpr/ast.mc
@@ -257,6 +257,7 @@ end
 type PatName
 con PName     : Name -> PatName
 con PWildcard : ()   -> PatName
+
 lang VarPat
   syn Pat =
   | PVar {ident : PatName}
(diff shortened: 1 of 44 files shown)


dlunde:
I should also mention that this PR includes some additional cleanup (besides adding the alpha equivalence). I did not separate these changes into another PR for practical reasons.

stdlib/mexpr/ast.mc
```diff
@@ -206,7 +206,7 @@ end
 lang BoolAst = ConstAst
   syn Const =
   | CBool {val : Bool}
-  | CNot {}
+  | CNot {} -- TODO dlunde 2020-09-29: This constant does not exist in boot. Remove?
```
lingmar: Not 100% sure, but I think we settled on using parentheses for TODOs, i.e. `TODO(dlunde, ...)`

PR #190, opened 2020-09-29
Mcore parsing

david-broman:
This PR includes a first version of a partial parser for MExpr. Features that are included so far are:

- Lambdas
- If expressions
- Integer literals
- String literals
- Boolean literals
- Sequences
- Single line and multi-line comments
- Variables

Besides MExpr, there is an extension that handles infix expressions.


stdlib/mexpr/ast.mc
```diff
@@ -21,21 +25,28 @@ end
...
+
 end
 
 lang FunAst = VarAst + AppAst
   syn Expr =
   | TmLam {ident : Name,
            tpe   : Option,
-           body  : Expr}
+           body  : Expr,
+	   fi    : Info}
```
elegios: This seems strangely indented. Mixed tabs and spaces?

stdlib/mexpr/extensions.mc
```diff
@@ -0,0 +1,121 @@
...
+  | op ->
+    let p2 = advanceCol p 1 in
+    Some {
+      val = lam x. lam y.
+        let op = TmConst {val = op, fi = makeInfo p p2} in
+        let app = lam x. lam y. 
+                TmApp {lhs = x, rhs = y, fi = NoInfo ()} in
+                --TmApp {lhs = x, rhs = y, fi = mergeInfo (info x) (info y)} in
+		--
+		--BUG: the above should work, but the info function does not exist
```
elegios: Is this still true?

stdlib/mexpr/parser.mc
```diff
@@ -0,0 +1,433 @@
...
+    parseInfix r2.pos prec exp r2.str 
+
+  sem parseInfix (p: Pos) (prec: Int) (exp: Expr) =
+
+  sem parseExprImp (p: Pos) =
+  | _ -> posErrorExit p "Parse error. Unknown character sequence."
+end
+
+
+-- Include this fragment is there are no infix operations
```
elegios: Is this "include if" or "include iff"? If it's the latter it's a bit awkward that we then can't add infix operators to a language that lacks them, since it will already have this fragment included.

stdlib/mexpr/parser.mc
```diff
@@ -0,0 +1,433 @@
...
+  | "false" ++ xs ->
+      let p2 = advanceCol p 5 in
+      {val = TmConst {val = CBool {val = false}, fi = makeInfo p p2}, pos = p2, str = xs}
+end
+
+
+-- Parsing of an unsigned integer
+lang UIntParser = ExprParser + ConstAst + IntAst
+  sem parseExprImp (p : Pos) =
+  | ("0" ++ s) & xs | ("1" ++ s) & xs | ("2" ++ s) & xs | ("3" ++ s) & xs | ("4" ++ s) & xs |
```
elegios: This could be shortened to `(['0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'] ++ s) & xs`

Sidenote: I guess it might be nice to have ranges in patterns.

stdlib/mexpr/parser.mc
```diff
@@ -0,0 +1,433 @@
...
+       {pos = advanceCol r.pos l, str = slice r.str l (subi (length r.str) l)}
+     else
+       posErrorExit r.pos (join ["Unknown character. Expected '", keyword, "'."])
+end
+
+
+-- Parsing if expressions
+lang IfParser = ExprParser + KeywordParser +  MatchAst + BoolPat
+  sem parseExprImp (p: Pos) =
+  | "if" ++ xs ->
```
elegios: Does this handle parsing `"iftrue 4"` properly (note the missing space, it's a function application of the function `iftrue` to `4`)? I feel like it will trigger on any identifier beginning with `if`.

stdlib/mexpr/parser.mc
```diff
@@ -0,0 +1,433 @@
...
+    match str with "\\" ++ xs then
+      match xs with "\\" ++ xs then ret '\\' xs 2 else
+      match xs with "n" ++ xs then ret '\n' xs 2 else
+      match xs with "t" ++ xs then ret '\t' xs 2 else
+      match xs with "\"" ++ xs then ret '\"' xs 2 else 
+      match xs with "'" ++ xs then ret '\'' xs 2 else
+      posErrorExit (advanceCol p 1) "Unknown escape character."
+    else match str with [x] ++ xs then ret x xs 1
+    else posErrorExit p "Unexpected end of file."
+    -- TODO David (2020-09-27): Shoud we allow newlines etc. inside strings
```
elegios: I believe the todo style we settled on was not quite this (`TODO(David, 2020-09-27): ...`)

stdlib/mexpr/extensions.mc
david-broman: The problem seems to be gone with your latest bugfix Viktor.

stdlib/mexpr/parser.mc
david-broman: Yes, this how it has to be done with the current design in MLang. If we extend MLang patterns to make a difference between pattern

```
| e -> ...
```

and 
```
| _ -> ...
```
where the former has higher priority, we can make a better design. I would suggest that we add this to MLang patterns.

stdlib/mexpr/parser.mc
david-broman: Good point. I was not aware that we allowed OR patterns inside sequences. Also, I agree, it would be good if we add range patterns: both for simplicity when writing out this kind of pattern, and also for compilation: it might be more efficient to check for ranges.

stdlib/mexpr/parser.mc
david-broman: Good. I have updated to the style we decided on at the meeting.

stdlib/mexpr/parser.mc
david-broman: Good point. It is now fixed.

stdlib/mexpr/parser.mc
```diff
@@ -0,0 +1,448 @@
...
+let parseIdent = lam upper. lam p. lam str.
+  recursive
+  let work = lam acc. lam first. lam p. lam str.
+    match str with [x] ++ xs then
+      let c = char2int x in
+      let m1 = or (not first) upper in
+      let m2 = or (not first) (not upper) in
+      if or (or (and m1 (and (geqi c 65) (leqi c 90)))
+                (and m2 (or (eqi c 95) (and (geqi c 97) (leqi c 122)))))
+	    (and (not first) (and (geqi c 48) (leqi c 57)))
```
johnwikman: I am not so keen on having these magic numbers directly in the code. Would it perhaps be better to use something like `leqChar` and `geqChar`? In that case you actually see the characters directly, so that you do not have to keep looking up the ASCII table when debugging. Though I would prefer to have something like `isUpper`, `isLower`, and `isDigit`.

david-broman: OK, I fixed it now and used the functions in char.mc.

PR #191, opened 2020-09-30
Fix merge order dependence

elegios:
This PR fixes the issue mentioned [here](https://github.com/miking-lang/miking/pull/189/files#diff-91ca39c73d626275340a6b6bc508e956R578-R580) (in #189), and probably the issue mentioned [here](https://github.com/miking-lang/miking/pull/190/files#diff-1cbc9131f0aa851151ad313dead4e210R31-R32) (in #190), though I could not reproduce the latter.

The bug arose because of an optimization I did for the `merge_inter_` function in `merge_lang_data` in `mlang.ml`. I make the assumption that if the two `sem` functions we are merging have the same `info`, then they are the same and we can just return the first `sem`. The thought here is that the `info` refers to where the function is defined, and I rather carefully make sure that the returned `sem` has the `info` from the first `sem`, not the second. This means that the copy of the `sem` function that lives in a particular language fragment (i.e., the flattened one with all cases in all its includes) will have the `info` of the `sem` that is written inside that fragment. The issue is that there might be no such `sem`, e.g., in #189 the language fragments that expose the bug look like this:

```
lang PPrintLang = MExprPrettyPrint + LAppPrettyPrint + LHoleAstPrettyPrint

lang TestLang = MExpr + ContextAwareHoles + LAppEval + PPrintLang
```

In this case the `None, Some a -> Some a` case fires, in which case the info for the `sem` function is that of the first fragment that had the function defined, in this case `MExprPrettyPrint` (which is also without its own definition, so it looks like it's defined in `VarPrettyPrint`). Somehow (I'm a little bit too lazy to track things fully) one of the other languages included in `TestLang` looks like it has a `pprintCode` defined in the same place, whereby the `pprintCode` from `PPrintLang` is never included, because it looks like it's defined at the same place, thus `TestLang` does not have cases for the decision point constructors in `pprintCode`. 

(sidenote: @lingmar, the `expr2str` between these two fragments appears to be dead code, `TestLang` contains an `expr2str` which shadows this one when it is `open`ed)

-------------------------------------------

I also leave the functions I wrote to debug this, mostly debug printing stuff (not actually called, just defined), as well as a fix in `pprint.ml` to print symbols for `let`, `lam`, and `recursive let` if `--symbol` is given.

### DIFF
--- a/src/boot/lib/mlang.ml
+++ b/src/boot/lib/mlang.ml
@@ -65,6 +65,16 @@ type lang_data = {
     syns: (info * cdecl list) Record.t;
   }
 
+let spprint_inter_data ({info; cases}): ustring =
+  List.map (fun (fi, {pat}) -> us"  " ^. ustring_of_pat pat ^. us" at " ^. info2str fi) cases
+  |> Ustring.concat (us"\n")
+  |> (fun msg -> us"My location is " ^. info2str info ^. us"\n" ^. msg)
+
+let spprint_lang_data ({inters}): ustring =
+  Record.bindings inters
+  |> List.map (fun (name, data) -> name ^. us"\n" ^. spprint_inter_data data)
+  |> Ustring.concat (us"\n")
+
 let compute_order fi ((fi1, {pat = pat1; pos_pat = p1; neg_pat = n1; _}), (fi2, {pat = pat2; pos_pat = p2; neg_pat = n2; _})) =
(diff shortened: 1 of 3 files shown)


dlunde:
Great! Regarding the @lingmar sidenote: this might be because I made expr2str a semantic function. Previously, it was a regular function, which is probably why Linnea had to redefine it for her extended language.

PR #193, opened 2020-09-30
Expand hashmap library with more functionality

johnwikman:
Adds a couple of convenience functions to hashmap.mc, as well as fixing the type signatures and revamping some old functions.

### DIFF
--- a/stdlib/hashmap.mc
+++ b/stdlib/hashmap.mc
@@ -3,10 +3,12 @@
 --
 -- A simple generic hashmap library.
 --
--- TODO(?,?):
---  - Resizing of buckets.
---  - Conversion to and from association lists.
+-- TODO(johnwikman, 2020-08-05): Resizing of buckets.
 --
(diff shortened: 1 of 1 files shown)


stdlib/hashmap.mc
```diff
@@ -30,12 +30,12 @@ let _hashmapBucketIdx = lam hash. lam hm. modi (absi hash) (length hm.buckets)
...
 
 -- 'hashmapEmpty' is an empty hashmap with a default number of buckets.
-let hashmapEmpty : HashMap =
+let hashmapEmpty : HashMap k v =
   {buckets = makeSeq _hashmapDefaultBucketCount [],
    nelems = 0}
 
 -- 'hashmapStrTraits' is traits for a hashmap with strings as keys.
-let hashmapStrTraits : HashMapTraits =
+let hashmapStrTraits : HashMapTraits String v =
```
lingmar: Should be just `HashMapTraits String` I guess

johnwikman: Yes, fixed it now!

PR #194, opened 2020-10-03
Update eq.mc on alpha equivalence

gizemcaylak:
Completed SeqTotPatEq, SeqEdgPatEq, AndPatEq, OrPatEq, NotPatEq and added relevant tests

### DIFF
--- a/stdlib/mexpr/eq.mc
+++ b/stdlib/mexpr/eq.mc
@@ -369,9 +369,10 @@ lang SeqTotPatEq = SeqTotPat
       if eqi (length ps2) (length ps1) then
         let z = zipWith (lam p1. lam p2. (p1,p2)) ps1 ps2 in
         optionFoldlM (lam fpEnv. lam ps.
-            match fpEnv with (f,p) then
-              eqPat env f p ps.0 ps.1 else None ())
-               (free,patEnv) z
+          match fpEnv with (f,p) then
+            eqPat env f p ps.0 ps.1 
+          else never)
+          (free,patEnv) z
       else None ()
     else None ()
 end
@@ -385,17 +386,19 @@ lang SeqEdgPatEq = SeqEdgePat
       match _eqpatname patEnv free mid1 mid2 with Some (f,p) then
         if eqi (length pre1) (length pre2) then
           let z1 = zipWith (lam p1. lam p2. (p1,p2)) pre1 pre2 in
-    let z2 = zipWith (lam p1. lam p2. (p1,p2)) post1 post2 in
-    match optionFoldlM (lam fpEnv. lam ps.
-              match fpEnv with (f,p) then
-                eqPat env f p ps.0 ps.1 else None ())
-                 (free,patEnv) z1 with Some (f1,p1) then
-      if eqi (length post1) (length post2) then
-        optionFoldlM (lam fpEnv. lam ps.
-              match fpEnv with (f,p) then
-                eqPat env f p ps.0 ps.1 else None ())
-                 (free,patEnv) z2
-      else None ()
+          let z2 = zipWith (lam p1. lam p2. (p1,p2)) post1 post2 in
+          match optionFoldlM (lam fpEnv. lam ps.
+            match fpEnv with (f,p) then
+              eqPat env f p ps.0 ps.1 
+            else never)
+            (free,patEnv) z1 with Some (f1,p1) then
+              if eqi (length post1) (length post2) then
+                optionFoldlM (lam fpEnv. lam ps.
+                  match fpEnv with (f,p) then
+                    eqPat env f p ps.0 ps.1 
+                  else never)
+                  (free,patEnv) z2
+              else None ()
           else None ()
         else None ()
       else None ()
@@ -458,9 +461,6 @@ end
 lang AndPatEq = AndPat
   sem eqPat (env : Env) (free : Env) (patEnv : NameEnv) (lhs : Pat) =
   | PAnd {lpat = l2, rpat = r2} ->
-    -- NOTE(gcaylak,2020-10-02): This check assumes order is important.
-    -- So 'l1 & r1 != l2 & r2' in terms of pattern matching.
-    -- Should we change this with irrelevant order?
     match lhs with PAnd {lpat = l1, rpat = r1} then
       match eqPat env free patEnv l1 l2 with Some (free,patEnv) then
         eqPat env free patEnv r1 r2
@@ -471,9 +471,6 @@ end
 lang OrPatEq = OrPat
   sem eqPat (env : Env) (free : Env) (patEnv : NameEnv) (lhs : Pat) =
   | POr {lpat = l2, rpat = r2} ->
-    -- NOTE(gcaylak,2020-10-02): This check assumes order is important.
-    -- So 'l1 | r1 != l2 | r2' in terms of pattern matching.
-    -- Should we change this with irrelevant order?
     match lhs with POr {lpat = l1, rpat = r1} then
       match eqPat env free patEnv l1 l2 with Some (free,patEnv) then
         eqPat env free patEnv r1 r2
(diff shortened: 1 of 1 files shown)


stdlib/mexpr/eq.mc
```diff
@@ -362,14 +362,44 @@ lang VarPatEq = VarPat
...
+              eqPat env f p ps.0 ps.1 else None ())
+               (free,patEnv) z
+      else None ()
+    else None ()
 end
 
-lang SeqEdgPatEq
+lang SeqEdgPatEq = SeqEdgePat
+  -- NOTE(gcaylak,2020-10-02): 'SeqEdgPatEq' is called 'SeqEdgePat' in 'ast.mc'.
+  -- Should we change 'SeqEdgPatEq' to 'SeqEdgePatEq' as well?
```
elegios: I originally named the pattern "edg" to match the length of "tot", but later regretted it as a poor abbreviation, and so changed to "edge"

stdlib/mexpr/eq.mc
```diff
@@ -362,14 +362,44 @@ lang VarPatEq = VarPat
...
+lang SeqTotPatEq = SeqTotPat
   sem eqPat (env : Env) (free : Env) (patEnv : NameEnv) (lhs : Pat) =
-  -- TODO(dlunde,2020-09-29)
+  | PSeqTot {pats = ps2} ->
+    match lhs with PSeqTot {pats = ps1} then
+      if eqi (length ps2) (length ps1) then
+        let z = zipWith (lam p1. lam p2. (p1,p2)) ps1 ps2 in
+        optionFoldlM (lam fpEnv. lam ps.
+            match fpEnv with (f,p) then
+              eqPat env f p ps.0 ps.1 else None ())
```
elegios: I believe this should be `else never`, since the pattern should be irrefutable. (there are a couple more places where this is true as well)

stdlib/mexpr/eq.mc
```diff
@@ -362,14 +362,44 @@ lang VarPatEq = VarPat
...
+  -- NOTE(gcaylak,2020-10-02): 'SeqEdgPatEq' is called 'SeqEdgePat' in 'ast.mc'.
+  -- Should we change 'SeqEdgPatEq' to 'SeqEdgePatEq' as well?
   sem eqPat (env : Env) (free : Env) (patEnv : NameEnv) (lhs : Pat) =
-  -- TODO(dlunde,2020-09-29)
+  | PSeqEdge {prefix = pre2, middle = mid2, postfix = post2} ->
+    match lhs with PSeqEdge {prefix = pre1, middle = mid1, postfix = post1} then
+      match _eqpatname patEnv free mid1 mid2 with Some (f,p) then
+        if eqi (length pre1) (length pre2) then
+          let z1 = zipWith (lam p1. lam p2. (p1,p2)) pre1 pre2 in
+    let z2 = zipWith (lam p1. lam p2. (p1,p2)) post1 post2 in
```
elegios: I think the indentation is weird here

stdlib/mexpr/eq.mc
```diff
@@ -425,19 +455,38 @@ lang BoolPatEq = BoolPat
...
 end
 
-lang OrPatEq
+lang OrPatEq = OrPat
   sem eqPat (env : Env) (free : Env) (patEnv : NameEnv) (lhs : Pat) =
-  -- TODO(dlunde,2020-09-29)
+  | POr {lpat = l2, rpat = r2} ->
+    -- NOTE(gcaylak,2020-10-02): This check assumes order is important.
+    -- So 'l1 | r1 != l2 | r2' in terms of pattern matching.
+    -- Should we change this with irrelevant order?
```
elegios: This check is just for alpha equivalence, so having the order matter is correct, since the only thing we're supposed to be flexible in is naming. There will be another separate analysis down the line that can check for pattern equivalence

stdlib/mexpr/eq.mc
gizemcaylak: Thanks, fixed it now!

stdlib/mexpr/eq.mc
```diff
@@ -362,14 +362,47 @@ lang VarPatEq = VarPat
...
+  -- Should we change 'SeqEdgPatEq' to 'SeqEdgePatEq' as well?
   sem eqPat (env : Env) (free : Env) (patEnv : NameEnv) (lhs : Pat) =
-  -- TODO(dlunde,2020-09-29)
+  | PSeqEdge {prefix = pre2, middle = mid2, postfix = post2} ->
+    match lhs with PSeqEdge {prefix = pre1, middle = mid1, postfix = post1} then
+      match _eqpatname patEnv free mid1 mid2 with Some (f,p) then
+        if eqi (length pre1) (length pre2) then
+          let z1 = zipWith (lam p1. lam p2. (p1,p2)) pre1 pre2 in
+          let z2 = zipWith (lam p1. lam p2. (p1,p2)) post1 post2 in
+          match optionFoldlM (lam fpEnv. lam ps.
```
dlunde: I'd prefer binding the match target (starting with `optionFoldlM (lam fpEnv.`) to a variable before the match. It is difficult to read at the moment.

stdlib/mexpr/eq.mc
```diff
@@ -362,14 +362,47 @@ lang VarPatEq = VarPat
...
+              eqPat env f p ps.0 ps.1 
+            else never)
+            (free,patEnv) z1 with Some (f1,p1) then
+              if eqi (length post1) (length post2) then
+                optionFoldlM (lam fpEnv. lam ps.
+                  match fpEnv with (f,p) then
+                    eqPat env f p ps.0 ps.1 
+                  else never)
+                  (free,patEnv) z2
+              else None ()
```
dlunde: Indentation looks a bit strange here maybe (for the uppermost `else None ()`)

dlunde (changes requested):

stdlib/mexpr/eq.mc
dlunde: I believe this is my error (I copied the names from `ast.mc` before Viktor's changes). Could you fix this Gizem?

gizemcaylak: Yes, I will fix that.

stdlib/mexpr/eq.mc
```diff
@@ -425,19 +457,32 @@ lang BoolPatEq = BoolPat
...
 end
 
-lang AndPatEq
+lang AndPatEq = AndPat
   sem eqPat (env : Env) (free : Env) (patEnv : NameEnv) (lhs : Pat) =
-  -- TODO(dlunde,2020-09-29)
+  | PAnd {lpat = l2, rpat = r2} ->
+    match lhs with PAnd {lpat = l1, rpat = r1} then
+      match eqPat env free patEnv l1 l2 with Some (f,p) then
+        eqPat env free patEnv r1 r2
```
elegios, line 465: It seems weird to not use `(f, p)` at all, is that intentional? I'd've thought it would work the same as names in, e.g., `SeqTotPatEq`, where the values are treated as an accumulator, forwarded left to right through the child patterns.

david-broman:
I could not see any tests where the patterns are used within match expressions, including the use of names. Are there such tests?

gizemcaylak:
> I could not see any tests where the patterns are used within match expressions, including the use of names. Are there such tests?

Yes, the tests including 'pgen' (`let pgen = lam p. match_ (int_ 1) p (int_ 2) (int_ 3) in`) are tests within match expressions
e.g.
```
let pvar1 = pvar_ "x" in
let pvar2 = pvar_ "y" in
let pdata1 = pcon_ "Const1" (pcon_ "Const2" prec1) in
let pdata2 = pcon_ "Const2" (pcon_ "Const1" prec1) in
let pand10 = pand_ pvar1 pdata1 in
let pand11 = pand_ pvar2 pdata2 in
```
If that is what you mean. But, I need to fix the point @elegios indicated on environments

stdlib/mexpr/eq.mc
gizemcaylak: I see, then I should left it as it is before (the order of the binding should matter?):
```
match eqPat env free patEnv l1 l2 with Some (free, patEnv) then
        eqPat env free patEnv r1 r2
```
 I made the same mistake in SeqEdgePatEq, I will fix that also. Thanks!

david-broman:
@gizemcaylak OK, good. Yes, that is what I mean.

PR #197, opened 2020-10-07
Changed scope of functions in pprint.mc

br4sco:
This PR changes the scope (i.e. name) of some functions in `pprint.mc`. Some of these functions will be used in the ocaml pprint language fragment. One function (`varString`) was already used in `eval.mc`. Other functions are exposed since they are naturally exposed as a group (e.g. pprint environment manipulation functions and identifier parser functions). 

I also removed the `MExprPrettyPrint` language fragment from the `MExpSym` language fragment. This inhibited you from constructing a new language from `MExprPrettyPrint` fragments and at same time include `MExpSym`.

### DIFF
--- a/stdlib/mexpr/eval.mc
+++ b/stdlib/mexpr/eval.mc
@@ -55,7 +55,7 @@ lang VarEval = VarAst
     match _evalLookup ident ctx.env with Some t then
       eval ctx t
     else
-      error (concat "Unknown variable: " (varString ident))
+      error (concat "Unknown variable: " (pprintVarString ident))
 end
 
 lang AppEval = AppAst
--- a/stdlib/mexpr/pprint.mc
+++ b/stdlib/mexpr/pprint.mc
@@ -21,8 +21,6 @@ let pprintNewline = lam indent. concat "\n" (pprintSpacing indent)
 -- Increment [indent] by 2
 let pprintIncr = lam indent. addi indent 2
 
-let symbolDelim = "'"
-
 ---------------------------------
 -- PRETTY PRINTING ENVIRONMENT --
 ---------------------------------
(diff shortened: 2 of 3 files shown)


elegios:
> I also removed the `MExprPrettyPrint` language fragment from the `MExpSym` language fragment. This inhibited you from constructing a new language from `MExprPrettyPrint` fragments and at same time include `MExpSym`.

This is probably a good change to not include more than you actually want in the `MExprSym` fragment, but it shouldn't cause issues to include a fragment twice, that should be idempotent. Did something break without this change?

br4sco:
> > I also removed the `MExprPrettyPrint` language fragment from the `MExpSym` language fragment. This inhibited you from constructing a new language from `MExprPrettyPrint` fragments and at same time include `MExpSym`.
> 
> This is probably a good change to not include more than you actually want in the `MExprSym` fragment, but it shouldn't cause issues to include a fragment twice, that should be idempotent. Did something break without this change?

So this was our problem, a bit simplified:

```
include "mexpr/ast.mc"
include "mexpr/pprint.mc"
include "mexpr/symbolize.mc"

lang MyPPrint = ConstAst + MExprSym
  sem pprintCode (indent : Int) (env : Env) =
  | TmConst _ -> "const!"
end

mexpr
()
```

We wanted to redefine `pprintCode` for some AST fragments but not all of them. Since `TmConst` appears in `pprintCode` in both `MyPPrint` and in `MExprPrettyPrint`, which is included in `MExprSym`, we got:

```
FILE "/home/oerikss/src/miking/test/mlang/test.mc" 6:2-6:45 ERROR: Patterns at FILE "/home/oerikss/src/miking/test/mlang/test.mc" 7:4-7:13 and FILE "/home/oerikss/src/miking/stdlib/mexpr/pprint.mc" 316:4-316:13 cannot be ordered by specificity; they match exactly the same values.
```

dlunde (approved):
Looks good to me!

PR #198, opened 2020-10-07
Fix error messages when running dune build.

br4sco:
This PR fixes some errors,  related to unused variables/modules and unmatched record labels, that occurs when running `dune build` manually in the project root. For some reason (maybe this is intentional) these errors are not reported when running `make`.

### DIFF
--- a/make
+++ b/make
@@ -17,7 +17,7 @@ cd stdlib; export MCORE_STDLIB=`pwd`; cd ..;
 
 # General function for building the project
 build() {
-    mkdir -p build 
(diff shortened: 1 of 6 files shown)


dlunde:
The culprit seems to be the `-p boot` flag used in `make`, which makes it a release build.

br4sco:
> The culprit seems to be the `-p boot` flag used in `make`, which makes it a release build.

Aha! nice. I guess we do not want that flag until we have a release ready.

PR #202, opened 2020-10-18
Add lexicographical comparison of strings to stdlib

johnwikman:
Adds the functions `ltString` and `gtString` to the standard library `string.mc`.

### DIFF
--- a/stdlib/string.mc
+++ b/stdlib/string.mc
@@ -14,6 +14,12 @@ recursive
            else false
 end
 
+utest eqString "" "" with true
+utest eqString "" "a" with false
+utest eqString "a" "" with false
+utest eqString "a" "a" with true
+utest eqString "a" "aa" with false
+
 let str2upper = lam s. map char2upper s
 let str2lower = lam s. map char2lower s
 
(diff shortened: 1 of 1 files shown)


br4sco:
Looks good! Speaking of ordering I think it is time that we decide upon a common signature for compare functions (this was mentioned by @elegios a while back ago). In `seq.mc`, `sort` assumes a compare function with the signature `a -> a -> Int` like in `ocaml` but there are of-course other options.

johnwikman:
I would personally like to have both. It is nice to have the `a -> a -> Int` for efficiency (and in some cases compactness), but also the `a -> a -> Bool` for when I do not want to check for `-1`, `0`, and `1` on the return value. I wouldn't mind having these as a complement to the `lt`-family:
```
let cmpi: Int -> Int -> Int
let cmpf: Float -> Float -> Int
let cmpc: Char -> Char -> Int
let cmpString: String -> String -> Int
```
Though I do not think that this should be the canonical method of comparisons in MCore, since it relies on a total order. If I had to choose one over the other, I would choose the `a -> a -> Bool` variant.

PR #215, opened 2020-11-05
Type updates and cleanup

dlunde:
- Add type annotations to lets and recursive lets in the interpreter and boot.
- Renamed `PatSeqEdg` to `PatSeqEdge` in boot (to match the interpreter).
- Replaced `TyDyn` with `TyUnknown` in the interpreter and boot.
- Replaced the `tpe` abbreviation for types with `ty` in the interpreter.
- Renamed `PVar` to `PNamed` in the interpreter to match boot.
- Other minor cleanup.

To discuss:
- I removed the type annotation on closures in boot (it was already removed in the interpreter), since this is not relevant at runtime?

### DIFF
--- a/stdlib/mexpr/ast.mc
+++ b/stdlib/mexpr/ast.mc
@@ -41,7 +41,7 @@ end
 lang FunAst = VarAst + AppAst
   syn Expr =
   | TmLam {ident : Name,
-           tpe   : Option Type,
+           ty    : Type,
            body  : Expr,
            fi    : Info}
 
@@ -75,6 +75,7 @@ end
 lang LetAst = VarAst
   syn Expr =
   | TmLet {ident  : Name,
+           ty     : Type,
            body   : Expr,
            inexpr : Expr,
 	   fi     : Info}
@@ -92,6 +93,7 @@ end
 lang RecLetsAst = VarAst
   syn Expr =
   | TmRecLets {bindings : [{ident : Name,
+                            ty    : Type,
                             body  : Expr}],
                inexpr   : Expr}
 
@@ -124,7 +126,7 @@ end
 lang DataAst
   syn Expr =
   | TmConDef {ident  : Name,
-              tpe    : Option Type,
+              ty     : Type,
               inexpr : Expr}
   | TmConApp {ident : Name,
               body : Expr}
@@ -286,15 +288,15 @@ type PatName
 con PName     : Name -> PatName
 con PWildcard : ()   -> PatName
 
-lang VarPat
+lang NamedPat
   syn Pat =
-  | PVar {ident : PatName}
+  | PNamed {ident : PatName}
 
   sem smap_Pat_Pat (f : Pat -> a) =
-  | PVar p -> PVar p
+  | PNamed p -> PNamed p
 
   sem sfold_Pat_Pat (f : a -> b -> a) (acc : a) =
-  | PVar _ -> acc
+  | PNamed _ -> acc
 end
 
 lang SeqTotPat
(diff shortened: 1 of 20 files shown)


stdlib/mexpr/ast.mc
```diff
@@ -503,10 +504,10 @@ lang MExprAst =
   CmpIntAst + CmpFloatAst + CharAst + SymbAst + CmpSymbAst + SeqOpAst
 
   -- Patterns
-  + VarPat + SeqTotPat + SeqEdgePat + RecordPat + DataPat + IntPat + CharPat +
+  + NamedPat + SeqTotPat + SeqEdgePat + RecordPat + DataPat + IntPat + CharPat +
   BoolPat + AndPat + OrPat + NotPat
 
   -- Types
-  + FunTypeAst + DynTypeAst + UnitTypeAst + CharTypeAst + StringTypeAst +
+  + FunTypeAst + UnknownTypeAst + UnitTypeAst + CharTypeAst + StringTypeAst +
```
elegios, line 511: I think it might be nice to have one fragment that does not include `Unknown`. Then we can parse to the version that contains `Unknown`, then typecheck/infer to the version that does not contain it, whereby we can write analyses that do not have to consider `Unknown`.

Maybe `MExprTyped` and/or `MExprUntyped`?

stdlib/mexpr/pprint.mc
```diff
@@ -208,7 +208,7 @@ lang AppPrettyPrint = PrettyPrint + AppAst
     else error "Impossible"
 end
 
-lang FunPrettyPrint = PrettyPrint + FunAst
+lang FunPrettyPrint = PrettyPrint + FunAst + UnknownTypeAst
```
elegios, line 211: I suppose splitting out `Unknown` has implications here. You could make one fragment that does printing in the presence of unknown and one in the absence, the former would have a more specific pattern, so they would merge correctly.

Another option is of course also to not special case unknown, print it like any other type, though that depends on it being possible to explicitly write the type `Unknown` in the file, otherwise the pretty printed version wouldn't parse the same.

stdlib/mexpr/pprint.mc
```diff
@@ -302,7 +305,7 @@ lang LetPrettyPrint = PrettyPrint + LetAst
     else never
 end
 
-lang RecLetsPrettyPrint = PrettyPrint + RecLetsAst
+lang RecLetsPrettyPrint = PrettyPrint + RecLetsAst + UnknownTypeAst
```
elegios, line 308: Same here as for functions.

stdlib/mexpr/pprint.mc
```diff
@@ -345,7 +351,7 @@ lang ConstPrettyPrint = PrettyPrint + ConstAst
   | TmConst t -> (env,getConstStringCode indent t.val)
 end
 
-lang DataPrettyPrint = PrettyPrint + DataAst
+lang DataPrettyPrint = PrettyPrint + DataAst + UnknownTypeAst
```
elegios, line 354: and here

stdlib/mexpr/pprint.mc
dlunde: I think a version of your second suggestion makes the most sense: print the type `Unknown` as `""` (empty string) in `getTypeStringCode` (or, I guess the more appropriate thing is to return an `Option String` from `getTypeStringCode`), and then detect this in `pprintCode` (to remove the `:` as well)

stdlib/mexpr/ast.mc
dlunde: Maybe, but isn't it easier to just (implicitly) rely on the typechecker to produce an AST without `Unknown`? I considered doing a similar thing for ANF (create an ANF AST), but ultimately decided against this for simplicity.

elegios: Whenever we consume a `Type` later on exhaustiveness checking will tell us to consider `Unknown` (once we have an exhaustiveness checker). In my eyes a big reason for creating languages as these compositions is the ability to have slightly different ASTs for different stages, so that the typesystem can know about the absence of the things we have removed in earlier stages, or things we've added in later stages.

It is possible that the way we want to do that is by having some form of subtraction on the composition, as opposed to not adding it, but I'm not entirely sure what that would imply yet.

dlunde: Yes, it's always nice to have static guarantees, as long as the work required for defining the different ASTs is not too cumbersome. I think composition with subtraction could be very useful for this. This is also a bit related to what I wrote in the `mcore` channel on Slack a couple of days ago (overriding included data constructors). Sounds like a prime topic for the meeting!

PR #216, opened 2020-11-09
Implemented compilation function.

br4sco:
This PR contains a prototype implementation of a compile to OCaml function. The function `compile` takes an OCaml program as a string and returns a function with type `String -> [String] -> {stdout: String, stderr: String, returncode: Int}` which when called executes the program. The first argument is given to the program via `stdin` and the second argument is arguments passed to the program.

Behind the scenes `compile` creates a dune project in the platform specified tempdirectory via the python intrinsic. Calls to the returned run function then subsequently calls `dune exec ...` which will build and/or execute the project as necessary. To call functions from the `intrinsics.ml` it is necessary to run `dune install` in the `miking` root folder to make the `boot` library globally available on the system.

Going forward we would like to:
- Separate `intrinsics.ml` into its own library and package.
- Possibly supply the OCaml program as a OCaml AST rather than a string.
- Create the dune project using `dune init ...`.

### DIFF
--- a/stdlib/char.mc
+++ b/stdlib/char.mc
@@ -72,6 +72,9 @@ let isUpperAlpha = lam c.
 
 let isAlpha = lam c. or (isLowerAlpha c) (isUpperAlpha c)
 
+let isLowerAlphaOrUnderscore = lam c.
+  or (isLowerAlpha c) (eqChar c '_')
+
 utest isAlpha 'a' with true
 utest isAlpha 'm' with true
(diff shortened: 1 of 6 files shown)


br4sco:
In addition we should remove platform specific system calls, e.g. `echo` and path concatenations.

br4sco:
> In addition we should remove platform specific system calls, e.g. `echo` and path concatenations.

Fixed.

PR #219, opened 2020-11-16
Compilation of sequence intrinsics to OCaml.

br4sco:
In this PR we:
- Re-factored identifier pretty printing into a language fragment `IdentifierPrettyPrint`.
- Synchronized sequence intrinsics between `boot` and the interpreter. 
- Implemented compilation of sequence intrinsics. `splitAt` is not tested since we cannot yet compile tuple projection.
- Moved responsibility of pretty printing names (as defined in `name.mc`) to `IdentifierPrettyPrint`.
- Moved ocaml name-escaping from `generate.mc` to `pprint.mc` (via `IdentifierPrettyPrint`). We don't escape valid fully qualified identifiers.

### DIFF
--- a/stdlib/mexpr/eval.mc
+++ b/stdlib/mexpr/eval.mc
@@ -836,8 +836,12 @@ with int_ 3 in
 
 -- Builtin sequence functions
 -- get [1,2,3] 1 -> 2
-let getAst = nth_ (seq_ [int_ 1, int_ 2, int_ 3]) (int_ 1) in
+let getAst = get_ (seq_ [int_ 1, int_ 2, int_ 3]) (int_ 1) in
 utest eval getAst with int_ 2 in
+
+-- set [1,2] 0 3 -> [3,2]
+let setAst = set_ (seq_ [int_ 1, int_ 2]) (int_ 0) (int_ 3) in
+utest eval setAst with seq_ [int_ 3, int_ 2] in
 
 -- cons 1 [2, 3] -> [1,2,3]
 let consAst = cons_ (int_ 1) (seq_ [int_ 2, int_ 3]) in
@@ -854,14 +858,21 @@ let concatAst = concat_
 utest eval concatAst
 with seq_ [int_ 1, int_ 2, int_ 3, int_ 4, int_ 5, int_ 6] in
 
--- length [1, 2, 3] = 3
+-- length [1, 2, 3] -> 3
 let lengthAst = length_ (seq_ [int_ 1, int_ 2, int_ 3]) in
 utest eval lengthAst with int_ 3 in
 
--- reverse [1, 2, 3] = [3, 2, 1]
+-- reverse [1, 2, 3] -> [3, 2, 1]
 let reverseAst = reverse_ (seq_ [int_ 1, int_ 2, int_ 3]) in
 utest eval reverseAst with seq_ [int_ 3, int_ 2, int_ 1] in
 
+-- splitAt [1,4,2,3] 2 -> ([1,4],[2,3])
+let splitAtAst = splitat_ (seq_ [int_ 1, int_ 4, int_ 2, int_ 3]) int_ 2 in
+utest eval splitAtAst with (seq_ [int_ 1, int_ 4], seq_ [int_ 2, int_ 3]) in
+
+-- makeSeq 3 42 -> [42, 42, 42]
+let makeSeqAst = makeseq_ (int_ 3) (int_ 42) in
+utest eval makeSeqAst with (seq_ [int_ 42, int_ 42, int_ 42]) in
 
 -- Unit tests for CmpFloatEval
 utest eval (eqf_ (float_ 1.0) (float_ 1.0)) with true_ in
--- a/stdlib/mexpr/eval.mc
+++ b/stdlib/mexpr/eval.mc
@@ -483,7 +483,7 @@ lang SeqOpEval = SeqOpAst + IntAst + BoolAst + ConstEval
       TmSeq {tms = CSplitAt2 s.tms}
     else error "Not splitAt of a constant sequence"
   | CMakeSeq _ ->
-    match arg with TmConst {val = CInt n} then
+    match arg with TmConst {val = CInt {val = n}} then
       TmConst {val = CMakeSeq2 n}
     else error "n in makeSeq is not a number"
   | CMakeSeq2 n ->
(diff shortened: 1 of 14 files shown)


stdlib/mexpr/eval.mc
```diff
@@ -838,30 +861,22 @@ let concatAst = concat_
...
-utest eval nullAst with true_ in
-
--- reverse [1, 2, 3] = [3, 2, 1]
+-- reverse [1, 2, 3] -> [3, 2, 1]
 let reverseAst = reverse_ (seq_ [int_ 1, int_ 2, int_ 3]) in
 utest eval reverseAst with seq_ [int_ 3, int_ 2, int_ 1] in
 
+-- splitAt [1,4,2,3] 2 -> ([1,4],[2,3])
+let splitAtAst = splitat_ (seq_ [int_ 1, int_ 4, int_ 2, int_ 3]) (int_ 2) in
+-- utest eval splitAtAst
```
elegios: Why is this test commented out? I initially thought it was because of the "`splitAt` is not tested since we cannot yet compile tuple projection" comment, but this is the interpreter, no?

stdlib/mexpr/pprint.mc
```diff
@@ -107,15 +107,15 @@ let _parserStr = lam str. lam prefix. lam cond.
 -- overriding them.
 
 -- Constructor string parser translation
-let pprintConString = lam str.
+let _pprintConString = lam str.
```
elegios: Is there a reason not to inline these in the `IdentifierPrettyPrint` fragment?

stdlib/ocaml/generate.mc
```diff
@@ -51,7 +52,44 @@ with ("abcABC____'", gensym ()).0
...
+    nameSym (join ["Boot.Intrinsics.Mseq.", op])
+  in
+  foldl (lam a. lam op. hashmapInsert hashmapStrTraits op (mkOp op) a)
+        hashmapEmpty _seqOps
+
+let _seqOp = lam op.
+  nvar_ (hashmapLookupOrElse
+    hashmapStrTraits
+    (lam _. error (strJoin " " ["Operation", op, "not found"]))
+    op _seqOpHashMap)
```
elegios: The absence of a linebreak here feels weird to me

stdlib/mexpr/pprint.mc
```diff
@@ -148,7 +148,24 @@ let _record2tuple = lam tm.
...
-lang PrettyPrint
+lang IdentifierPrettyPrint
+  sem pprintConString =
+  sem pprintVarString =
+  sem pprintLabelString =
+end
+
+lang MExprIdentifierPrettyPrint = IdentifierPrettyPrint
+  sem pprintConString =
+  | s -> _pprintConString s
```
elegios: I think there are now two things changing how identifiers are displayed; the functions in this fragment, and the renaming to avoid conflicts. Can the latter be changed, and how does it interact with these functions?

(I'm guessing this might be more of a question for the meeting, when @dlunde is present)

stdlib/mexpr/eval.mc
br4sco: Fixed.

stdlib/mexpr/pprint.mc
br4sco: Only laziness, fixed.

stdlib/ocaml/generate.mc
br4sco: Agree, I think it looks better now.

stdlib/mexpr/pprint.mc
br4sco: We discussed this so we leave it as is for now.

elegios (approved):

br4sco:
We have updated this PR since our last discussion. In particular:
- Moved responsibility of pretty printing names (as defined in name.mc) to IdentifierPrettyPrint.
- Moved ocaml name-escaping from generate.mc to pprint.mc (via IdentifierPrettyPrint). We don't escape valid fully qualified identifiers.

lingmar:
[This diff](https://github.com/miking-lang/miking/pull/219/files/8c35babbf6c303139ed6d0e32f5160f4bfbaf7c8..8823516d1906bedcc03e6194ee55234c06499580) shows the latest changes (that is, changes since the last time we discussed this PR at a Miking meeting)

dlunde:
I think this looks like a nice solution!

david-broman:
When I run `make test-all` I get the following error.

david-broman:
When I run `make test-all` I get the following error:

```
/Users/davidbroman/code/main-miking/stdlib/ocaml/pprint.mc: .................................................... OK
/Users/davidbroman/code/main-miking/stdlib/ocaml/generate.mc: ........................'dune build' failed on program:

print_string (string_of_int(Boot.Intrinsics.Mseq.length
  (Boot.Intrinsics.Mseq.make
     2
     0)))

exit code: 1

standard error:
Info: Creating file dune-project with this contents:
| (lang dune 2.6)
File "program.ml", line 4, characters 5-6:
4 |      0)))
         ^
Error: This expression has type int but an expression was expected of type
         int -> 'a
make: *** [test-all] Error 1
```

br4sco:
> When I run `make test-all` I get the following error:
> 
> ```
> /Users/davidbroman/code/main-miking/stdlib/ocaml/pprint.mc: .................................................... OK
> /Users/davidbroman/code/main-miking/stdlib/ocaml/generate.mc: ........................'dune build' failed on program:
> 
> print_string (string_of_int(Boot.Intrinsics.Mseq.length
>   (Boot.Intrinsics.Mseq.make
>      2
>      0)))
> 
> exit code: 1
> 
> standard error:
> Info: Creating file dune-project with this contents:
> | (lang dune 2.6)
> File "program.ml", line 4, characters 5-6:
> 4 |      0)))
>          ^
> Error: This expression has type int but an expression was expected of type
>          int -> 'a
> make: *** [test-all] Error 1
> ```

Since this PR includes updates to `boot` you have to run `dune install` to install the latest version of boot on your system before running all tests. In this particular case we have changed the signature of `Boot.Intrinsics.Mseq.make`.

david-broman:
Thanks, that worked.

The test time increased quite a bit with this PR, so we need to think about this in future iterations. 

Also, it would be good if we could automate the `dune` installation part. That is, can `make` do `dune install` if needed?

Anyway, I am merging this PR now.

br4sco:
> Thanks, that worked.
> 
> The test time increased quite a bit with this PR, so we need to think about this in future iterations.
> 
> Also, it would be good if we could automate the `dune` installation part. That is, can `make` do `dune install` if needed?
> 
> Anyway, I am merging this PR now.

Yes the test time is a problem. There is a PR which adds `dune install` to `make install`. I don't know if it is a good idea to add it to `make test-all` but it could be done.

PR #220, opened 2020-11-18
Pretty-printing and symbolize updates

dlunde:
### Changes
- Add a function `seq2assoc` for constructing assocMaps from tuple lists.
- Add `escapeChar` and `escapeString` to `char.mc` and `string.mc`, respectively.
- MExpr pretty-printing
   * Added `pprintAddStr` in `pprint.mc` for making sure certain names can only print as their base string (useful for, e.g., external library names)
- C pretty-printing
   * Escape (some) string characters in C pretty-printing
   * Allow for supplying a list of initial names using `pprintAddStr`. Needed for the variables `main`, `argc`, `argv`, `malloc`, `free`, etc. to always print correctly
- Symbolize changes:
   * Change environment to use separate assocMaps for vars and constructors (cleaner, similar to how it's done in `eq.mc`). Also rename symbolize environment type to `SymEnv` (was just `Env` before).
   * **Allow for leaving existing named symbols intact when symbolizing. They are all simply ignored, and assumed to be distinct from all variables without symbols.** _<-- We should probably discuss this more before merging_
- Rename equality environment to `EqEnv` (was just `Env` before)
- Rename all `CAst` data constructors to begin with `C` (avoiding conflicts).
- Add/change a few things to/in `ast-builder.mc` (ascription using `let`, in particular).

This might have conflicts with #219, so I marked it as draft for now just to be safe.


stdlib/assoc.mc
```diff
@@ -27,6 +27,13 @@ let assocInsert : AssocTraits k -> k -> v -> AssocMap k v -> AssocMap k v =
                     (lam i. set m i (k,v))
                     (index (lam t. traits.eq k t.0) m)
 
+-- 'assocConstruct traits ls' constructs a new association map from a sequence
```
br4sco: Maybe `assocOfSeq` or something similar is more descriptive?

stdlib/c/ast.mc
```diff
@@ -46,108 +46,108 @@ lang CAst
...
-  | ONeg {}
-  | ONot {}
+  syn CExpr =
+  | CEVar        { id: Name }
+  | CEApp        { fun: Name, args: [CExpr] }
+  | CEInt        { i: Int }
+  | CEFloat      { f: Float }
+  | CEChar       { c: Char }
+  | CEString     { s: String }
+  | CEBinOp      { op: CUnOp, lhs: CExpr, rhs: CExpr }
```
br4sco: Shouldn't `CUnOp` and `CBinOp` be swapped here?

stdlib/mexpr/compile-c.mc
```diff
@@ -0,0 +1,499 @@
...
+  | TyCon _ | TyApp _ | TyVar _ -> error "Type not currently supported"
+
+  -------------
+  -- HELPERS --
+  -------------
+
+  -- Translate sequence of lambdas to C function. Takes an explicit type as
+  -- parameter, because the lambdas do not explicitly give the return type,
+  -- which is required in C.
+  sem compileFun (id: Name) (ty: Type) =
```
br4sco: I don't know if we agreed on a convention, but maybe consider using `_identifier` naming if this is a internal helper function.

br4sco (commented):
I think it looks good. Spotted some minor errors (if I'm not mistaken). We could discuss the placement of `compile-c.mc`. Right now its in the `mexpr` folder. If that is the way to go we should probably move `compile.mc` from `ocaml` to `compile-ocaml.mc` in the `mexpr` folder as well. Additionally, are there any tests for for this compilation? Moreover, someone with better knowledge of `symbolize.mc` and `pprint.mc` might want to have a look at this PR.

dlunde:
> I think it looks good. Spotted some minor errors (if I'm not mistaken). We could discuss the placement of `compile-c.mc`. Right now its in the `mexpr` folder. If that is the way to go we should probably move `compile.mc` from `ocaml` to `compile-ocaml.mc` in the `mexpr` folder as well. Additionally, are there any tests for for this compilation? Moreover, someone with better knowledge of `symbolize.mc` and `pprint.mc` might want to have a look at this PR.

Thanks for the review Oscar. Yes, I also think we should discuss the placement of compiler code. I guess this boils down to placing it in the directory for the source or target language.

Right now, there are no proper tests for the compiler, since it is still at a very early prototype stage (and I have been working for a deadline). It might actually be better that I keep the compiler in my own fork for now, and then push it when it is more polished (so then this PR would only consist of the list of minor changes). It would be great to discuss the MExpr -> C compiler at the meeting today, however, so I'll leave it in here for now.

The changes to `pprint.mc` are very minor (just added a function, essentially), but it would be nice if someone could look at the symbolize changes. I can present the changes at the meeting as well (they are quite straightforward).

stdlib/mexpr/compile-c.mc
dlunde: Yes, that sounds good. Did we agree on this for semantic functions as well, or is the convention only for regular functions?

stdlib/c/ast.mc
dlunde: That sounds reasonable :)

stdlib/assoc.mc
dlunde: That was my original naming, but I avoided it since we have a convention that says we should not write conversion functions in that form (it should be `type2type`). With the `assoc` namespacing, it would then be `assocseq2assoc`, which looks horrible. I'm starting to agree that we should use the other form instead (I believe it was `typeToType` which I guess would translate to `assocSeqToAssoc` in this case).

stdlib/mexpr/compile-c.mc
dlunde: Actually, come to think of it, pretty much everything in the compiler language fragment could then be defined as helper functions (except `compile`). Where do we draw the line?

PR #228, opened 2020-12-07
Compile match to OCaml, except records and ADTs

elegios:
This PR implements a first version of pattern compilation to OCaml, and fixes some minor things I found elsewhere in the codebase while implementing. The approach is relatively simple, with the intent to support everything correctly, shortcircuit properly, and never repeat computations in a single pattern. It would not be surprising in the least if we will need to generate better code here down the road. For example, the approach here computes intermediate sequences using `split_at`, which might be unnecessarily costly in cases when the intermediate results aren't really needed (i.e., when the middle name is a wildcard).

Briefly, a match is translated to a match on an `option`, and a pattern is translated to an expression that evaluates to an `option`. For example:

```
match sequence with [a] ++ b ++ [3] then
  f a b
else
  true
```
becomes
```ocaml
(* Exact (long) generated version: *)
match
  let target =
    sequence
  in
  if
    (<)
      (Boot.Intrinsics.Mseq.length
         target)
      2
  then
    Option.None
  else
    let
      (prefix, splitTemp)
    =
      Boot.Intrinsics.Mseq.split_at
        target
        1
    in
      let
        (middle, postfix)
      =
        Boot.Intrinsics.Mseq.split_at
          splitTemp
          ((-)
             (Boot.Intrinsics.Mseq.length
                splitTemp)
             1)
      in
        let seqElem =
          Boot.Intrinsics.Mseq.get
            prefix
            0
        in
        let seqElem1 =
          Boot.Intrinsics.Mseq.get
            postfix
            0
        in
        if
          (=)
            seqElem1
            3
        then
          Option.Some (seqElem, middle)
        else
          Option.None
with
| Option.Some (a, b) ->
    (f
       a
       b)
| Option.None ->
    true

(* Or a bit more concicely: *)
open Boot.Intrinsics.Mseq in
match
  let target = sequence in
  if length target < 2
  then None
  else
    let (prefix, splitTemp) = split_at target 1 in
    let (middle, postfix) = split_at splitTemp (length splitTemp - 1) in
    let seqElem = get prefix 0 in
    let seqElem1 = get postfix 0 in
    if seqElem1 = 3 then
      Some (seqElem, middle)
    else None
with
| Some (a, b) -> f a b
| None -> true
```

The strategy for generating this stuff centers primarily around `generatePat`, which has type `Name -> Pat -> (Map Name Name, Expr -> Expr)`.

- The first argument is the name of the variable holding the value to be matched against.
- Second argument is the pattern we're compiling.
- The first return value is a mapping of names that are bound. The mapping is from name of pattern (i.e., the name that appears in the pattern itself) to name of value (i.e., the value the current pattern computed). In the example `a` and `b` are names bound by the pattern and `seqElem`, and `middle` are names of computed values.
- The second return value is the expression checking the current pattern, but in something like continuation passing style. The function takes the expression to continue with if the current pattern matches. We want this function to be linear, since dropping the argument means we don't check remaining patterns, and duplicating the argument means duplicating the code checking the rest of the pattern.

------------------------

Here's a (possibly complete) list of the other things that changed along the way, in no particular order:
- `compile.mc` now turns off two warnings for unused variables since this approach generates unused variables on wildcard patterns (and also if the AST being compiled contains unused variables, naturally). This is done because `dune` makes errors into warnings by default, thus this makes compilation fail.
- `symbolizePat` for record patterns used `env` in a location where it should have used `patEnv`. I also added a test that catches this case.
- Added an `assoc2seq` (which presently is just the identity function) to extract key-value pairs (+ tests).
- Added an `assocMapWithKey` function (+ tests).
- `pprint.mc` now also has a `printPatParen` helper, similar to `printParen` for `Expr`.
- The OCaml fragment now supports `match`, tuples & applied constructors (note that these are unrelated to those in mexpr, since the semantics are different), and patterns.
  - There is now an `ocaml/symbolize.mc` file to support these. However, since there is no construct that introduces constructors in the AST we merely error if we encounter a constructor.
- Changed `compile.mc` to not use `capture_output`, and use `cleanup` instead of `rmtree`, to be compatible with my (older) version of Python 3.
- There's now an `unzip` function in `seq.mc`.

### DIFF
--- a/stdlib/assoc.mc
+++ b/stdlib/assoc.mc
@@ -112,6 +112,11 @@ let assocMap : AssocTraits k -> (v1 -> v2) -> AssocMap k v1 -> AssocMap k v2 =
   lam _. lam f. lam m.
     map (lam t. (t.0, f t.1)) m
 
+-- 'assocMapWithKey f m' maps over the values of 'm' using function 'f', where 'f' additionally has access to the key of the value being operated upon.
+let assocMapWithKey : AssocTraits k -> (k -> v1 -> v2) -> AssocMap k v1 -> AssocMap k v2 =
+  lam _. lam f. lam m.
+    map (lam t. (t.0, f t.0 t.1)) m
+
 -- 'assocFold traits f acc m' folds over 'm' using function 'f' and accumulator
 -- 'acc'.
 -- IMPORTANT: The folding order is unspecified.
@@ -168,6 +173,7 @@ let remove = assocRemove traits in
 let keys = assocKeys traits in
 let values = assocValues traits in
 let map = assocMap traits in
+let mapWithKey = assocMapWithKey traits in
 let fold = assocFold traits in
 let foldOption = assocFoldlM traits in
 let mapAccum = assocMapAccum traits in
@@ -206,6 +212,11 @@ let m = seq2assoc seq in
 
 utest sort (lam l. lam r. subi l.0 r.0) (assoc2seq m)
 with [(1, '1'), (2, '2'), (3, '3')] in
+
+let withKeyMap = mapWithKey (lam k. lam v. (k, v)) m in
+utest lookup 1 withKeyMap with (1, '1') in
+utest lookup 2 withKeyMap with (2, '2') in
+utest lookup 3 withKeyMap with (3, '3') in
 
 let nextChar = lam c. int2char (addi 1 (char2int c)) in
 
--- a/stdlib/assoc.mc
+++ b/stdlib/assoc.mc
@@ -103,10 +103,6 @@ let assocValues : AssocTraits k -> AssocMap k v -> [v] =
   lam _. lam m.
     map (lam t. t.1) m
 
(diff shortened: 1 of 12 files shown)


stdlib/assoc.mc
```diff
@@ -103,11 +103,20 @@ let assocValues : AssocTraits k -> AssocMap k v -> [v] =
   lam _. lam m.
     map (lam t. t.1) m
 
+-- 'assoc2seq traits m' returns a sequence of all pairs stored in 'm'
+let assoc2seq : AssocTraits k -> AssocMap k v -> [(k, v)] =
+    lam _. lam m. m
+
```
dlunde: I can't see any difference compared to the existing `assoc2seq`? Additionally, the old version has not been removed?

dlunde (approved):
Looks good to me. I only looked at it very briefly, however.

stdlib/assoc.mc
elegios: I didn't actually see the old one, oops

Issue #240, opened 2021-02-05
Error `Not_found` in REPL


lingmar:
When using the MCore REPL, the evaluation of the input sometimes results in a `Not_found` error. E.g.:
```
$ mi repl
>> include "seq.mc"
>> map
Not_found
>> include "string.mc"
>> float2string 0.0
"0.0"
>> float2string 1.0
Not_found
```
I have found that the error message arises in the `eval` function in `mexpr.ml` when looking up the binding of a variable in the environment, when the symbol of the variable is in fact not bound in `env`:
```
| TmVar (_, us, s) ->
    eval env (List.assoc s env) -- BOOM: OCaml says Not_found
```
My guess is that there is a bug in how the environments are created/maintained in the REPL implementation.

lingmar:
Fixed by #241

PR #241, opened 2021-02-05
Fix types not being recorded as toplevel definitions in REPL

aathn:
This fixes the Not_found bug in the REPL (issue #240), introduced in commit 227f2c9.

I haven't followed the recent developments regarding types, so I don't know if there are any deeper problems regarding this, but this seems to make things work as far as I can see.

### DIFF
--- a/src/boot/lib/mexpr.ml
+++ b/src/boot/lib/mexpr.ml
@@ -1503,5 +1503,19 @@ let rec eval_toplevel (env : (Symb.t * tm) list) = function
       eval_toplevel (Lazy.force env') t2
   | TmCondef (_, _, _, _, t) ->
       eval_toplevel env t
-  | t ->
+  | ( TmVar _
+    | TmLam _
+    | TmClos _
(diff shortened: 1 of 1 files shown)


src/boot/lib/mexpr.ml
```diff
@@ -1484,6 +1484,8 @@ let rec eval (env : (Symb.t * tm) list) (t : tm) =
 let rec eval_toplevel (env : (Symb.t * tm) list) = function
   | TmLet (_, _, s, _ty, t1, t2) ->
       eval_toplevel ((s, eval env t1) :: env) t2
+  | TmType (_, _, _, _, t1) ->
```
dlunde, line 1487: Ah, I missed adding this when I revised the boot types. A suggestion: replace the catch all case for this match with an explicit list of remaining terms. Then, we'll get compiler errors when adding new terms such as `TmType`.

aathn: Fixed!

PR #243, opened 2021-02-11
Boot parser

david-broman:
New intrinsic functions and AST generator for using the parser from boot, to enable parsing in the MCore compiler. Includes parsing of MExpr, excluding type and info fields.


stdlib/mexpr/pprint.mc
```diff
@@ -657,7 +662,7 @@ lang SeqEdgePatPrettyPrint = SeqEdgePat + PatNamePrettyPrint
   sem getPatStringCode (indent : Int) (env : PprintEnv) =
   | PSeqEdge {prefix = pre, middle = patname, postfix = post} ->
     match _pprint_patseq getPatStringCode indent env pre with (env, pre) then
-    match _pprint_patname env patname with (env, pname) then
+    match getPatStringCode indent env patname with (env, pname) then
```
elegios: Also related to the comment on `PatName` vs `Pat`, the middle should just be a `PatName`, not a full `Pat`, so the previous implementation was correct. I guess the `pprint` "tests" don't contain an edge pattern, so this doesn't explode.

stdlib/mexpr/boot-parser.mc
```diff
@@ -0,0 +1,258 @@
...
+let idPatOr = 409
+let idPatNot = 410
+-/
+
+  -- Functions for transferring types and info are not yet implemented.  
+  -- These functions are place holders.
+  sem gtype (t:Unknown) = | n -> TyUnknown()
+  sem ginfo = | t -> NoInfo()
+
+  sem strToPatName =
```
elegios, line 168: This does not produce a `PatName`, but rather a `Pat`.

stdlib/mexpr/boot-parser.mc
```diff
@@ -0,0 +1,258 @@
...
+         matchPat t2 (bootParserGetId t2)
+
+  sem matchPat (t:Unknown) =
+  | 400 /-PatNamed-/ -> strToPatName (gstr t 0) 
+  | 401 /-PatSeqTot-/ ->
+    PSeqTot {pats = makeSeq (lam n. gpat t n) (glistlen t 0) }
+  | 402 /-PatSeqEdge-/ ->
+    let len = glistlen t 0 in
+    PSeqEdge {prefix = makeSeq (lam n. gpat t n) len,
+              middle = strToPatName (gstr t 0),
```
elegios, line 139: Related to the other comment on `strToPatName`, this should be a `PatName`, not a `Pat`.

elegios (changes requested):

stdlib/mexpr/pprint.mc
david-broman: The problem has been solved.

PR #244, opened 2021-02-11
ANF utests and some general cleanup

dlunde:


stdlib/mexpr/anf.mc
```diff
@@ -312,33 +425,31 @@ utest _anf inv1 with inv1 using eqExpr in
 let debug = false in
 
 let debugPrint = lam t.
-  let s = symbolize t in
-  let n = normalizeTerm s in
-  if debug then
+    let s = symbolize t in
+    let n = normalizeTerm s in
```
elegios, line 429: Is there value in having this run even when `debug` is not set, as it was before? `pprint` runs a bunch of code without checking the results just as a sanity check to see that it doesn't produce any run-time errors, but maybe there is enough actual test coverage here?

PR #248, opened 2021-02-17
Sequence operator

david-broman:
This PR adds a sequence operator `;` as syntactic sugar to MExpr. The PR includes updates to the boot parser, and pretty printing in the MLang compiler. Specifically, the PR includes the following:

- The expression `e1 ; e2` is syntactic sugar for `let #var"" = e1 in e2`. Note that we do not add a special wildcard for `let`, to avoid unnecessary complications when transferring `let` expressions.
- An expression `let #var"" = e1 in e2` is pretty printed as `e1 ; e2`.
- We add syntactic sugar for lambdas where the lambda variable is not used. That is,
`lam. e1` is syntactic sugar for `lam #var"". e1`.
- `_` is not possible to use in MExpr code. Instead, `_` is a keyword that is only used in patterns. Hence, `let _ = ...` is not a legal form and it is not possible to use `_` as a variable directly.

### DIFF
--- a/test/mlang/mlang.mc
+++ b/test/mlang/mlang.mc
@@ -87,7 +87,7 @@ lang FooCombined = FooA + FooTrans
 
 mexpr
 
-use ArithBool2 in
+(use ArithBool2 in
   utest eval (Add (Num 1, Num 2)) with 3 in
   utest eval (If (IsZero (Num 0)
                  ,Num 1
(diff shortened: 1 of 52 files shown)


test/mlang/mlang.mc
```diff
@@ -87,8 +87,7 @@ lang FooCombined = FooA + FooTrans
 
 mexpr
 
-let _ =
-  use ArithBool2 in
+use ArithBool2 in
```
elegios: I think this changes how far the `use` is valid, right? I think there is a point in keeping the same scoping here, to minimize risk of names in one language making a test for a later language work, just because the former happens to still be in scope, since these are tests that look at precisely that. I'm guessing parentheses around the `use` and its scope would be the easiest way.

The same should also apply to the other tests in `mlang`.

david-broman: Thanks. Fixed.

PR #250, opened 2021-02-18
Tensor implementation

br4sco:
This PR contains a tensor implementation in `boot`. Numeric tensors are implemented using Bigarrays and general tensors are implemented using Arrays. These are laid out in row-major format and no operation except `copy_exn` does any copying. The row-major layout is also preserved under all operations.

I've also added a small utility library for tensors in `stdlib/tensors.mc`, which also contains a more compact documentation of their semantics compared to the tests file. A library that uses tensors to implement references is defined in `stdlib/ref.mc`.

Things to do:
- Implement tensors in the interpreter.
- Implement `eqTensor`


README.md
```diff
@@ -502,6 +502,89 @@ utest match (1,[["a","b"],["c"]],76) with (1,b++[["c"]],76) then b else []
...
+We construct tensors using `tensorCreate shape f`, where `shape` is a
+sequence denoting the shape of the tensor and `f` is a function taking a
+index as an argument and returning the element at that index.
+
+We can construct a zero-order tensor with value 1 as
+```
+let t0 = tensorCreate [] (lam _. 'a') in
+utest tensorRank t0 with 0 in
+utest tensorShape t0 with [] in
+```
```
larshum: In the README it says you are creating a (zero-order) tensor with value `1`, but in the code example you create one with value `'a'`.

src/boot/lib/mexpr.ml
```diff
@@ -119,6 +119,16 @@ let builtin =
...
+  ; ("tensorGetExn", f (CtensorGetExn None))
+  ; ("tensorSetExn", f (CtensorSetExn (None, None)))
+  ; ("tensorRank", f CtensorRank)
+  ; ("tensorShape", f CtensorShape)
+  ; ("tensorReshapeExn", f (CtensorReshapeExn None))
+  ; ("tensorCopyExn", f (CtensorCopyExn None))
+  ; ("tensorSliceExn", f (CtensorSliceExn None))
+  ; ("tensorSubExn", f (CtensorSubExn (None, None)))
+  ; ("tensorIteri", f (CtensorIteri None))
   ; ("mapBindings", f CmapBindings) (* MCore intrinsics: Boot parser *)
```
larshum: I believe `"mapBindings"` should be placed with the other map intrinsics, i.e. before the tensor intrinsics.

stdlib/ref.mc
```diff
@@ -0,0 +1,8 @@
+-- Construct a reference
+let ref = lam x. tensorCreate [] (lam _. x)
+
+-- De-reference
+let deref = lam t. tensorGetExn t []
+
+-- Mutate reference
+let modref = lam t. lam v. tensorSetExn t [] v
```
larshum, line 8: If these are added, shouldn't we remove the existing intrinsics? Otherwise the existing intrinsics are used unless this file is included, as can be observed in the REPL (because the intrinsic references are printed differently from a rank 0 tensor):
```
>> ref 5
(ref)
>> include "ref.mc"
>> ref 5
tensor
```

src/boot/lib/tensor.ml
```diff
@@ -0,0 +1,199 @@
...
+  let create shape f =
+    let rank = Array.length shape in
+    let size = prod shape in
+    let data = Array.init size (fun i -> f (inverse_row_major_ofs shape i)) in
+    let left_ofs = 0 in
+    {data; rank; shape; left_ofs; size}
+
+  let is_valid_index shape is =
+    Array.length shape = Array.length is
+    && Array.for_all2 (fun n i -> i >= 0 && i < n) shape is
```
larshum: The function `Array.for_all2` requires OCaml version 4.11.0, which is newer than the most recent one provided by some package managers (see https://ocaml.org/docs/install.html). I thought it could be good to point this out since I could not find a version requirement of OCaml in the README.

larshum (commented):
I added a few comments, but I think it looks good overall.

README.md
br4sco: You're right!

src/boot/lib/mexpr.ml
br4sco: Good spotting!

stdlib/ref.mc
br4sco: I didn't want to remove stuff from the language before this PR was accepted. But the idea is to do what you are saying if no one objects.

src/boot/lib/tensor.ml
br4sco: Good to know, then I'll change this implementation.

larshum: Since there are no guidelines, it's hard to know where to draw the line. Also, even though 4.11.0 isn't available from all package managers, it could be installed via `opam` as described [here](https://stackoverflow.com/a/40901155). I think a minimum supported OCaml version could be a good discussion point in a future meeting.

src/boot/lib/mexpr.ml
```diff
@@ -989,6 +1058,208 @@ let delta eval env fi c v =
       TmSeq (fi, Mseq.Helpers.of_list binds)
   | CmapBindings, _ ->
       fail_constapp fi
+  (* MCore intrinsics: Tensors *)
+  | CTensor _, _ ->
+      fail_constapp fi
+  | CtensorCreate None, TmSeq (_, seq) ->
+      let shape = tm_seq2int_array fi seq in
+      TmConst (fi, CtensorCreate (Some shape))
+  | CtensorCreate (Some shape), (TmClos (clfi, _, _, _, _) as cl) ->
```
lingmar: Matching on a closure here means that e.g. `tensorCreate [] (addi 1)` is not valid, but maybe that's OK.

README.md
```diff
@@ -502,6 +502,89 @@ utest match (1,[["a","b"],["c"]],76) with (1,b++[["c"]],76) then b else []
 with [["a","b"]] in
 ```
 
+### Tensors
+Tensors are mutable data structures and can be of up to rank 16. The index
+of an element is represented as a sequence of integers.
+
+We construct tensors using `tensorCreate shape f`, where `shape` is a
+sequence denoting the shape of the tensor and `f` is a function taking a
+index as an argument and returning the element at that index.
```
lingmar: a index -> an index

src/boot/lib/mexpr.ml
br4sco: That is not intended, thx! Even though `addi 1` has the wrong type.

README.md
br4sco: Thank you!

src/boot/lib/mexpr.ml
lingmar: Right, I realised I misunderstood the type signature of `tensorCreate`. I believe a correct example is `tensorCreate [] length`

PR #251, opened 2021-02-22
Optimize interpreter

larshum:
This PR adds a few optimizations to the interpreter:
* Terms in the environment are now only evaluated if they are a `TmFix` (as discovered by @elegios).
* Functions `head` and `last` now use `get` instead of `splitAt` (better complexity).
* Adds a `sub` intrinsic, replacing the `slice` function of the standard library. Functions `tail` and `init` now use `sub` instead of `splitAt` (it seemed unnecessary to me to throw away half the result of `splitAt`, but in practice the difference is negligible).

The first two points had a significant impact on the performance of the test suite (mainly the first one). The time it takes to run the tests using `make test` (on my machine) decreased from 11s to 6s.

### DIFF
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -93,6 +93,7 @@ and const =
   | Csnoc of tm Mseq.t option
   | CsplitAt of tm Mseq.t option
   | Creverse
+  | Csub of tm Mseq.t option * int option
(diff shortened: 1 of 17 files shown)


br4sco:
Looks good!

On my machine it was 20s vs 10s.

PR #252, opened 2021-02-23
Multicore primitives

lingmar:
This PR adds multicore primitives to Miking in the form of intrinsic functions whose implementations are piggybacking on the [OCaml Multicore project](https://github.com/ocaml-multicore/ocaml-multicore). In particular, it adds primitives for atomic operations as well as for creation and synchronisation of threads that run in parallel.

Some documentation of the underlying implementation in OCaml can be found [here](https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/stdlib/atomic.mli) (for atomic references) and [here](https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/stdlib/domain.mli) (for thread management), but I’ll also provide some explanations below ~(some version of it should probably be put as documentation somewhere).~ **See the README**

## Atomic references 
Atomic references are similar to ordinary references, except that operations are atomic (no other thread can interfere with the operations). Right now the following operations are implemented (let `ARef a` denote the type of an atomic reference):
- `atomicMake : a -> ARef a` creates an atomic reference to a value of type `a`.
- `atomicGet : ARef a -> a` read the value of an atomic reference
- ~`atomicSet : ARef a -> a -> Unit` set the value of an atomic reference~ part of auxiliary library in stdlib.
- `atomicCAS : ARef a -> a -> a -> Bool` compare-and-set: `atomicCAS r seen v` sets the value of the atomic reference to `v` only if its value is physically equal to `seen`, return `true` if the update was done, else `false`.
- `atomicExchange : ARef a -> a -> a` set a new value and return the old one.
- **`atomicFetchAndAdd : ARef Int -> Int -> Int` add a value to the reference and return the old value.**

## Thread creation

- `threadSpawn : (Unit -> a) -> Proc a` creates a new thread. I.e.  when calling `threadSpawn f` , the function `f` will be run concurrently, and if possible, in parallel with the current thread.
- `threadJoin : Proc a -> a` awaits the result of a thread and should be called by the parent thread once for each thread that was spawned.
- Threads have unique identifiers that can be converted to integers in a uniqueness-preserving way. Functions for reading thread identifiers are:
	- `threadGetID : Proc a -> Pid` get the ID af a thread.
	- `threadSelf : Unit -> Pid` get the ID of the current thread.
	- `threadID2int : Pid -> Int` convert thread ID to integer.
- **`threadCPURelax : Unit -> Unit` improves performance in busy waiting.**

Note that the threads created in the underlying OCaml implementation are quite expensive to spawn and join, so one should be careful not to spawn too many threads. If many tasks are to be executed, it is better to use a task pool with a limited number of threads rather than spawning and joining a thread for each task.

## Thread synchronisation
The synchronisation primitives are `threadCriticalSection`, `threadWait`, and `threadNotify`. Semantically, they implement a monitor for each thread, where each monitor consists of one mutex and one condition variable. However, the underlying OCaml implementation is different and probably more efficient than that of a traditional monitor.  E.g. from [the documentation](https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/stdlib/domain.mli) “ `critical_section f` is cheaper than acquiring a mutex, and performs no more atomic operations than `f` does”. It can anyhow be useful as a mental model for the semantics.

- `threadCriticalSection : (Unit -> a) -> a` enter a critical section for the current thread. A critical sections is in one of two states: *notified*, or *not notified*. Initially it is not notified. It becomes notified when some other thread calls `notify p` (see below). Once it has been notified, it stays notified. A mental model could be that `threadCriticalSection f` is equivalent to:
``` 
acquire mutex for current thread
  run f	
release mutex for current thread
```
- `threadWait : Unit -> Unit` Must be called from within a critical section. Blocks until current critical section becomes notified.  Note that two calls to `threadWait` within the same critical section is not meaningful, as the second one will immediately return (because a `wait` does not “consume” a notification). A mental model could be that `threadWait ()` is equivalent to:
```
release mutex of the current thread
  wait until current critical section is notified
acquire mutex of the current thread 
```
- `threadNotify : Pid -> Unit` Notify any ongoing critical section running  in the given thread. Will block until any in-progress critical section finishes. If `p` is not in a critical section, `threadNotify p` returns immediately. A mental model could be that `threadNotify p` is equivalent to:
```
acquire mutex of thread p
  notify any ongoing critical section in p
release mutex of thread p
wait until any ongoing critical section in p has finished
```

## Choice of primitives
I chose the current set of primitives based on what primitives are used in the [`Domainslib` library](https://github.com/ocaml-multicore/domainslib) to implement task pools and communication channels, thinking that we probably want to implement something similar as a bit more high-level parallelisation abstraction in Miking. Additionally, hopefully they are generic enough so that we can implement the same primitives in another backend language than OCaml. However, the set of primitives may be subject to changes in the future depending on what we actually need.

## Questions 
* ~Do we want the multicore primitives to be an integral part of Making, or should they be optional (similar to how e.g. Python intrinsics are implemented)?~ They are now optional, see the README.
* ~Should the atomic reference be its own AST node `TmAtomicRef`, or a constant `CAtomicRef`?~ It is now part of the external multicore AST.

## Current TODOs
* ~Tests and documentation~
* I have a build problem, when running `opam pin ocamlformat 0.15.0` as required, there is a compilation error of the `ocaml-migrate-parsetree` package (which, I presume, is a prerequisite for `ocamlformat`).  It would be good to know if anyone else gets the same error. I include the output below.
<details><summary>Click to reveal</summary>
<p>

```
[ERROR] The compilation of ocaml-migrate-parsetree failed at "/Users/lingmar/.opam/opam-init/hooks/sandbox.sh build dune build -p ocaml-migrate-parsetree
        -j 7".

#=== ERROR while compiling ocaml-migrate-parsetree.1.8.0 ======================#
# context     2.0.5 | macos/x86_64 | ocaml-variants.4.10.0+multicore | https://opam.ocaml.org#03087e69
# path        ~/.opam/4.10.0+multicore/.opam-switch/build/ocaml-migrate-parsetree.1.8.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p ocaml-migrate-parsetree -j 7
# exit-code   1
# env-file    ~/.opam/log/ocaml-migrate-parsetree-28648-6d07ae.env
# output-file ~/.opam/log/ocaml-migrate-parsetree-28648-6d07ae.out
### output ###
#       ocamlc src/.migrate_parsetree.objs/byte/migrate_parsetree__Ast_410.{cmi,cmo,cmt} (exit 2)
# (cd _build/default && /Users/lingmar/.opam/4.10.0+multicore/bin/ocamlc.opt -w -40 -g -bin-annot -I src/.migrate_parsetree.objs/byte -I /Users/lingmar/.opam/4.10.0+multicore/lib/ocaml/compiler-libs -I /Users/lingmar/.opam/4.10.0+multicore/lib/ppx_derivers -I /Users/lingmar/.opam/4.10.0+multicore/lib/result -no-alias-deps -open Migrate_parsetree__ -o src/.migrate_parsetree.objs/byte/migrate_par[...]
# File "src/ast_410.ml", lines 248-303, characters 2-44:
# Error: This variant or record definition does not match that of type
#          Parsetree.pattern_desc
#        Constructors number 17 have different names, Ppat_effect and Ppat_extension.



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+- The following actions failed
| - build ocaml-migrate-parsetree 1.8.0
+- 
```

</p>
</details>


README.md
```diff
@@ -895,6 +895,188 @@ As part of the experimental setup of Miking, we currently support a way
...
+```
+utest atomicFetchAndAdd a 1 with 2 in
+-- Current value is now 3
+utest atomicFetchAndAdd a (subi 0 45) with 3 in
+-- Current value is now -42
+```
+
+##### Multi-Threaded Execution
+
+The following example program spawns 10 number of threads that compete for printing
```
dlunde: -> "... spawns 10 threads ..."

dlunde (approved):
Looks good!

README.md
lingmar: Thanks, fixed!

PR #259, opened 2021-03-02
String identfiers

david-broman:
This PR adds a small library `stringid.mc` for faster equality comparison of strings. The basic idea is to keep a global and unique mapping between integer identifiers and strings. Instead of comparing if strings are equal or not, the integer identifiers are instead compared. Then later, if the actual string is needed, the string can be retrieved via the mapping. See file `stringid.mc` for details.


dlunde (approved):
Looks good!

PR #263, opened 2021-03-02
Evaluate result of variable lookups iff they are a fixpoint application

lingmar:
This fixes a bug introduced in #251 in `eval` in boot where we previously evaluated fixpoint terms instead of *applications* of fixpoint terms, and makes the same optimization in the interpreter in `eval.mc`. Thanks to @elegios for finding the error.

I noticed no difference in the runtime of `make all` before and after the changes.

### DIFF
--- a/stdlib/mexpr/eval.mc
+++ b/stdlib/mexpr/eval.mc
@@ -59,11 +59,19 @@ let _seqOfCharToString = use MExprAst in
 -- TERMS --
 -----------
 
-lang VarEval = VarAst + IdentifierPrettyPrint
+-- Fixpoint operator is only needed for eval. Hence, it is not in ast.mc
+lang FixAst = LamAst
+  syn Expr =
+  | TmFix ()
+end
+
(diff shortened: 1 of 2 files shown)


stdlib/mexpr/eval.mc
```diff
@@ -59,11 +59,19 @@ let _seqOfCharToString = use MExprAst in
...
+  syn Expr =
+  | TmFix ()
+end
+
+lang VarEval = VarAst + IdentifierPrettyPrint + FixAst + AppAst
   sem eval (ctx : {env : Env}) =
   | TmVar {ident = ident} ->
     match _evalLookup ident ctx.env with Some t then
-      eval ctx t
+      match t with TmApp {lhs = TmFix _} then
```
elegios, line 72: There is a possible concern here that we're tying the two fragments together. I would argue that they were tied together before, just implicitly so, because fix was the only thing that needed the re-evaluation.

elegios (approved):

PR #267, opened 2021-03-03
Ocaml pprint optimization

elegios:
This PR changes how we handle identifiers that are supposed to refer to externally defined values in the compilation to OCaml. The idea is to have two versions of every AST node that refers to names (as opposed to introduces them):

- One that uses `Name`s and refers to a name defined in the current program.
- One that uses `String`s and refers to values defined externally, typically in the OCaml standard library.

In particular, we get these variants:
- `TmVar` (from mexpr originally) and `OTmVarExt`
- `OTmConApp` and `OTmConAppExt`
- `OPatCon` and `OPatConExt`

The `Ext` versions use strings that are included in the final program without processing. They also do not interact at all with `pprint` finding unique names, so they must be made unique in some other way, typically by just making them fully qualified names, e.g. `OTmVarExt {ident = "Boot.Intrinsics.Mseq.length"}`, since normal names cannot contain dots.

This means that `pprintVarName`, which was relatively expensive, is called less (since the `Ext` variants don't use it) and does less work (since it doesn't have to check if the name is a valid fully qualified name).

Basic benchmarking (no averaging over multiple runs, etc., so not super precise) suggests the following improvements:
- `pprintVarName` from 893 calls, 2613.768 ms, to 548 calls, 724.159 ms. This means that time taken went from 2.923 ms/call to 1.321 ms/call.
- `sameSemantics` from 39165 ms to 35535 ms, i.e., 3630 ms improvement

Note that we save more time in total (for `sameSemantics`) than we gained from the runtime improvement of `pprintVarName`. I believe this is because many names don't need to interact with the unique name printing mechanism at all, of which `pprintVarName` is only one part.

### DIFF
--- a/stdlib/ocaml/ast.mc
+++ b/stdlib/ocaml/ast.mc
@@ -30,7 +30,11 @@ end
 -- This fragment contains variants of other ocaml constructs where the
 -- names should appear exactly as specified, intended to be used to
 -- refer to externally defined names, e.g., in the ocaml standard
--- library
+-- library. Note that these names will not affect other names in any
+-- way, meaning that these names should be chosen so that they
+-- *cannot* overlap with other names. An easy way to do that is to
+-- always use fully qualified names, since normal names cannot contain
+-- dots.
 lang OCamlExternal
   syn Expr =
   | OTmVarExt { ident : Name }
--- a/stdlib/ocaml/ast.mc
+++ b/stdlib/ocaml/ast.mc
@@ -37,7 +37,7 @@ end
 -- dots.
 lang OCamlExternal
   syn Expr =
-  | OTmVarExt { ident : Name }
+  | OTmVarExt { ident : String }
   | OTmConAppExt { ident : String, args : [Expr] }
 
   syn Pat =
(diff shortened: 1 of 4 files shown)


stdlib/ocaml/ast.mc
```diff
@@ -27,11 +27,29 @@ lang OCamlData
...
+-- names should appear exactly as specified, intended to be used to
+-- refer to externally defined names, e.g., in the ocaml standard
+-- library. Note that these names will not affect other names in any
+-- way, meaning that these names should be chosen so that they
+-- *cannot* overlap with other names. An easy way to do that is to
+-- always use fully qualified names, since normal names cannot contain
+-- dots.
+lang OCamlExternal
+  syn Expr =
+  | OTmVarExt { ident : Name }
```
lingmar: Shouldn't the type be `String` here?

lingmar (approved):
Looks good!

stdlib/ocaml/ast.mc
elegios: True, fixed

PR #268, opened 2021-03-04
Preparation for type lifting

dlunde:
- OCaml map now contains the compare function as an MExpr term, and not as an OCaml function (makes it possible to extract in MExpr code)
- Added map intrinsics
  * `mapIsEmpty`
  * `mapGetCmpFun`
  * `mapFindOrElse`
  * `mapFoldWithKey`
  * `mapEq`
- Implemented map functions
  * `mapLength`
  * `mapKeys`
  * `mapValues`
  * `mapMapAccum`
  * `mapFoldlM`
- Renamed map intrinsic `mapFind` -> `mapFindWithErr`
- Added skeleton for `assoc-seq.mc` (Library for ordered association lists)
- Refactored `TmRecord`, `PatRecord`, and `TyRecord` to use (intrinsic) maps with `SID` as type (rather than `String`).
- Updated `TmRecordUpdate` to use `SID` as key instead of `String`.

### DIFF
--- a/stdlib/map.mc
+++ b/stdlib/map.mc
@@ -7,7 +7,7 @@ include "option.mc"
 include "seq.mc"
 
 let mapLength : Map k v -> Int =
-  lam m. mapFoldWithKey (lam acc. lam. lam. addi 1 acc) 0 m
+  mapFoldWithKey (lam acc. lam. lam. addi 1 acc) 0
 
 -- Aliases
 let mapLookupOrElse : (Unit -> v) -> k -> Map k v -> v =
--- a/stdlib/map.mc
+++ b/stdlib/map.mc
@@ -6,10 +6,18 @@
 include "option.mc"
 include "seq.mc"
 
+let mapLength : Map k v -> Int =
+  lam m. mapFoldWithKey (lam acc. lam. lam. addi 1 acc) 0 m
+
+let mapLookupOrElse : (Unit -> v) -> k -> Map k v -> v =
+  mapFindOrElse
+
 let mapLookup : k -> Map k v -> Option v =
   lam k. lam m.
-    match mapMem k m with false then None ()
-    else Some (mapFind k m)
+    let res = mapLookupOrElse (lam. None ()) k m in
+    match res with None () then None ()
+    else match res with x then Some x
+    else never
 
 let mapInsertWith : (v -> v -> v) -> k -> v -> Map k v -> Map k v =
   lam f. lam k. lam v. lam m.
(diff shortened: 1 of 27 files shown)


stdlib/mexpr/ast-smap-sfold-tests.mc
```diff
@@ -190,7 +193,7 @@ in
 let countNodes = bottomUp (sfold_Expr_Expr addi 1) in
 
 utest bottomUp identity tmVarX with tmVarX in
-utest bottomUp cInt2cChar tmRecordI with tmRecordC in
+utest bottomUp cInt2cChar tmRecordI with tmRecordC using eqExpr in
```
larshum, line 196: How come this test case needed to use `eqExpr`?

larshum (commented):
Looks good!

stdlib/mexpr/ast-smap-sfold-tests.mc
dlunde: Not sure, I guess it has to do something with the new map data type used to hold record fields. The default utest (without `using`) failed at least.

stdlib/map.mc
```diff
@@ -6,10 +6,18 @@
...
+  lam m. mapFoldWithKey (lam acc. lam. lam. addi 1 acc) 0 m
+
+let mapLookupOrElse : (Unit -> v) -> k -> Map k v -> v =
+  mapFindOrElse
+
 let mapLookup : k -> Map k v -> Option v =
   lam k. lam m.
-    match mapMem k m with false then None ()
-    else Some (mapFind k m)
+    let res = mapLookupOrElse (lam. None ()) k m in
```
elegios: This is technically not type-correct, and can't be written type-correct atm. It's possible that this would actually lead to an issue once we're compiling (e.g. we read a `0`, which is representationally equal with `None`), so I think it's probably worth fixing. We'd need a lookup function that applies a function to the contained value before returning it I think.

stdlib/map.mc
```diff
@@ -23,16 +31,39 @@ let mapUnion : Map k v -> Map k v -> Map k v = lam l. lam r.
...
+  mapFoldWithKey (lam vs. lam. lam v. snoc vs v) [] m
+
+let mapMapAccum : (acc -> k -> v1 -> (acc, v2)) -> acc -> Map k v1 -> (acc, Map k v2) =
+  lam f. lam acc. lam m.
+    mapFoldWithKey
+      (lam tacc. lam k. lam v1.
+         match f tacc.0 k v1 with (acc, v2) then (acc, mapInsert k v2 tacc.1) else never)
+      (acc, mapEmpty (mapGetCmpFun m)) m
+
+let mapFoldlM : (acc -> k -> v -> Option acc)
```
elegios: This is `foldlM` in the `Option` monad, it might be worth mentioning it in the name. I think the other occurrence of it is over sequences, but that function is in `option.mc`.

stdlib/mexpr/boot-parser.mc
```diff
@@ -308,15 +314,15 @@ let s = "{}" in
 utest lside s with rside s in
 let s = "{a = 5}" in
 utest lside s with rside s in
-let s = "{foo = 123, bar = \"Hello\"}" in
+let s = "{bar = \"Hello\", foo = 123}" in
```
elegios, line 317: The ordering here stems from the order in which the `SID`s were allocated, which in this case depends on which tests were run at which times, right? That means there's a little bit of test interdependence here? Not really sure what to do about it, or if it's fine, at least for now.

elegios (changes requested):

stdlib/map.mc
dlunde: Yes, I see what you mean. I can make the signature `mapLookupOrElse: (v1 -> v2) -> (Unit -> v2) -> Map k v1 -> v2` instead, as you suggest. Should I also rename it? Maybe `mapLookupApplyOrElse`?

stdlib/map.mc
dlunde: `mapFoldlOption`?

stdlib/mexpr/boot-parser.mc
dlunde: Right, I think the records are printed in the order of their `SID`s. We could make it so that records are always printed by the order of the actual labels (strings) instead (with a complexity cost, of course).

stdlib/map.mc
elegios: I think renaming is probably a good idea. I don't think I've seen that version, so that's probably as reasonable a name as any. It's a bit meh to have to take the cost of a function application even for cases where the old `mapLookupOrElse` would suffice, but weighing that against more intrinsics I'd probably lean towards fewer intrinsics at the moment.

stdlib/mexpr/boot-parser.mc
elegios: I feel like the cost shouldn't be terribly high given that most records will be pretty small, but I'm also not sure it's worth changing at this point in time. Down the road we might want to do something else anyway, e.g., if we want to use parse and pprint for code formatting, since we probably don't want code formatting to change the printed order of fields.

stdlib/map.mc
dlunde: Well, it is easy to keep the old `mapLookupOrElse` as well, and all of this is temporary in any case. So I'll add both!

dlunde: Unless that makes compilation more complicated (maybe that was what you meant)?

elegios: Not more complicated, just a little bit more work. Then again, this set of intrinsics is meant to be temporary, so they'll be removed down the line anyway, so I guess it doesn't really matter to have both?

elegios (approved):

PR #272, opened 2021-03-11
Type lifting

larshum:
In this PR:
* `AssocMap` replaced by `AssocSeq` as the type environment in `eq.mc`
* Type lifting of records and variants
* Updated `TyVariant` to contain constructor names _and_ argument types (using `Map`)


stdlib/assoc-seq.mc
```diff
@@ -1,14 +1,42 @@
...
 -- insertion is O(1). This makes this data type suitable for e.g., evaluation
 -- environments and similar.
 
+include "seq.mc"
+
 type AssocSeq k v = [(k, v)]
 type AssocTraits k = {eq: k -> k -> Bool}
 
-let assocSeqInsert: k -> v -> AssocSeq k v -> AssocSeq k v =
+let assocSeqEmpty : AssocSeq k v = []
```
dlunde: I'm not sure if we need to have this layer of abstraction for everything here (as we did in assocMap). My idea was that this file should simply contain a set of convenience functions over lists of tuples (as OCaml does with the `assoc` and `mem_assoc` functions).

In summary: I propose we simply use `[]` directly without having an `assocSeqEmpty`.

stdlib/assoc-seq.mc
```diff
@@ -1,14 +1,42 @@
...
+
+let assocSeqInsert : k -> v -> AssocSeq k v -> AssocSeq k v =
   lam k. lam v. lam s.
     cons (k,v) s
 
--- lookup
+let assocSeqLength : AssocSeq k v -> Int =
+  lam s.
+    length s
```
dlunde: Same here, I think we can just use `length` directly instead of wrapping it in an `assocSeqLength`.

stdlib/assoc-seq.mc
```diff
@@ -1,14 +1,42 @@
...
+  lam f. lam s.
+    map (lam t. (t.0, f t.1)) s
+
+let assocSeqFold : (acc -> k -> v -> acc) -> acc -> AssocSeq k v -> acc =
+  lam f. lam acc. lam s.
+    foldl (lam acc. lam kv. f acc kv.0 kv.1) acc s
+
+let seq2assocSeq : [(k, v)] -> AssocSeq k v =
+  lam s.
+    assocSeqFold (lam acc. lam k. lam v. assocSeqInsert k v acc) assocSeqEmpty s
```
dlunde: Same here

stdlib/assoc-seq.mc
```diff
@@ -1,14 +1,42 @@
...
+
+let assocSeq2seq : AssocSeq k v -> [(k, v)] =
+  lam s.
+    s
```
dlunde: And here

stdlib/mexpr/type-lift.mc
```diff
@@ -5,19 +5,613 @@
...
+      (env, ty)
+    else match mapLookup fields env.records with Some name then
+      (env, ntyvar_ name)
+    else
+      let recName = nameSym "Rec" in
+      let recTyVar = ntyvar_ recName in
+      let env = {{env with records = mapInsert fields recName env.records}
+                      with typeEnv = assocSeqInsert recName ty env.typeEnv} in
+      (env, recTyVar)
+end
```
dlunde, line 427: Shouldn't there be a recursive call to `typeLiftType` here somewhere? We need to lift out nested records before lifting the outer record.

dlunde: In `RecordTypeTypeLift`, that is.

dlunde (commented):
Looks good overall! See my specific comments below. Also, it would be great if you could show some examples of how this works (would be very helpful for myself at least).

stdlib/assoc-seq.mc
larshum: Yeah, that sounds better. I pushed a commit where the functions you mentioned were removed.

stdlib/mexpr/type-lift.mc
larshum: Ooops, yes that's right. This should be fixed now.

PR #273, opened 2021-03-12
Compile map intrinsics to OCaml

lingmar:
* Breaks out the logic of the map intrinsics into separate functions (added to the `intrinsics.ml*` files)
* Compiles the map intrinsics to OCaml in `ocaml/generate.mc`
* All of the `utest`s comparing records now seem to need `using eqExpr` (I'm not sure why)
* The `mapGetCmpFun` is not compiled in this PR, because I had to do a series of hacks in order to make it work (see [this diff](https://github.com/lingmar/miking/compare/compile-map-intrinsics...lingmar:map-get-cmp-fun) to see what I did). I would like to discuss how to do it, or whether we actually need this function.

EDIT: I figured out how to compile `mapGetCmpFun` in a less hacky way, so it's added now.


larshum (commented):
I think this looks good. We will need to add type annotation for these intrinsics, but perhaps it is best if I include that in #276.

PR #277, opened 2021-03-16
Updates to pretty printers

lingmar:
* Boot and interpreter: escapes the characters `\t`, `\\`, `\'`, and `\"` (along with `\n` from before).
* Boot: escapes variables, constructors and labels with `#var`, `#con`, and `#label` when needed (as done in the interpreter).
* Boot: only print type annotations for `TmLet`, `TmLam`, `TmVar`, `TmRecLets`, and `TmConDef` when the type is not `Unknown` (as done in the interpreter).
* Boot: pretty print a let-expression with an empty variable name using sequence operator `;` (as done in the interpreter).

The motivation is that I'm using the boot pretty printer to compile from OCaml to MCore, so I need the output to be parsable. (The last point is not needed for parsability but it annoyed me.)

### DIFF
--- a/src/boot/lib/pprint.ml
+++ b/src/boot/lib/pprint.ml
@@ -532,9 +532,13 @@ and print_tm' fmt t =
         (print_ty_if_known ty) print_tm (Lam, t1)
   | TmLet (_, x, s, ty, t1, t2) ->
       let x = string_of_ustring (ustring_of_var x s) in
-      let ty = ty |> ustring_of_ty |> string_of_ustring in
-      fprintf fmt "@[<hov 0>@[<hov %d>let %s%s =@ %a in@]@ %a@]" !ref_indent x
-        (print_ty_if_known ty) print_tm (Match, t1) print_tm (Match, t2)
+      if x = "" then
+        fprintf fmt "@[<hov 0>@[<hov %d>%a;@]@ %a@]" !ref_indent print_tm
+          (Match, t1) print_tm (Match, t2)
+      else
+        let ty = ty |> ustring_of_ty |> string_of_ustring in
+        fprintf fmt "@[<hov 0>@[<hov %d>let %s%s =@ %a in@]@ %a@]" !ref_indent
+          x (print_ty_if_known ty) print_tm (Match, t1) print_tm (Match, t2)
   | TmType (_, x, s, ty, t1) ->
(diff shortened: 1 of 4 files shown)


src/boot/lib/ustring.ml
```diff
@@ -279,6 +279,10 @@ module Op = struct
 
   let uc c = int_of_char (if c = '\x0D' then '\x0A' else c)
 
+  let is_lower_alpha c = uc 'a' <= c && c <= uc 'z'
+
+  let is_upper_alpha c = uc 'A' <= c && c <= uc 'Z'
+
```
larshum: As I understand it, a `uchar` represents a Unicode character. Therefore I think these functions should be named something like `is_ascii_lower_alpha` and `is_ascii_upper_alpha` to clarify that these functions only return true for ascii-characters (there are more Unicode characters that could be classified as lowercase/uppercase).

src/boot/lib/pprint.ml
```diff
@@ -143,7 +184,7 @@ let rec ustring_of_ty = function
   | TyVariant _ ->
       failwith "Printing of non-empty variant types not yet supported"
   | TyVar (_, x, s) ->
-      ustring_of_var x s
+      ustring_of_con x s
```
larshum: How come you use `ustring_of_con` here? In `pprint.mc` the identifier of a `TyVar` is printed using `pprintEnvGetStr` which does not, as far as I can tell, prepend invalid names with `#con`. You also use `ustring_of_con` for `PatCon`, `TmConDef` and `TmConApp`, but their constructor identifiers are printed using `pprintConName` which does prepend `#con`.

This question also applies to the pretty-printing of `TmType`.

larshum (changes requested):
It looks good overall. I found two details that seemed like they might need to be changed (one of them for sure, the other one could just be a fault on my side).

src/boot/lib/ustring.ml
lingmar: Makes sense, I renamed them like you suggested.

src/boot/lib/pprint.ml
lingmar: Thanks! I thought `TyVar` and `TmType` were escaped as well, but I should have checked.

larshum (approved):

PR #287, opened 2021-03-19
Add test cases for some intrinsics in `generate.mc`

gizemcaylak:
This PR 
- Adds CDPrint (it does nothing except returning unit)
- Adds test cases for some intrinsics in `generate.mc`
- Changes return type of the map binding to Mseq in `intrinsics.ml`


larshum (approved):
I think it looks good!

PR #293, opened 2021-03-22
Adding using term for TmUtest in the standard library.

gizemcaylak:
This PR

- Adds using term for TmUtest in the standard library.
- Changes type signature of `not` in `bool.mc`
- Adds epsilon tolerated float equivalence function in `math.mc`

### DIFF
--- a/stdlib/mexpr/ast-builder.mc
+++ b/stdlib/mexpr/ast-builder.mc
@@ -424,7 +424,7 @@ let appf8_ = use MExprAst in
 
 let utestu_ = use MExprAst in
   lam t. lam e. lam n. lam u.
-  TmUtest {test = t, expected = e, next = n, tusing = u, ty = TyUnknown {}, info = NoInfo ()}
+  TmUtest {test = t, expected = e, next = n, tusing = Some u, ty = TyUnknown {}, info = NoInfo ()}
 
 let utest_ = use MExprAst in
   lam t. lam e. lam n.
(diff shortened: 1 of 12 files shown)


stdlib/mexpr/ast.mc
```diff
@@ -324,6 +324,7 @@ lang UtestAst
   | TmUtest {test : Expr,
              expected : Expr,
              next : Expr,
+             tusing : Expr,
```
lingmar: Isn't the type of `tusing` `Option Expr`?

stdlib/mexpr/ast-builder.mc
```diff
@@ -422,10 +422,13 @@ let appf8_ = use MExprAst in
   lam f. lam a1. lam a2. lam a3. lam a4. lam a5. lam a6. lam a7. lam a8.
   app_ (appf7_ f a1 a2 a3 a4 a5 a6 a7) a8
 
+let utestu_ = use MExprAst in
+  lam t. lam e. lam n. lam u.
+  TmUtest {test = t, expected = e, next = n, tusing = u, ty = TyUnknown {}, info = NoInfo ()}
```
lingmar: I think `tusing = Some u` is better than `tusing = u` (so that you don't have to send in an option type when using the `utestu_` function).

stdlib/mexpr/utesttrans.mc
```diff
@@ -221,13 +221,25 @@ let utest_info_ =
...
+          , ty = TyUnknown {}
+          , info = default_info}
+in
+
+let utestu_info_ =
+  lam t. lam e. lam n. lam u.
+  TmUtest { test = t
+          , expected = e
+          , next = n
+          , tusing = u
```
lingmar: I think it should be `tusing = Some u` here as well

PR #296, opened 2021-03-24
Don't do `Obj.repr` on record bindings

lingmar:
* Bugfix: changes so that `Obj.repr` is *not* done on bindings in records.
* Bugfix: adds parentheses around bindings in pretty-printing of records (otherwise `;` might be interpreted as the sequence operator)
* Adds a test case that would fail without the changes in this PR.
* Fixes the type signature of `phRunCommand`


larshum (approved):
Looks good to me!

PR #297, opened 2021-03-25
Improve utest runner type support

larshum:
Improves the type support of the utest runner:
* Adds print functions `float`s and `char`s, as well as sequences containing any of the supported types (including nested sequences).
* Adds default equality functions for `char`s and sequences containing any of the supported types (including nested sequences).

Edit: I also had time to implement printing and equality for record types as well, so I added it to this PR.

### DIFF
--- a/stdlib/mexpr/utesttrans.mc
+++ b/stdlib/mexpr/utesttrans.mc
@@ -189,43 +189,94 @@ let float2stringName = optionGetOrElse
 let withUtestRunner = lam term.
   bind_ utestRunner term
 
-recursive let _printFunc = use MExprAst in
-  lam ty.
-  match ty with TyInt _ then
-    nvar_ int2stringName
-  else match ty with TyFloat _ then
-    nvar_ float2stringName
-  else match ty with TyBool _ then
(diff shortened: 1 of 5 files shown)


stdlib/mexpr/utesttrans.mc
```diff
@@ -94,45 +182,66 @@ let int2stringName = optionGetOrElse
...
-  match ty with TyInt {} then
+  match ty with TyInt _ then
     nvar_ int2stringName
-  else match ty with TyBool {} then
+  else match ty with TyFloat _ then
+    nvar_ float2stringName
+  else match ty with TyBool _ then
     ulam_ "b" (if_ (var_ "b") (str_ "true") (str_ "false"))
+  else match ty with TyChar _ then
+    ulam_ "c" (seq_ [char_ '\'', var_ "c", char_ '\''])
```
lingmar: At some point we should escape special chars (but it can wait I think)

lingmar (approved):
Looks good!

stdlib/mexpr/utesttrans.mc
larshum: I made a fix so that the OCaml pretty-printer should now escape special characters like these, so I think this should not be a problem.

PR #316, opened 2021-03-31
Don't generate OCaml code for empty variant types

lingmar:
* Before this PR empty variant types did not compile:
```
type Foo
```
because it was translated into 
```
type var_Foo =
  ;;
```
This PR fixes this by not generating any code at all for empty variant types.

* I also made a change to improve the performance of `ocaml/pprint.mc`. For me, it reduced the runtime of `compile-all.sh` from `142.92s` to `105.48s`.


larshum (approved):
Looks good! 👍

PR #317, opened 2021-03-31
support utf-8 format in compiler

gizemcaylak:
This PR adds support for utf-8 encoding for chars in compiler (also utest in `generate.mc`).
The only concern is that the efficiency of printing all characters as chars [ `int2string (char2int c)` ]

### DIFF
--- a/stdlib/ocaml/generate.mc
+++ b/stdlib/ocaml/generate.mc
@@ -785,14 +785,14 @@ let ocamlEvalBool = lam ast.
 in
 
 let ocamlEvalChar = lam ast.
-  ocamlEval (wrapOCamlAstInPrint ast (printf "'%c'"))
+  ocamlEval (wrapOCamlAstInPrint ast (ulam_ "x" (app_ (printf ("'%c'")) (int2char_ (var_ "x")))))
 in
 
 utest ocamlEvalInt (int_ 1) with int_ 1 using eqExpr in
 utest ocamlEvalFloat (float_ 1.) with float_ 1. using eqExpr in
 utest ocamlEvalBool true_ with true_ using eqExpr in
 utest ocamlEvalChar (char_ '1') with char_ '1' using eqExpr in
-
+utest '島'with '島' in
 -- Compares evaluation of [mexprAst] as a mexpr and evaluation of
 -- [ocamlAst] as a OCaml expression.
 let sameSemantics = lam mexprAst. lam ocamlAst.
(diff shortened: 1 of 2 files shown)


stdlib/ocaml/generate.mc
```diff
@@ -785,14 +785,14 @@ let ocamlEvalBool = lam ast.
...
-  ocamlEval (wrapOCamlAstInPrint ast (printf "'%c'"))
+  ocamlEval (wrapOCamlAstInPrint ast (ulam_ "x" (app_ (printf ("'%c'")) (int2char_ (var_ "x")))))
 in
 
 utest ocamlEvalInt (int_ 1) with int_ 1 using eqExpr in
 utest ocamlEvalFloat (float_ 1.) with float_ 1. using eqExpr in
 utest ocamlEvalBool true_ with true_ using eqExpr in
 utest ocamlEvalChar (char_ '1') with char_ '1' using eqExpr in
-
+utest '島'with '島' in
```
larshum: I think this test should compare the evaluation of a character AST node with the character, as the current version only compares the characters in the interpreter. Try doing something like:
```suggestion
utest ocamlEvalChar (char_ '島') with char_ '島' in
```

stdlib/ocaml/generate.mc
```diff
@@ -785,14 +785,14 @@ let ocamlEvalBool = lam ast.
 in
 
 let ocamlEvalChar = lam ast.
-  ocamlEval (wrapOCamlAstInPrint ast (printf "'%c'"))
+  ocamlEval (wrapOCamlAstInPrint ast (ulam_ "x" (app_ (printf ("'%c'")) (int2char_ (var_ "x")))))
```
larshum, line 788: I'm pretty sure the `%c` printf-format only works for ASCII characters, so you'll have to do something more complicated here. I think you can convert the integer to a ustring (of course, you want it to be treated as a unicode character rather than as an actual integer) and then print the resulting string using the `%s` format.

stdlib/ocaml/pprint.mc
```diff
@@ -148,7 +148,7 @@ lang OCamlPrettyPrint =
   | CGeqf _ -> "((>=) : float -> float -> bool)"
   | CNeqf _ -> "((!=) : float -> float -> bool)"
   | CInt2float _ -> "float_of_int"
-  | CChar {val = c} -> showChar c
+  | CChar {val = c} -> int2string (char2int c)
   | CEqc _ -> "((=) : char -> char -> bool)"
   | CChar2Int _ -> "int_of_char"
   | CInt2Char _ -> "char_of_int"
```
larshum, line 154: If we compile MExpr characters into OCaml integers, then these constants should be translated into no-ops in OCaml.

larshum (changes requested):
I think this is a good start, but there are a few things left to make this work.

PR #331, opened 2021-04-09
Example programs

br4sco:
This PR adds a flag `--debug-generate` to `mi.mc` that will pretty print the generated ocaml program when running the `compile` sub command. Additionally there is a temporary flag `--exclude-intrinsics-preamble` to omit printing the intrinsics preamble.

There is also a number of small example programs where the ocaml code generation results in verbose code.


src/main/mi.mc
```diff
@@ -8,43 +8,44 @@ include "seq.mc"
...
+  "Usage: mi [compile|run] <files>",
   "",
   "Options:",
   "  --debug-parse                    Print the AST after parsing",
-  "  --exit-before                    Exit before evaluation or compilation"]
+  "  --debug-generate                 Print the AST after code generation",
+  "  --exit-before                    Exit before evaluation or compilation",
   "  --test                           Generate utest code",
-  "  --exclude-intrinsics-premable    Exclude the intinsics premable"]
+  "  --exclude-intrinsics-premable    Exclude the intinsics preamble"]
```
larshum: There is a typo in the name of the flag.
```suggestion
  "  --exclude-intrinsics-preamble    Exclude the intinsics preamble"]
```

larshum (commented):
Apart from the typo, I think it looks good.

PR #337, opened 2021-04-14
Removed prelude

david-broman:
This PR removes the prelude completely from MCore. As decided before, we should instead use explicit includes, if needed.


david-broman:
As discussed at the meeting, I will merge this PR before the code elimination.

larshum (approved):

PR #339, opened 2021-04-14
Replace python process helpers with `command` intrinsic

lingmar:
This PR adds a new intrinsic `command` for executing shell commands. The old process helpers based on Python calls used for compilation are replaced by calls to `command` and `readFile`/`writeFile`. Note that the used commands will only work on Unix systems.


larshum (approved):

br4sco:
Looks good! Feel free to rename `process-helpers.mc` I'm not to fond of that name but I don't have any good alternatives right now.

PR #341, opened 2021-04-15
Dead code elimination after parsing in Boot

david-broman:
This PR adds dead code elimination, as an analysis and transformation step after parsing. The task is performed in Boot, and affects all parts of the compiler, including analysis, transformations, and code generation.  

Currently, the time for compiling and running all tests went down from approximately 6min and 20 seconds to 2 min and 53 seconds. However, there might still be room for some more improvements.

Although the PR makes quite a big change in performance, there are a few potential improvements in the future:
1. Recursive lets cannot be eliminated in certain cases because the analysis cannot exclude side effects (the recursive variable is not available during the analysis. E.g.
```
recursive
let fax = lam f. lam e. f (fax f) e in

let fact = fax (lam recur. lam n.
  if eqi n 0
    then 1
    else muli n (recur (subi n 1)))
in
()
```

2. Elimination of unused constructors is not yet implemented.
3. If a let starts with a partial application and not a lambda, it cannot be detect right now that it is OK in some cases. For instance, just run `include "stdlib/mexpr/pprint.mc"` in an empty file.

### DIFF
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -21,6 +21,8 @@ let enable_debug_after_symbolize = ref false
 
 let enable_debug_after_dead_code_elimination = ref false
 
+let enable_debug_dead_code_info = ref false
+
 let enable_debug_after_mlang = ref false
 
 let enable_debug_symbol_print = ref false
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -19,6 +19,8 @@ let enable_debug_after_parse = ref false
 
 let enable_debug_after_symbolize = ref false
 
+let enable_debug_after_dead_code_elimination = ref false
+
 let enable_debug_after_mlang = ref false
 
 let enable_debug_symbol_print = ref false
--- a/src/boot/lib/ast.ml
+++ b/src/boot/lib/ast.ml
@@ -391,13 +391,11 @@ let sfold_tm_tm (f : 'a -> tm -> 'a) (acc : 'a) = function
   | TmLet (_, _, _, _, t1, t2) ->
       f (f acc t1) t2
   | TmRecLets (_, lst, tm) ->
-      f
-        (f acc (List.fold_left (fun acc (_, _, _, _, t) -> f acc t) acc lst))
-        tm
+      f (List.fold_left (fun acc (_, _, _, _, t) -> f acc t) acc lst) tm
   | TmConst (_, _) ->
       acc
   | TmSeq (_, tms) ->
-      Mseq.Helpers.fold_right f tms acc
+      Mseq.Helpers.fold_left f acc tms
   | TmRecord (_, r) ->
       Record.fold (fun _ t acc -> f acc t) r acc
   | TmRecordUpdate (_, r, _, t) ->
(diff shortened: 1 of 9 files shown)


src/boot/lib/deadcode.ml
```diff
@@ -0,0 +1,103 @@
...
+        work (collect_in_body s nmap free t1) t2
+    | TmRecLets (_, lst, tt) ->
+        let f (nmap, free) (_, _, s, _, t) = collect_in_body s nmap free t in
+        work (List.fold_left f (nmap, free) lst) tt
+    | t ->
+        sfold_tm_tm work (nmap, free) t
+  in
+  work (SymbMap.empty, SymbSet.empty) t
+
+(* Returns a new namp, where it is marked with true everywhere we have
```
larshum:
```suggestion
(* Returns a new nmap, where it is marked with true everywhere we have
```

src/boot/lib/ast.ml
```diff
@@ -478,6 +480,139 @@ let ty_info = function
...
+  | CreadFile
+  | CwriteFile _
+  | CfileExists
+  | CdeleteFile
+  | Cerror
+  | Cexit ->
+      true
+  (* MCore intrinsics: Symbols *)
+  | CSymb _ | Cgensym | Ceqsym _ | Csym2hash ->
+      true
```
larshum, line 565: I don't think `eqsym` and `sym2hash` has side-effects. At least not in the current implementation.

src/boot/lib/ast.ml
```diff
@@ -478,6 +480,139 @@ let ty_info = function
...
+  | CtensorGetExn _
+  | CtensorSetExn _
+  | CtensorRank
+  | CtensorShape
+  | CtensorCopyExn _
+  | CtensorReshapeExn _
+  | CtensorSliceExn _
+  | CtensorSubExn _
+  | CtensorIteri _ ->
+      true
```
larshum, line 599: I don't think all the tensor intrinsics have side-effects. For example `tensorRank` does not modify the tensor.

src/boot/lib/deadcode.ml
```diff
@@ -0,0 +1,103 @@
...
+let elimination t =
+  (* Collect all lets and store a graph in 'nmap' and free variable in 'free' *)
+  let nmap, free = collect_lets t in
+  if !enable_debug_dead_code_info then (
+    _symbmap := symbmap t ;
+    print_endline "-- Dead code info: collected lets --" ;
+    pprint_nmap !_symbmap nmap |> uprint_endline ;
+    print_endline "-- Dead code info: free variables --" ;
+    pprint_named_symbset !_symbmap free |> uprint_endline ) ;
+  (* Mark all lets that used in the graph *)
```
larshum:
```suggestion
  (* Mark all lets that are used in the graph *)
```

src/boot/lib/ast.ml
br4sco: In light of the discussion today, `tensorRank`, `tensorShape`, `tensorReshapeExn`, `tensorSliceExn`, and `tensorSubExn` are free from side effects.

src/boot/lib/deadcode.ml
david-broman: Thanks. Fixed.

src/boot/lib/ast.ml
david-broman: Might be correct now, but since the underlying handling of symbols has side effects, I would like to be conservative right now and just say that things that are guaranteed to be side effect free in the future are marked as such.

src/boot/lib/deadcode.ml
david-broman: Fixed.

src/boot/lib/ast.ml
david-broman: I am waiting with that, for the same reasons as for symbols.

PR #344, opened 2021-04-20
Made the bootstrapped compiler the default mi command

david-broman:
In this PR, I have renamed the boot interpreter to `boot` and also added compilation of the bootstrapped compiler when running `make`. The the new bootstrapped compiler is now called `mi`. The bootstrapped compiler is only recompiled if it does not exist. Both `mi` and `boot` are installed when run `make install` (in the future, we might actually not install `boot`). All tests are still running `boot` as before. I have also changed so that the `boot` command invokes tests by using the flag `--test`, as is done in the bootstrapped compiler.

### DIFF
--- a/make
+++ b/make
@@ -22,15 +22,16 @@ cd stdlib; export MCORE_STDLIB=`pwd`; cd ..;
 build() {
     mkdir -p build
     dune build
+    dune install > /dev/null 2>&1 
     cp -f _build/install/default/bin/boot.mi build/$BOOT_NAME
     if [ -e build/$MI_NAME ]
     then
         echo "Bootstrapped compiler already exists. Run 'make clean' before to recompile. "
     else
         echo "Bootstrapping the Miking compiler (1st round, might take a few minutes)"
-        time (build/$BOOT_NAME src/main/mi.mc -- compile src/main/mi.mc)
-        echo "Bootstrapping the Miking compiler (2st round, might take some more time)"
-        time ($MI_NAME compile src/main/mi.mc)
+        time build/$BOOT_NAME src/main/mi.mc -- compile src/main/mi.mc
+        echo "Bootstrapping the Miking compiler (2nd round, might take some more time)"
+        time ./$MI_NAME compile src/main/mi.mc
         mv -f $MI_NAME build/$MI_NAME
         rm -f mi
     fi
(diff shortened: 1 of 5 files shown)


make
```diff
@@ -12,66 +12,82 @@
...
-    cp -f _build/install/default/bin/boot.mi build/mi
+    cp -f _build/install/default/bin/boot.mi build/$BOOT_NAME
+    if [ -e build/$MI_NAME ]
+    then
+        echo "Bootstrapped compiler already exists. Run 'make clean' before to recompile. "
+    else
+        echo "Bootstrapping the Miking compiler (1st round, might take a few minutes)"
+        time (build/$BOOT_NAME src/main/mi.mc -- compile src/main/mi.mc)
+        mv -f $MI_NAME build/$MI_NAME
+        echo "Bootstrapping the Miking compiler (2st round, might take some more time)"
```
larshum: Is there a benefit to bootstrapping the compiler twice?

make
```diff
@@ -12,66 +12,82 @@
...
+        time (build/$MI_NAME compile src/main/mi.mc)
+        mv -f $MI_NAME build/$MI_NAME
+        rm -f mi
```
larshum, line 36: I don't think this line is needed, as the line above should have already moved the `mi` file to the `build` directory.

make
david-broman: Right now, I would like to do it for testing, both regarding time measurements and that it works.

make
david-broman: Actually, I can remove the first copy operation. It is fixed now.

make
```diff
@@ -12,66 +12,81 @@
...
     dune build
-    cp -f _build/install/default/bin/boot.mi build/mi
+    cp -f _build/install/default/bin/boot.mi build/$BOOT_NAME
+    if [ -e build/$MI_NAME ]
+    then
+        echo "Bootstrapped compiler already exists. Run 'make clean' before to recompile. "
+    else
+        echo "Bootstrapping the Miking compiler (1st round, might take a few minutes)"
+        time (build/$BOOT_NAME src/main/mi.mc -- compile src/main/mi.mc)
+        echo "Bootstrapping the Miking compiler (2st round, might take some more time)"
```
br4sco: Should be 2nd round ... right?

david-broman: Yes, thanks.

PR #352, opened 2021-04-23
Add `range` to `seq.mc`

gizemcaylak:
This PR add `range` function which generates sequence of numbers from `s` to `e` (excluding the end `e`) with step size `by` to `seq.mc`


br4sco:
Looks good! You could warp `unfoldr` if you want.

gizemcaylak:
> Looks good! You could warp `unfoldr` if you want.

That makes sense, thanks!

br4sco (approved):

PR #353, opened 2021-04-25
Microbenchmark and command line updates

david-broman:
This PR updates the command line options for `boot` and `mi` slightly, as well as adding a micro benchmark for profiling different command line options (running `boot` using eval vs. compiling and running Miking code, vs. interpreting using `mi`, vs. running native OCaml code etc.).

The command line for `mi` is as follows:

```
Usage: mi <command> file [<options>|<args>]

Commands:
  eval      Evaluates a .mc file using an internal interpreter
  compile   Compiles a .mc file into an executable with the same name
  run       Combines eval and compile, to run the program as fast as possible

If no command is given, the file will be executed using the run command
and all arguments after the file are arguments to the .mc executed file.

Options:
  --debug-parse       Print the AST after parsing
  --debug-generate    Print the AST after code generation
  --exit-before       Exit before evaluation or compilation
  --test              Generate utest code
  -- [args]           If the run or eval commands are used, then the texts
                      following -- are arguments to the executed program
```

That is, it is now possible to run `mi` in a `run` mode, where parameters to the program are given directly. Note the difference between `eval` and `run` (although they behave the same at the moment).

Boot has been updated accordingly, 

```
Usage: boot [eval|repl] <files>
```
which means that you need to type `boot eval test.mc` to actually evaluate a program (same syntax as for `mi`).

Finally, in folder `/test/microbenchmark` there is now a simple profiling tool, which is invoked using command `run` (from the folder `/test/microbenchmark`). 

If you execute just `./run` you get the following help:
```
Usage: run <benchmark-name-without-postfix-name> <iteration> [excludes]

Example: run factorial 1000

To exclude certain tests, add excluded numbers. For instance
command 'run factorial 1000 24' excludes the Miking interpreter
and OCaml byte code tests (test numbers 2 and 4).
```

Right now, there are 2 small tests `factorial` and `tree`, but it is easy to add more. 

If we run it using `factorial`, we get

```
>> ./run factorial 2000
1. Boot interpreter:     0.096276s  0.095236s
2. Miking interpreter:   2.184327s  2.209865s
3. Miking compiler:      0.208660s  0.006239s
4. OCaml byte code       0.008265s  0.007732s
5. OCaml native:         0.160605s  0.005109s
```

The first column shows the warm-up run (clear difference for the Miking compiler). We can also observe that the Miking interpreter is much slower than the Boot interpreter (for some reason).

When we want to compare the Miking compiler and OCaml native, it is convenient to exclude some tests, e.g., 1, 2 and 4. You can do that as follows (and run 1000000 iterations instead):

```
>> ./run factorial 1000000 124
3. Miking compiler:      0.536543s  0.462859s
5. OCaml native:         0.511623s  0.461593s
```

As you can see, the Miking compiler generates basically as good code as the OCaml compiler (for this simple test program). But, we need to check which compile time optimizations that are used in the different cases (the OCaml code is built using `ocamlbuild`)


src/main/mi.mc
```diff
@@ -6,37 +6,66 @@
...
+and all arguments after the file are arguments to the .mc executed file.
+
+Options:
+  --debug-parse       Print the AST after parsing
+  --debug-generate    Print the AST after code generation
+  --exit-before       Exit before evaluation or compilation
+  --test              Generate utest code
+  -- [args]           If the run or eval commands are used, then the texts
+                      following -- are arguments to the executed program
+"
```
larshum: How come there is both a `test` command and a `--test` flag now? What is the difference between them?

src/main/mi.mc
```diff
@@ -6,37 +6,66 @@
...
   else
-    [printLn (join ["Unknown command '", get argv 1, "'"]), exit 1]
+    -- No, not a well known command
+    -- Is it a .mc file instead of a command
+    if isSuffix eqChar ".mc" cmdString then
+       -- Yes, run the 'run' command with arguments and default options
+       run [cmdString] options (cons "mi" rest)
+    else
+       -- No, print error
+       [printLn (join ["Unknown command '", get argv 1, "'"]), exit 1]
```
larshum: Shouldn't this be using the sequencing "operator" instead of creating a sequence that will never be used?
```suggestion
       printLn (join ["Unknown command '", get argv 1, "'"];
       exit 1
```

src/main/mi.mc
david-broman: The idea was to use the test command where you can give arguments to the actual program (not to `mi`) directly. E.g.

```
mi test hi.mc --foo
```
means that `--foo` is a parameter to `hi.mc` and not to `mi`. 

An alternative is to shadow certain parameters from `mi`, so that the they take over. In such a case the `test` command is not needed. Let's discuss the alternatives at the Miking meeting.

src/main/mi.mc
david-broman: Yes, that is correct. It is actually old code from before we had the sequence operator. I will update this.

PR #354, opened 2021-04-26
Speed up OCaml dune compilation

larshum:
This PR makes a few changes to the `dune` file used to compile the generated OCaml code:
* One `dune` file is generated. It contains different profiles (`dev` for fast compilation, `opt` for optimized compilation) that are chosen based on flags given to the MCore compiler.
  * Both modes now pass the `-without-runtime` command to `ocamlc`, which seems to speed up compilation (even though we weren't even using the bytecode it produces). If we make the fast compile option generate bytecode later on, we could still use this option and run the produced bytecode with `ocamlrun`.
  * `dev` passes the flag `-linscan` to `ocamlopt` to make it use a simpler register allocation algorithm, which significantly speeds up compilation. It also uses `-inline 1` to decrease the amount of inlining of functions, which seemed to speed up compilation as well (default value is `1.25`).
  * `opt` passes the `-O3` flag to `ocamlopt` which admittedly does not seem to make a big difference. I also cannot find any documentation about it except when running `-help` (it does not show up in the manual), so I'm not sure it is always supported.
* Always disable warnings in `dune` compilation (warnings are treated as errors, and our generated code almost always gets some warnings, so I doubt we will ever want them).
* Adds a flag `--disable-optimizations` which selects the `dev` profile. The `opt` profile is chosen by default.


dlunde:
`-O3` is mentioned here: https://ocaml.org/manual/flambda.html

david-broman:
Is the `-flambda` option enabled?

larshum:
No, it doesn't seem to be and I don't explicitly use that option. My installation does not include `flambda` and the current instructions on installing the multicore version in the README does not include `flambda`, so I think we would need to update those in case we want that.

david-broman:
@larshum Did you not add a flag `--disable-optimizations` to `mi`? I cannot see anything in the help about this when you type `mi`.

larshum:
Oh, it was accidentally removed when I merged. I readded it now.

PR #355, opened 2021-04-27
Const transformer

david-broman:
This PR transforms variables pointing to builtins in the environment to constant expressions. It is implemented as a call to the boot parser and is only used in the MCore compiler.


david-broman:
Wrong target.

PR #361, opened 2021-04-29
Switch to make for bootstrap tests

dlunde:
My machine ran out of memory + swap space when running `compile-all.sh`, so I decided it was finally time to switch to using `make` instead. This means we now use a proper process pool, reducing both memory consumption and process switching overhead compared to the previous naive approach to parallelism.

Basically, you can now run `make -j 5 test-bootstrap` and this combines what `compile-all.sh` and `run-all.sh` did previously, but now with a proper process pool (in this case with a limit of 5 processes = my number of cores + 1 which is usually a good choice).

~~P.S. As a side benefit, you can now also use `make name/of/file.mc` to run compile and run tests for just one file. Basically, each file specified in either `compile_files` or `run_files` is now a make target.~~
Edit: I reorganized this PR a bit, so the above is not the syntax you should use. Now you can use `make -f test-run.mk FILENAME` or `make -f test-compile.mk FILENAME` instead.


larshum:
I haven't tried this out yet, but I think it sounds great. I have also had problems with memory when running `compile-all.sh` so I have made the `boot` part sequential when running locally, but this is a far better approach.

PR #364, opened 2021-05-06
Improved pattern typing

larshum:
This PR makes the type annotations of patterns use the structure of the patterns to derive type information, even when the type of the `target` in a match has an unknown type. The difference is that semantic functions that operate on tuples had to be expressed as
```
| (TyVar t1, TyVar t2) ->
  type vTy = {info : Info, ident : Name} in
  let t1 : vTy = t1 in
  let t2 : vTy = t2 in
  ...
```
but after the changes in this PR the extra annotations are not needed. This only applies to tuples as we cannot safely assume all fields of a record are listed in the pattern (but we assume that this is the case for tuples).


dlunde:
I'm probably missing something here, but why can't we handle incomplete records as well? It is only the actual record type we can't figure out: the component types (which are most often the ones we need) are often inferable.

For example, if we had `| { a = TyVar t1, c = TyVar t2 } ->` instead, we should be able to deduce the types of the bindings `a` and `c` (and `t1` and `t2`) in the branch, even if the type of the target is, say, `{ a: ..., b: ..., c: ... }` (which we of course don't know).

larshum:
Yes, that part will work. But later on the compilation will fail anyway because the exact record type has to be known when we are matching on the contents in OCaml.

As an example, the following is a rough estimation of what could happen if we generate code for the match you suggested:
```
type Rec1 = {a : Obj.t; b : Obj.t; c : Obj.t};;
type Rec2 = {a : Obj.t; c : Obj.t};;

fun __sem_target ->
  let T {a = a; c = c} = Obj.obj __sem_target in
  ...
```
If we don't know the record type, we cannot know what type we should use for `T` in the destructuring let. We could assume that `a` and `c` are the only fields meaning the type `T` should be `Rec2`. But as the position of the `c` field is different in the record declarations, this would lead to the wrong result or a segmentation fault if it the record actually has type `Rec1`.

dlunde:
OK, I understand. I'm happy it now works for tuples at least!

PR #365, opened 2021-05-06
Add an empty .mli file when compiling OCaml

larshum:
By adding an empty `.mli` file when compiling the generated OCaml code the compilation seems to be significantly faster.

My hypothesis is that this becomes faster because when the public interface is empty the compiler is allowed to eliminate unused let-expressions at the top-level (similar to `static` functions in C).


david-broman:
This is a very interesting observation!

PR #367, opened 2021-05-07
Updates for removing TmExt during dead code elimination

david-broman:
This PR removed externals (TmExt) during dead code elimination, if the externals are not used in the program. It handles side effects and lambda counts based on the type signature of the external.

Also, this PR adds a make target `boot`, which makes it possible to compile boot only, without bootstrapping the whole compiler. To do this, just type `make boot`. This is useful when working on just the boot code in OCaml.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@
 
 .PHONY :\
   all\
+  boot\
   test\
   install\
   lint\
@@ -30,6 +31,8 @@
 
 all: build/mi
 
+boot:
+	@./make boot
 build/mi:
 	@./make
 
--- a/make
+++ b/make
@@ -18,12 +18,18 @@ export MI_NAME=mi
 # Setup environment variable to find standard library
 cd stdlib; export MCORE_STDLIB=`pwd`; cd ..;
 
-# General function for building the project
-build() {
+# Compile and build the boot interpreter
+build_boot(){
     mkdir -p build
     dune build
-    dune install > /dev/null 2>&1
     cp -f _build/install/default/bin/boot.mi build/$BOOT_NAME
+}
+
+
+# General function for building the project
+build() {
+    build_boot
+    dune install > /dev/null 2>&1
     if [ -e build/$MI_NAME ]
     then
         echo "Bootstrapped compiler already exists. Run 'make clean' before to recompile. "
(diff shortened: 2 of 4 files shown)


br4sco:
Running `boot eval ./src/main/mi.mc -- eval --exit-before --debug-parse ...` on the program
```
external ext ! : ()
``` 
results in
```
{}
```
I.e. the external is removed even if its marked as having side effects. Is this correct?

PS. I made a PR with two example programs. DS.

david-broman:
> Running `boot eval ./src/main/mi.mc -- eval --exit-before --debug-parse ...` on the program
> 
> ```
> external ext ! : ()
> ```
> 
> results in
> 
> ```
> {}
> ```
> 
> I.e. the external is removed even if its marked as having side effects. Is this correct?

Yes it is correct. The external function has a side effect, but the function is never used. Here is an example with a few different cases:

```
external foo! : Float -> Float -> Float

external boo! : Int

external boo2! : Int

external bar : Int


let a = foo

let b = foo 1.0

let c = foo 1.0 2.0

let d = boo

let e = bar

```

When running `mi eval t1.mc --debug-parse --exit-before` we get

```
external foo! : (Float) -> ((Float) -> (Float))
in
external boo! : Int
in
let c =
  foo
    1.0e-0
    2.0e+0
in
let d =
  boo
in
{}
```

`foo` is preserved because of declaration `c`, but if line `c` is removed, `foo` is removed too. It counts the lambdas (a bit difficult to explain). `boo` is kept because `d` is using it (open `let` with no closing lambda). `boo2` is removed because no-one is using `boo2`. It does not matter that it has a side effect then. Finally `bar` is (in contrast to `boo`) removed even if it is used in the let of `e`, because `bar` has no side effect. Finally, if `bar` would be used in the final expression, e.g. `mexpr bar`, then even `bar` would have been preserved.

br4sco (approved):
Looks good!

david-broman:
I updated the explanation. Does it make sense?

br4sco:
Yes, maybe we can add that program in `test/examples/externals` ?

david-broman:
Sure! Please feel free to do that in your next PR about externals.

PR #372, opened 2021-05-12
Build Miking with `4.12.0+domains` switch

lingmar:
This PR:
* ~Adds installation instructions for how to build Miking with the multicore flambda switch created in [this PR](https://github.com/miking-lang/ocaml-multicore/pull/6).~
* Adds installation instructions for how to build Miking with the 4.12.0 multicore OCaml compiler.
* Miking is now always built with multicore support, so the section about parallel programming is moved up in the README, the `dune` file is updated and the `*.par-skel.ml` files are removed.

Note that the installation instructions for Miking now are a bit horrible but that they will happen under the hood in the future. 

Some pointers I found for going forward:
* [Local opam switches](https://opam.ocaml.org/blog/opam-local-switches/#Current-status). It's not clear to me if it's yet fully supported and if it's enough for what we want.
* [Opam source code](https://github.com/ocaml/opam). Perhaps we can build OPAM from source code and use non-global directories for the package installations.

For reference, to install the switch with `flambda` support, the switch installation command would be:
```
opam switch create 4.12.0+domains+flambda --packages=ocaml-variants.4.12.0+domains,ocaml-option-flambda --repositories=multicore=git+https://github.com/ocaml-multicore/multicore-opam.git,default
```
However, this causes a stack overflow error when running `mi compile --test stdlib/ocaml/generate.mc`. Bootstrapping works, though. I have created an [issue](https://github.com/ocaml-multicore/ocaml-multicore/issues/556) about it in the ocaml multicore repo.


README.md
```diff
@@ -9,14 +9,41 @@ Before you can use the Miking system, you need to install
 [OCaml](https://ocaml.org/) and the
 [OPAM](https://opam.ocaml.org/) package manager.
 
-
-After the installation, you need to install the `opam` packages `dune`, `batteries`, and `linenoise` by running the following:
+After the installation, you need to install the OCaml multicore compiler by
+following these steps:
+* Clone the `miking-lang/multicore-ocaml` repository by running the following in
```
larshum:
```suggestion
* Clone the `miking-lang/ocaml-multicore` repository by running the following in
```

README.md
```diff
@@ -9,14 +9,41 @@ Before you can use the Miking system, you need to install
...
+  your choice of directory:
+  ```
+  git clone https://github.com/miking-lang/ocaml-multicore.git
+  ```
+* Make sure you are on the `no-effect-syntax` branch in the
+  `miking-lang/multicore-ocaml` repository:
```
larshum:
```suggestion
  `miking-lang/ocaml-multicore` repository:
```

README.md
```diff
@@ -9,14 +9,41 @@ Before you can use the Miking system, you need to install
...
+  ```
+  git checkout no-effect-syntax
+  ```
```
larshum: If the commands are copy-pasted, then this command will fail because one has to change directory before running this command. Perhaps you could also add a line `cd ocaml-multicore` to this command?

larshum (commented):
The installation instructions worked on my machine, I had no problems when running all the tests.

However, I get some warnings when I run `opam upgrade` that I do not get when using other switches. I'm not sure if they are a problem or not, but I thought it would be good to mention just in case:
```
The following packages are not being upgraded because the new versions conflict with other installed packages:
  - ocaml.4.13.0
    ∗ batteries.3.3.0 is installed and requires ocaml (>= 4.00.0 & < 4.13.0)
    ∗ dune.2.8.5 is installed and requires ocaml < 4.08~~
    ∗ ocaml-migrate-parsetree.2.1.0 is installed and requires ocaml (>= 4.02.3 & < 4.13)
    ∗ ocamlformat.0.18.0 is installed and requires ocaml (>= 4.08 & < 4.13)
    ∗ odoc.1.5.2 is installed and requires ocaml (>= 4.02.0 & < 4.13)
    ∗ ppxlib.0.22.0 is installed and requires ocaml (>= 4.04.1 & < 4.13)
  - ocaml-config.2
  - ppxlib.0.22.0
```

README.md
lingmar: Thanks, prefixed a `cd` command

lingmar:
Replaced by #388

PR #373, opened 2021-05-14
Added dist file.

david-broman:
WIP for handling statistical distributions using external functions.


br4sco:
https://github.com/miking-lang/miking/pull/381 introduces support for labeled arguments so you should not have to write boilerplate code if it gets accepted.

PR #376, opened 2021-05-18
Making magic

elegios:
This PR changes our strategy for `Obj.t` in the OCaml compilation, and also changes the representation of constructors whose arguments are precisely `{}`. Briefly:
- We put `Obj.magic` in the places that need a particular type, and leave all others to whatever OCaml infers them to be. This is essentially the target of a match (e.g. `match Obj.magic x with ...`) and function application (e.g., `Obj.magic f a b c`).
  - We also put `Obj.magic` around all but one of the branches of a `match` (or `if`). This is typically not needed since most arms will infer to the same type, the bootstrapping doesn't need it for example, but some of our tests need it, and it is more correct to do it this way.
  - Arguments to constructors (except inlined records) and records are `Obj.repr`ed.
  - As a consequence, we can remove the preamble entirely.
  - I also moved the insertion of these things into `generate` itself. Performance-wise this probably doesn't matter all that much, `objWrap` did not take up very much compilation time, but `generate` already has access to type information that is convenient, especially for constructor application.
- Constructors whose arguments are explicitly `{}` are represented by nullary constructors in OCaml, meaning they do not require any allocation or garbage collection. We may need to revisit this choice when we implement extensible records, or have two versions of such constructors.
  - We also special case constructors that are applied to record literals, to not have to deconstruct them first. I *think* this wasn't done before, but I'm not entirely sure.

The sum total of these changes bring the second stage of bootstrapping from ~20s to ~10s on my machine, which seems a little bit too good to be honest, given what the changes are, but things seem to work.


larshum (approved):

david-broman:
This is great. For me, the the 1st compilation round when from 57.6s to 45.3s. The second round went from 14.1s to 6.7s.

Wow!

larshum:
I ran the `factorial` microbenchmark with this version and the `develop` version using both the previous `multicore` compiler switch and the one with `flambda` enabled (#372). I found that this version performed slightly better for both compiler versions than `develop`, but there were no huge differences.

However, I got a stack overflow error when compiling `stdlib/ocaml/generate.mc` (using `mi compile ...`). It seems to happen during the `dune` compilation. I didn't get this error when compiling `mi` in the `develop` branch.

larshum:
I got the stack overflow error on a different branch I'm working on now (which is not related to this in any way). It only seems to happen when compiling OCaml with our custom `flambda` enabled compiler, so I think this PR is fine.

elegios:
Hm, if it causes issues in combination with flambda that might still be an issue. Did it happen on just this branch with flambda, or with additional changes + flambda?

larshum:
It seems that running `mi compile stdlib/ocaml/generate.mc` results in a stack overflow error on `develop`, this branch, the branch I was working on and also the `flambda-switch` branch (#372) when I'm using the custom `flambda`-enabled compiler. So the problem is probably related to that PR - or might might just be that I have messed up my installation somehow.

elegios:
I can imagine that we generate very large expressions that flambda might not quite be written to handle. Annoying.

And just to be clear, it's `ocamlc` that is overflowing the stack, not `mi`?

larshum:
Yes. It fails when `mi` is running `dune build program.exe` and it seems to be when running `ocamlopt`.

david-broman:
Any news on this issue? Is the problem just related to flambda? @larshum is not the same problem appearing when you do the same thing without this patch, but with flambda?

larshum:
Yes, this PR is fine. The problem does not pop up when I'm using the `4.10.0+multicore+no-effect-syntax` switch. When compiling with `flambda` I'm getting the same errors on the `develop` branch, so I think the problem is related to that switch.

PR #381, opened 2021-05-25
Sundials interface

br4sco:
This PR became quite big and contains a number of things.

**Related to data conversion from and to external functions**
*Containers*
E.g. sequences to and from lists and arrays. These are now handled properly, recursively converting if necessary.

*Tuples*
I added support for conversions between tuples types. If the elements in the tuple do not need any data conversion the tuple is passed directly as the corresponding MExpr record have the same data representation as the OCaml tuple. If indeed conversion of the elements are needed, these are projected, then converted, and then finally a new tuple is constructed from the converted elements.

*Records*
I also added support for records. As of now MExpr records are unordered and we therefore have to project from the MExpr record and then construct an OCaml record from the projected elements, and vice versa. However, in the external definition, the order of the record type fields as parsed in MExpr matters when mapping the record type to an OCaml record type. As an example, the external definition
```
external myRec : {a : Int, b : [Int]}
```
where `myRec` maps to a OCaml value
```
type my rec_t : {field_a : Int; field_b : int list}
let my_rec : my_rec_t = {field_a = 1; field_b = [2]}
```
Maps the field `a` to `field_a` and the field `b` to `field_b`. In other words the mapping is based on position in the syntax. 

You could of course do this mapping based on names (i.e. we name the fields the same in OCaml and MExpr), but then I feel that there is an abstraction leakage from the target language into MExpr. Now the mapping is based on position as described above. This requires us to maintain the order of the record type fields as defined in the syntax so I encode this in a sequence which is added to `TyRecord`. This sequence is only ever used when converting data, so in all other aspects record fields are still unordered in MExpr. However, I had to update the type-lifting to keep record types with the same type on the fields, but with different order in the type environment. These were previously reduced to one record type. I made this behavior modular since the c code generation depended on the latter behavior.

*Labeled arguments*
I added support to interface with OCaml functions with labeled or optional arguments, i.e. `fun ~a:b ~c:d -> ...`.

*Bigarray.Genarray, Bigarray.Array1, and Bigarray.Array2*
I added support for converting numerical tensors to `Bigarray.Genarray`, `Bigarray.Array1`, and  `Bigarray.Array2`. These assume that you have created your tensor with `tensorCreateCArrayInt` or `tensorCreateCArrayFloat`. Later on this is something we want to ensure/check at compile time. Right now you get a runtime error if this is not the case. Similarly, `Bigarray.Array1` and  `Bigarray.Array2` requires the tensor to be of rank 1 and rank 2 respectively, something that hopefully will be encoded in the tensor type in the future.

*Tests and other comments*
Test for externals can be found in `stdlib/ext/ext-test.mc` together with `stdlib/ext/ext-test.ext-ocaml.mc`. I also did some refactoring of the data conversion since you have to be careful how you introduce `Obj.magic` which became apparent when implemented the sundials interface.

**Sundials**
I implemented an interface to sundials in `stdlib/sundials/sundials.mc`, `stdlib/sundials/sundials.ext-ocaml.mc`. Because we do not yet convert between MExpr constructors and OCaml constructors I had to include some wrapper code in `src/boot/lib/sundials_wrapper.sd.ml`. I removed the old interface.

**Other**
- I added `zip` to `seq.mc`
- I split `make test-compile` into `make test-compile`, which runs the compilation tests using only the compiled compiler, and `make test-boot-compile`, which runs the compilation tests using only the evaluated compiler.
- I added a `--debug-type-annot` to pretty print the AST after type annotation.
- I changed `make test` to only run the basic `boot` tests. 
- `tuple_`, `record_`, and `unit_` now requires a type. There are un-typed version prefixed with `u` similar to `lam_` and `ulam_`.


stdlib/mexpr/type-lift.mc
```diff
@@ -104,6 +143,7 @@ lang TypeLift = Cmp
     let emptyTypeLiftEnv : TypeLiftEnv = {
       typeEnv = [],
       records = mapEmpty (mapCmp cmpType),
+      labels = setEmpty (seqCmp subi),
```
larshum: SIDs are represented as integers right now but I don't think we should assume that.
```suggestion
      labels = setEmpty (seqCmp cmpSID),
```

stdlib/ocaml/generate.mc
```diff
@@ -402,8 +383,42 @@ lang OCamlGenerate = MExprAst + OCamlAst + OCamlMatchGenerate
...
+        --         ident = name,
+        --         tyIdent = TyUnknown { info = info },
+        --         ty = TyUnknown { info = info },
+        --         body = body,
+        --         info = info
+        --       })
+        --     body
+        --     names
+        -- in
+        -- let body = generate env body in
```
larshum: Perhaps this commented-out code should be removed? Or un-commented?

larshum (commented):
I think you accidentally added a binary file `src/main/compile` that should not be included. Apart from that and two other minor comments below I think it looks good.

stdlib/mexpr/type-lift.mc
br4sco: Thanks!

stdlib/ocaml/generate.mc
br4sco: Yes, thanks!

PR #393, opened 2021-05-31
Smap stuff

elegios:
This PR switches the `smap` and `sfold` implementations to use `smapAccumL` (and also adds the latter) for everything in `mexpr/ast.mc`. Additionally, it removes a lot of boilerplate in `type-lift` and `symbolize` by using these instead.

I believe that most of the changes themselves should be fairly uncontroversial, but there appears to be some performance implications of these changes during bootstrapping. Below are some highly unscientific performance numbers (one run per commit and switch).

I feel like these are things that we want to be able to optimize later, to have them be zero-cost relative to manually written  
boilerplate, but for now we'd take a penalty by including them, so I figured it was worth bringing up.

UPDATE: the timings seem fairly unstable when I rerun them today, so I'm not sure how much this actually shows. Yesterday it felt like it was fairly consistent that runs at the end of the commits were slower than before them, but that's not as obvious now. I haven't run these super-properly, I was still using the laptop in the meanwhile (30 min per switch, it would have been dreadfully boring otherwise).

## 4.10.0+multicore+no-effect-syntax

| Commit | Phase 1 (s) | Phase 2 (s) |
| --- | --- | --- |
| Pre | 196.95 | 18.81 |
| "smapAccumL_Pat_Pat" | 202.93 | 19.81 |
| "smapAccumL_Expr_Expr" | 212.58 | 20.83 |
| "smapAccumL_Type_Type" | 213.93 | 22.00 |
| "smapAccumL_Expr_Type" | 218.43 | 23.59 |
| "minor-fixes" | 218.18 | 24.67 |
| "smapAccumL-in-type-lift" | 212.65 | 23.56 |
| "smap-in-symbolize" | 217.44 | 22.83 |

## 4.12.0+flambda

| Commit | Phase 1 (s) | Phase 2 (s) |
| --- | --- | --- |
| Pre | 191.99 | 24.10 |
| "smapAccumL_Pat_Pat" | 193.25 | 21.58 |
| "smapAccumL_Expr_Expr" | 198.30 | 23.84 |
| "smapAccumL_Type_Type" | 210.14 | 24.71 |
| "smapAccumL_Expr_Type" | 201.35 | 24.73 |
| "minor-fixes" | 204.21 | 26.04 |
| "smapAccumL-in-type-lift" | 204.86 | 24.89 |
| "smap-in-symbolize" | 201.62 | 26.11 |

## 4.12.0+domains+flambda

| Commit | Phase 1 (s) | Phase 2 (s) |
| --- | --- | --- |
| Pre | 197.08user | 19.50user |
| "smapAccumL_Pat_Pat" |  201.43user | 19.64user |
| "smapAccumL_Expr_Expr" |  206.97user | 19.41user |
| "smapAccumL_Type_Type" |  220.55user | 22.41user |
| "smapAccumL_Expr_Type" |  212.19user | 24.07user |
| "minor-fixes" |  223.85user | 24.07user |
| "smapAccumL-in-type-lift" |  214.55user | 23.75user |
| "smap-in-symbolize" |  214.64user | 23.53user |

### DIFF
--- a/stdlib/mexpr/ast.mc
+++ b/stdlib/mexpr/ast.mc
@@ -43,6 +43,19 @@ lang Ast
     let res: (acc, Expr) = smapAccumL_Expr_Expr (lam acc. lam a. (f acc a, a)) acc p in
     res.0
 
+  sem smapAccumL_Type_Type (f : acc -> a -> (acc, b)) (acc : acc) =
+  | p -> (acc, p)
+
+  sem smap_Type_Type (f : a -> b) =
+  | p ->
+    let res: ((), Type) = smapAccumL_Type_Type (lam. lam a. ((), f a)) () p in
+    res.1
+
+  sem sfold_Type_Type (f : acc -> a -> acc) (acc : acc) =
+  | p ->
+    let res: (acc, Type) = smapAccumL_Type_Type (lam acc. lam a. (f acc a, a)) acc p in
+    res.0
+
   sem smapAccumL_Pat_Pat (f : acc -> a -> (acc, b)) (acc : acc) =
   | p -> (acc, p)
 
@@ -891,9 +904,6 @@ lang UnknownTypeAst = Ast
   sem tyWithInfo (info : Info) =
   | TyUnknown t -> TyUnknown {t with info = info}
 
-  sem smap_Type_Type (f : Type -> a) =
-  | TyUnknown t -> TyUnknown t
-
   sem infoTy =
   | TyUnknown r -> r.info
 end
@@ -904,9 +914,6 @@ lang BoolTypeAst = Ast
 
   sem tyWithInfo (info : Info) =
   | TyBool t -> TyBool {t with info = info}
-
-  sem smap_Type_Type (f : Type -> a) =
-  | TyBool t -> TyBool t
 
   sem infoTy =
   | TyBool r -> r.info
@@ -919,9 +926,6 @@ lang IntTypeAst = Ast
   sem tyWithInfo (info : Info) =
   | TyInt t -> TyInt {t with info = info}
 
-  sem smap_Type_Type (f : Type -> a) =
-  | TyInt t -> TyInt t
-
   sem infoTy =
   | TyInt r -> r.info
 end
@@ -933,9 +937,6 @@ lang FloatTypeAst = Ast
   sem tyWithInfo (info : Info) =
   | TyFloat t -> TyFloat {t with info = info}
 
-  sem smap_Type_Type (f : Type -> a) =
-  | TyFloat t -> TyFloat t
-
   sem infoTy =
   | TyFloat r -> r.info
 end
@@ -946,9 +947,6 @@ lang CharTypeAst = Ast
 
   sem tyWithInfo (info : Info) =
   | TyChar t -> TyChar {t with info = info}
-
-  sem smap_Type_Type (f : Type -> a) =
-  | TyChar t -> TyChar t
 
   sem infoTy =
   | TyChar r -> r.info
@@ -963,8 +961,13 @@ lang FunTypeAst = Ast
   sem tyWithInfo (info : Info) =
   | TyArrow t -> TyArrow {t with info = info}
 
-  sem smap_Type_Type (f : Type -> a) =
-  | TyArrow t -> TyArrow {{t with from = f t.from} with to = f t.to}
+  sem smapAccumL_Type_Type (f : acc -> a -> (acc, b)) (acc : acc) =
+  | TyArrow t ->
+    match f acc t.from with (acc, from) then
+      match f acc t.to with (acc, to) then
+        (acc, TyArrow {{t with from = from} with to = to})
+      else never
+    else never
 
   sem infoTy =
   | TyArrow r -> r.info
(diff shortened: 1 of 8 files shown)


david-broman:
@elegios can you extend the benchmarking a little bit, to see how it performs on the the multicore switch vs. the flambda switch?

elegios:
There are more numbers now.

stdlib/mexpr/ast.mc
```diff
@@ -19,6 +20,70 @@ lang Ast
...
+  | p -> (acc, p)
+
+  sem smap_Expr_Expr (f : a -> b) =
+  | p ->
+    let res: ((), Expr) = smapAccumL_Expr_Expr (lam. lam a. ((), f a)) () p in
+    res.1
+
+  sem sfold_Expr_Expr (f : acc -> a -> acc) (acc : acc) =
+  | p ->
+    let res: (acc, Expr) = smapAccumL_Expr_Expr (lam acc. lam a. (f acc a, a)) acc p in
```
dlunde, line 43: Type name `acc` shoule be `Acc` maybe?

dlunde (commented):
The reduction in code size is fantastic, but we should probably discuss the performance implications on Thursday.

stdlib/mexpr/ast.mc
dlunde: Or wait, it's a variable. Nevermind

elegios:
I redid the measurements after re-adding the explicit implementations of `smap` and `sfold`, and the numbers are even more confusing. Re-adding seems to slow things down, even though it should just restore previous behavior. My best guess at this point is that the slowdown is actually related to the amount of code we have to process, not any overhead from calling one version of `smap` or another.

With this as my hypothesis I ran `boot eval src/main/mi.mc -- compile --debug-generate --exit-before | wc -lmw` to get some numbers on the size of the code we give to OCaml. I also added the number of lines added/removed in each commit, more as a curiosity than anything else. I used the `4.12.0+flambda` switch.

| Commit | Phase 1 (s) | Phase 2 (s) | Lines | Words | Characters | LoC delta in `.mc` |
| --- | --- | --- | --- | --- | --- | --- |
| Pre | 209.73 | 22.58 | 175278 | 268129 | 8410185 |   |
| "smapAccumL_Pat_Pat" | 210.07 | 24.95 | 176010 | 270313 | 8511446 | +2 |
| "readd-smap-Pat-Pat" | 212.75 | 25.43 | 178318 | 274045 | 8688906 | +69 |
| "smapAccumL_Expr_Expr" | 224.34 | 24.62 | 180961 | 280942 | 8972364 | +15 |
| "readd-smap-Expr-Expr" | 229.82 | 26.45 | 186438 | 289928 | 9654386 | +109 |
| "smapAccumL_Type_Type" | 235.87 | 28.15 | 186766 | 290808 | 9694530 | +17 |
| "readd-smap-Type-Type" | 240.78 | 29.29 | 187754 | 292464 | 9768218 | +36 |
| "smapAccumL_Expr_Type" | 239.95 | 29.57 | 187754 | 292464 | 9768218 | +40 |
| "minor-fixes" | 239.95 | 31.24 | 187757 | 292479 | 9768497 | 0 |
| "smapAccumL-in-type-lift" | 237.93 | 29.51 | 185981 | 289052 | 9670719 | -223 |
| "smap-in-symbolize" | 238.73 | 30.02 | 187825 | 293528 | 10112385 | -155 |

There does seem to be some correlation between the amount of code we pass to OCaml and the time taken, though it is certainly not a direct fit. It does however fit somewhat well when we compare with the previous runs; all commits in this run have more code than the corresponding ones in the previous runs, and it does indeed take longer. On the other hand, "Pre" is the state before any patches, which should be the same between all of them, so there is clearly a lot of variance (I don't think I've pulled any other changes since, so they should be in the same state).

Theorizing some more, I can imagine that all this happens because of our approach to language fragments. Each language fragment gets a copy of each semantic function, with the appropriate match-arms. The `s`functions in particular are in the `Ast` fragments, which are included in a very large portion of the fragments we create. I don't know how well the dead code elimination deals with this, nor how much effort it has to expend on it, nor how expensive the mlang pattern analyses are for these intermediate languages that aren't really used.

More investigation needs to be done to confirm this, but it seems fairly reasonable as an explanation to me. Unfortunately it probably means that this PR probably shouldn't be merged for a while yet, until we've figured that out (and made it fast).

elegios:
I did some more benchmarking to see the performance effects of #399, which gave the following results (Pre is `develop` as it looks right now, `switch-to-map-in-symbolize-ml` is `develop` plus #399, the rest are from this PR plus #399):

| Commit | Phase 1 (s) | Phase 2 (s) |
| --- | --- | --- |
| Pre | 189.07user | 20.50user |
| switch-to-map-in-symbolize-ml | 174.07user | 16.67user |
| smapAccumL_Pat_Pat | 176.97user | 15.51user |
| smapAccumL_Expr_Expr | 184.74user | 15.79user |
| smapAccumL_Type_Type | 183.64user | 15.46user |
| smapAccumL_Expr_Type | 185.27user | 15.99user |
| minor-fixes | 185.65user | 16.14user |
| smapAccumL-in-type-lift | 181.73user | 16.95user |
| smap-in-symbolize | 181.09user | 15.54user |
| readd-smap-Pat-Pat | 184.14user | 16.80user |
| readd-smap-Expr-Expr | 188.80user | 16.16user |
| readd-smap-Type-Type | 191.04user | 17.92user |

Notably, readding the explicit smaps either makes no difference (it might be within runtime variance) or is slightly slower. I would thus suggest that if we accept #399 we should also accept this one, but without the explicit smaps.

dlunde:
Sounds good Viktor. I did not quite understand the reference to #361, do you mean you used that PR as a baseline?

elegios:
I mentioned the wrong PR, oops

elegios:
This PR now no longer contains the explicit smap functions, since they did not give any speedup after #399 (actually a little bit of slowdown). This PR and #399 should be independent, so it should be possible to merge in any order, but it might be nicer to merge #399 first, since it'll make things go faster after.

One additional note on timing that I should have mentioned in the previous comment: changing `type-lift` and `symbolize` to use more `smap` stuff appears to make bootstrapping faster, presumably because there is less code to handle (or we're just seeing runtime variance, can't quite rule that out, but 1.13s faster seems decently likely to be an actual speedup).

PR #394, opened 2021-05-31
Update C pattern compilation and switch to using names for C struct and union members

dlunde:
## Major
- Change C pattern compilation to not generate nested derefs with ->
- Switch to using names, rather than strings, for C struct and union members.

## Minor
- Profiling patch from @lingmar  (prints in sorted order)
- Add some type annotations to `bool.mc` (needed for C compiler)
- Add `sfold_CStmt_CStmt` and `sfold_CTop_CStmt`.
- Add a few needed operators to C compiler.
- Add keywords handling in C pprint.
- Add external log function. **Discussion**: Naming, `log` vs `externalLog`?
- Add `mapToList` in `stdlib/map.mc` (makes dprinting maps more convenient)
- Improve performance of `any` in `stdlib/seq.mc` (now short-circuits)

### DIFF
--- a/stdlib/ext/math-ext.mc
+++ b/stdlib/ext/math-ext.mc
@@ -44,6 +44,3 @@ using eqf
 external externalAtan2 : Float -> Float -> Float
 let atan2 = lam x. lam y. externalAtan2 x y
 utest atan2 0. 1. with 0. using eqf
-
-external log : Float -> Float
-utest log 1. with 0. using _eqf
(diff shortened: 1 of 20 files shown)


stdlib/ext/math-ext.mc
```diff
@@ -44,3 +44,6 @@ using eqf
 external externalAtan2 : Float -> Float -> Float
 let atan2 = lam x. lam y. externalAtan2 x y
 utest atan2 0. 1. with 0. using eqf
+
+external log : Float -> Float
```
br4sco: The log function implementation is part of my PR as well. I do think (which I guess is what you want to discuss) we should prefix externals with external and then wrap them in lambdas to allow currying, i.e. `let log = lam x. externalLog x`

dlunde: Oh, I see. For some reason I only saw the `external` definitions above, and not the `let`s. That sounds like good approach. I'll remove the addition here to avoid conflicts with your PR

PR #395, opened 2021-06-01
Let type propagation

elegios:
This change makes `type-annot` aware of type annotations on `let`-bindings, in particular it now propagates function types to the parameters of lambdas, making things like the following work:

```
let snd: (a, b) -> b =
  lam x. x.1
```

Previously we would not have needed an additional annotation on `x`:

```
let snd: (a, b) -> b =
  lam x: (a, b). x.1
```

Sidenote: we're moving ever closer to a very ad-hoc typechecker with this, we should probably do something a bit more rigorous about that sooner rather than later.

WIP while tests are running


dlunde:
Great, this has been on my list of things to do for a long time now. I agree, it feels like we are moving closer and closer to type checking. In fact, I'm using the type annotator much like a type checker for the C compiler.

PR #401, opened 2021-06-11
Functions in fields of inlined records

elegios:
One of the intents of #376 was to make inlined records in constructors behave uniformly, in particular, make it so that every field has exactly type `Obj.t`. In actuality, it made things work this way when applying a constructor to an inlined record, but it did not make the appropriate change to the *declaration* of the constructor. Apparently none of our tests use a constructor with an explicit record, where at least one field has a function type, thus this did not trigger.

This PR fixes the type declarations.


larshum (approved):
Looks good to me!

PR #420, opened 2021-08-12
Solve major bootstrapping memory usage issue

larshum:
Solves an issue that caused the boot interpreter to use much more memory than necessary in the type lifting step of bootstrapping. For unknown reasons this was not a problem in the compiled compiler.


dlunde:
This is a good fix for now, but the question still remains why the original solution consumed so much memory in the interpreter. Should we maybe open an issue?

larshum:
I'll close this PR because #421 solves it in a better way.

PR #421, opened 2021-08-13
Re eval in delta

### LINKED ISSUE #406: Adding `foldl` and `foldr` as intrinsic functions in boot makes the bootstrapping take forever
The usual approach to implement intrinsic functions in the boot evaluator that take higher-order functions as arguments is to wrap the MCore AST node representing the function into a call to `eval`. E.g., this is how the boot interpreter evaluates `map`:
```ocaml
  | Cmap None, f ->
      let f x = eval env (TmApp (fi, f, x)) in
      TmConst (fi, Cmap (Some f))
  | Cmap (Some f), TmSeq (fi, s) ->
      TmSeq (fi, Mseq.map f s)
  | Cmap _, _ ->
      fail_constapp fi
```
However, when using this approach to implement `foldl` and `foldr` the first bootstrapping stage takes forever and eventually runs out of memory. See [this commit](https://github.com/miking-lang/miking/pull/389/commits/f4980abb08cb93eab562e7badda43cda162f4759#diff-5198c0da0dd2abcf5e63f5775dced5db14ab8309f9764c49bd9317ceb3ce68d3R847-R855) of PR #389. For reference, here is how this commit implements the evaluation of `foldl` and `foldr`:
```ocaml
  | Cfoldl (None, None), f ->
      let f a x = eval env (TmApp (fi, TmApp (fi, f, a), x)) in
      TmConst (fi, Cfoldl (Some f, None))
  | Cfoldl (Some f, None), a ->
      TmConst (fi, Cfoldl (Some f, Some a))
  | Cfoldl (Some f, Some a), TmSeq (_, s) ->
      Mseq.Helpers.fold_left f a s
  | Cfoldl _, _ ->
      fail_constapp fi
  | Cfoldr (None, None), f ->
      let f x a = eval env (TmApp (fi, TmApp (fi, f, x), a)) in
      TmConst (fi, Cfoldr (Some f, None))
  | Cfoldr (Some f, None), a ->
      TmConst (fi, Cfoldr (Some f, Some a))
  | Cfoldr (Some f, Some a), TmSeq (_, s) ->
      Mseq.Helpers.fold_right f a s
  | Cfoldr _, _ ->
      fail_constapp fi
```
Having `foldl` and `foldr` as builtins quite important when compiling to OCaml though, especially for the Rope sequence representation.
As a workaround, #402 makes the following hack: 
* In boot, in the builtin environment, bind `foldl` and `foldr` to ASTs representing an implementation of these functions.
* In the variable lookup rule in eval in boot, evaluate a `TmRecLet` if the variable is bound to that (so that the `fold` ASTs are evaluated).
* In the MCore compiler they are implemented the normal way, so that when compiling to OCaml the efficient OCaml version are used (`Mseq.Helpers.fold_*`).

I am not sure why the first implementation strategy runs out of memory. If we ever find the reason we should probably find a better solution to the hack.

elegios:
Fixes #406.

# The story and the impact

This PR is the culmination of a bunch of issues and looking into things by multiple people : @lingmar, @larshum, @dlunde, and myself, then finally identified by @larshum.

The first step of bootstrapping has been taking a lot of time and memory, which we mostly attributed to interpretation being slow and then didn't look into all that much, since `boot` is built to be temporary. It did however lead to #406 and a pretty hefty RAM requirement.

The issue turns out to be in the `delta` function, which handles application of constants to arguments. In particular, the constants that require functions as arguments (e.g. `map`) would later call the given function roughly like this:

```
eval env (TmApp (fi, f, a))
```

However, this will re-evaluate both `f` and `a`, even though they're already fully evaluated. This means that we go through the two terms completely, allocating new (identical) nodes as we go. The `type-lift` phase in particular ran into this really badly, taking about 90s out of 210s for me, and allocating lots of memory. After the fix, the first phase of bootstrapping takes about 116s.

# The fix

We create a new function `apply` that assumes its arguments are already fully evaluated, then applies the first (which must be a function of some kind) to the second. `eval` calls that when evaluating a `TmApp`, and all uses of `eval` in `delta` have been replaced by `apply`.

As a result, we can now also remove the hack in #402 and implement `foldl` and `foldr` as builtins as we'd normally do, so that's also in this PR.


dlunde (approved):
Fantastic :tada:

larshum (approved):
Looks good to me.

PR #425, opened 2021-08-18
Remove `batteries` as a dependency

lingmar:
* Removes or replaces everything that depends on `batteries`
  - `createFingerTree` is removed
  - `BatOption` is replaced by `Option` module
  - Some operations using `BatString` in `repl.ml` is rewritten
  - The implementation of `readBytesAsString` is rewritten with stdlib functions
* There remains an external test using `batteries`. It's not run with `make test-all`. But if you try to use the test external `batteriesZero` in a program you will get a compile error. I guess we should replace this test.
* The `threads` library in the dune file is needed for `Thread.delay` (used for `sleepMs` intrinsic).
* For some reason some formatting was made by `make fix` in the `constructor_tag` function.


larshum (approved):
Looks good!

PR #433, opened 2021-08-26
Move type ascription removal to separate transformation

dlunde:


larshum (commented):
You should use `removeTypeAscription` in `compile.mc` as well. I would suggest to place it after `typeAnnot` in https://github.com/miking-lang/miking/blob/86862a4169949df740f965622188bed0d4fd3057/src/main/compile.mc#L103-L105.

Apart from that I think it looks good.

dlunde:
My bad, I `grep`ed for `typeAnnot` only over `stdlib/`. I now added it to `src/main/eval.mc` and `src/main/compile.mc` as well.

PR #434, opened 2021-08-27
Updated readme to enable pyml installation.

david-broman:
There seems to be an issue with OPAM when installing `pyml` and using the multicore switch at the same time. This PR updates the readme and states how to solve it using a workaround.


david-broman:
As decided on a previous Miking meeting, I am merging this readme update directly.

PR #437, opened 2021-08-31
MExpr instrumented profiling

larshum:
This is the result of a side-project I have been working on during evenings and weekend the past few weeks. I have waited with opening a PR because I wanted to convince myself that it was mostly working correctly (which it didn't until yesterday), that the performance overhead is acceptable (which it kind of is, more on that later), and that it can actually be useful (I used this to identify the functions that I optimized in #436).

##

This PR adds instrumented profiling to `mi eval` and `mi compile`, made available through the  `--debug-profiling` flag. When this flag is used, the compiler will instrument code which generates a profiling report into the AST. I placed it prior to the generation of utests, so that we do not profile the internal utest functions. The instrumented code collects the following information for each top-level function:
* The position where the function is been defined (based on its info field)
* Its name
* The number of times it has been called, including self-recursive calls
* The exclusive and inclusive time spent in the function. The exclusive time only includes the time when the function was on the top of the call stack.

The above information is not collected for intrinsic functions. After the program execution has finished, the profiling data of all functions that were called at least once is written to a file “mexpr.prof” (this filename is currently hardcoded). The output is sorted by to the order in which the functions were defined in the AST. It is formatted such that it should be easy to sort it using the `sort` command - for example, to sort the functions by exclusive time (with the most time at the top), you can use the following command to store the sorted data in `out.txt`:

```cat mexpr.prof | sort -k4 -g -r > out.txt```

## Performance
The performance overhead of using this flag is significant, especially for the evaluator. I think it would be better to do something similar natively in the evaluator, because the overhead of the profiling becomes very big for anything but the smallest files in the standard library. In the compiler the overhead is negligible for smaller files, while for the largest files both the compilation time and the runtime take a noticeable hit. As an example of what the overhead might look like for a larger file, I measured the average compilation and execution time of `stdlib/mexpr/eval.mc` over 10 runs, with tests enabled:
| 				| Compilation time	| Execution time (compiled)	| Execution time (evaluated)  |
|-------------------|-------------------|-----------------------------|----------------------------|
| Without profiling	| 4.38 s			| 0.006 s					| 2.20 s					|
| With profiling	| 4.84 s		| 0.018 s					| 10.25 s				|


david-broman:
What is the overhead if the flag is not enabled?

larshum:
If the flag is not enabled then the AST will not be instrumented, just like how the unit testing is implemented. So the only overhead when compiling or evaluating should be negligible - this would be the extra time for processing the new profiling code (which should be small because of how little code it is) plus one extra branch for checking whether that flag is enabled. The compiled binary will have zero overhead.

PR #442, opened 2021-09-06
More optimization of standard library functions

larshum:
This PR optimizes frequently used standard library functions, further decreasing bootstrapping time of the first stage by another 10-15 seconds compared to #436 (on my machine).

### DIFF
--- a/stdlib/ocaml/pprint.mc
+++ b/stdlib/ocaml/pprint.mc
@@ -7,7 +7,8 @@ include "name.mc"
 include "intrinsics-ops.mc"
 
 let _isValidChar = lam c.
-  or (isAlphanum c) (or (eqChar c '_') (eqChar c '\''))
+  if isAlphanum c then true
+  else eqChar c '\''
 
 let _escapeChar = lam c.
   if _isValidChar c then c else '_'
--- a/stdlib/seq.mc
+++ b/stdlib/seq.mc
@@ -13,14 +13,18 @@ let init = lam seq. subsequence seq 0 (subi (length seq) 1)
 utest init [2,3,5] with [2,3]
 utest last [2,4,8] with 8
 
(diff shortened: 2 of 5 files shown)


stdlib/ocaml/pprint.mc
```diff
@@ -7,7 +7,8 @@ include "name.mc"
 include "intrinsics-ops.mc"
 
 let _isValidChar = lam c.
-  or (isAlphanum c) (or (eqChar c '_') (eqChar c '\''))
+  if isAlphanum c then true
+  else eqChar c '\''
```
lingmar, line 11: The old version accepts `'_'`, should the new version also do that?

stdlib/seq.mc
```diff
@@ -191,12 +195,14 @@ utest filter (lam x. eqi x 1) [1,2,4] with [1]
...
-    else find p (tail seq)
-end
+let find = lam p. lam seq.
+  foldl
+    (lam acc. lam x.
+      match acc with Some _ then acc
+      else if p x then Some x
+      else None ())
+    (None ())
+    seq
```
lingmar: The old version can exit early while the new version needs to go through the entire sequence. But I'm guessing the new version is cheaper in many cases because of `foldl`?

stdlib/ocaml/pprint.mc
larshum: The reason I removed it is because we escape invalid characters with `'_'`, so it's saving one extra computational step. And as far as I know that's the only thing we're using this function for.

stdlib/seq.mc
larshum: Yes. At least in `boot` it was a few seconds faster. I'm not sure about the compiled version though (it runs so fast that it's hard to distinguish improvements from runtime variances). We should probably have benchmarks specifically for functions like this.

PR #446, opened 2021-09-07
Make test-boot.mk set stdlib path

aathn:
Previously, `test-boot.mk` did not export the path to `stdlib`, which lead to weird behaviour when an old or erroneous copy of the standard library was installed on the system (in `~/.local/lib`).

To reproduce:

    make clean
    rm -rf ~/.local/lib/mcore/stdlib
    cp -r stdlib ~/.local/lib/mcore
    rm ~/.local/lib/mcore/stdlib/option.mc
    make test

This PR fixes this bug by setting

    MCORE_STDLIB=`pwd`/stdlib

before running the tests.


dlunde (approved):
Looks good, thanks

PR #447, opened 2021-09-07
Add lightweight bootstrapping-only compiler

larshum:
This PR adds a lightweight compiler `mi-lite.mc`, which is only meant to be used in place of `mi.mc` in the bootstrapping to speed up the compilation times. The point of having `mi-lite.mc` is to exclude features of the compiler that are irrelevant to the bootstrapping, such as the compiled interpreter and tuning. This means adding new features to the compiler that are not needed for bootstrapping, such as the profiling of #437, will not result in longer bootstrapping times for the first stage.

After this change, the steps to perform bootstrapping is as follows (the make script has been updated accordingly):
1. Bootstrapping `mi-lite.mc` using `boot eval src/main/mi-lite.mc -- 0 src/main/mi-lite.mc`
2. Bootstrapping the "complete" compiler using `./mi-lite 1 src/main/mi.mc`

(the argument 0/1 is used to disable/enable optimizations)


aathn:
Sweet! Shouldn't we then add a third stage of bootstrapping though, which compiles `mi` using itself? I imagine that (at least down the line) the full compiler would make use of additional optimizations and whatnot, which would not be included in `mi-lite`.

larshum:
> Sweet! Shouldn't we then add a third stage of bootstrapping though, which compiles `mi` using itself? I imagine that (at least down the line) the full compiler would make use of additional optimizations and whatnot, which would not be included in `mi-lite`.

Yes. I didn't want to add it because currently the compiler doesn't perform any optimizations of its own, so would just be a waste of time. Let's discuss it in the next meeting.

src/main/mi-lite.mc
```diff
@@ -0,0 +1,114 @@
...
+  match strLastIndex '/' path with Some idx then
+    subsequence path (addi idx 1) (length path)
+  else path
+
+let filenameWithoutExtension = lam filename.
+  match strLastIndex '.' filename with Some idx then
+    subsequence filename 0 idx
+  else filename
+
+let collectLibraries = lam extNameMap : ExternalNameMap.
```
br4sco: This removes changes made to `collectLibraries` in #449

src/main/mi-lite.mc
```diff
@@ -0,0 +1,114 @@
...
+  let g =
+    lam libs. lam impl : ExternalImpl. foldl f libs impl.libraries
+  in
+  let h = lam libs. lam. lam impls. foldl g libs impls in
+  let libs =
+    mapFoldWithKey h (setEmpty cmpString) extNameMap
+  in
+  setToSeq libs
+
+let ocamlCompile : Options -> [String] -> String -> String -> Unit =
```
br4sco: This removes changes made to `ocamlCompile` in #449

src/main/mi-lite.mc
larshum: It should be fixed now.

src/main/mi-lite.mc
larshum: It should be fixed now.

PR #448, opened 2021-09-07
Improve Makefile setup

aathn:
This PR refactors `./make` and `Makefile` and adds some quality-of-life changes.

- Use dune's installation of the `boot` binary instead of copying `build/boot` to `~/.local/bin`
- Refactor `./make` and `Makefile` to make them more consistent and legible
- Add `build-mi` make target, which recompiles using the current version of `mi` (skipping bootstrapping), creating the binary `mi-tmp`
- Add `uninstall` make target, which uninstalls all installed files

### DIFF
--- a/test-boot.mk
+++ b/test-boot.mk
@@ -1,30 +1,31 @@
 
-BOOT_NAME=boot
+BOOT_NAME = boot
 
-base-files=
-base-files+=$(wildcard test/mexpr/*.mc)
-base-files+=$(wildcard test/mlang/*.mc)
-base-files+=$(wildcard stdlib/mexpr/*.mc)
-base-files+=$(wildcard stdlib/c/*.mc)
-base-files+=$(wildcard stdlib/ad/*.mc)
-base-files+=$(wildcard stdlib/parser/*.mc)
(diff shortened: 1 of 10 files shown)


test-boot.mk
```diff
@@ -1,35 +1,31 @@
...
+base-files += $(wildcard test/mlang/*.mc)
+base-files += $(wildcard stdlib/mexpr/*.mc)
+base-files += $(wildcard stdlib/c/*.mc)
+base-files += $(wildcard stdlib/ad/*.mc)
+base-files += $(wildcard stdlib/parser/*.mc)
+base-files += $(wildcard stdlib/*.mc)
 
-stdlib-root-files=$(wildcard stdlib/*.mc)
+sundials-files += $(wildcard test/sundials/*.mc)
+sundials-files += $(wildcard stdlib/sundials/*.mc)
```
br4sco: The sundials implementation should not be tested with boot since they are "compiled only" now so those files should be removed.

aathn: Done!

br4sco: Thx!

PR #453, opened 2021-09-11
Add lint back to test-all

br4sco:


david-broman:
I think we need to discuss this at the next meeting again. We need to decide when this check should happen. If it is during installation, everyone (not just contributors) need to have ocamlformat installed.

PR #454, opened 2021-09-11
Delete .merlin file

aathn:
This PR removes the `.merlin` file from the project's root folder. The current contents of this file makes merlin unable to provide proper language support, on my machine at least.

The readme of the [merlin project](https://github.com/ocaml/merlin) states the following:

> To use Merlin with a multi-file project, it is necessary to have a .merlin file unless your project is built using dune. Note that, in a project using Dune, user-created .merlin files will take precedence over the configuration provided by Dune to Merlin.

Hence, it should be fine to remove it.


aathn:
This PR is now included in https://github.com/miking-lang/miking/pull/458

PR #455, opened 2021-09-12
Fix REPL autocompletion

aathn:
This PR fixes boot's REPL autocompletion, which was barely working due to two subtle errors in `repl.ml`. The first was in the following line:

    Str.search_backward nonword_characters str (pos - 1) + 1

If `pos` is 0 (i.e. the cursor is right at the prompt), the argument to `Str.search_backward` becomes -1, causing an error. `Str.search_backward` probably did not error on negative arguments when the REPL code was first written, or we would have noticed.

The second was in the following lines:

    let len = min (String.length str) (pos - start_pos + 1) in
    let word_to_complete = String.sub str start_pos len in

It turns out that `len` here should simply be `pos - start_pos`, and the above version gives an off-by-one error.

I also took the opportunity to polish the code a bit, but only very minor changes.

### DIFF
--- a/src/boot/lib/repl.ml
+++ b/src/boot/lib/repl.ml
@@ -82,9 +82,9 @@ let handle_command str =
   | ":q" ->
       save_history_and_exit ()
   | ":h" ->
-      print_endline help_message ; true
+      print_endline help_message ; Some ()
   | _ ->
-      false
+      None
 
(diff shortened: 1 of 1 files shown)


aathn:
This PR is now included in https://github.com/miking-lang/miking/pull/458

PR #459, opened 2021-09-13
Unified collection syntax

aathn:
This is a draft PR adding syntax and primitives for unified collections (UCs). A collection can be created and manipulated using the three primitives

```
empty : Coll p a
insert : a -> Coll p a -> Coll p a
fold : (a -> b -> b) -> b -> Coll p a -> b
```

As of now, these are simply implemented by MCore sequences (`[]`, `cons` and `foldr`), and the `p` in the collection type is ignored. The final goal is to have semantic properties, such as Sequential / Nonsequential and Unique / Multiple elements, specifying the collection's behavior and allowing us to express most collection types in a uniform way. We would use automatic data structure selection to pick the most efficient representation in each use case, and representations would be able to provide a larger set of operations than just the three primitives above.

The following example program illustrates the idea.
```
let append = lam c1. lam c2. fold insert c2 c1
let size = fold (lam. addi 1) 0

recursive let myFun : Int -> a -> Coll p a =
  lam n. lam x.
    if lti n 1 then
      insert x empty
    else
      let c = myFun (subi n 1) x in
      append c c
end

let mySequence : Coll {}       = myFun 3 1
let mySet      : Coll {UQ, NS} = myFun 3 1

mexpr
utest size mySequence with 8 in
utest size mySet      with 1 in
()
```
Here, `UQ` stands for unique elements, and `NS` stands for nonsequential (i.e. unordered), and the empty property set corresponds to a sequence (this interface is up for discussion). Currently, the above program can be run, but the second test fails, since properties are not implemented yet.


aathn:
Closing this for now, as we have more thoughts about the design that we probably want to consider.

PR #460, opened 2021-09-14
Added a number of probability distributions

david-broman:
This PR adds sample functions and probability mass functions (PMFs) or probability density functions (PDFs) for Beta, Gaussian, Bernoulli, Binomial, and Uniform distributions. It is implement as external functions using the OWL OCaml library.

### DIFF
--- a/stdlib/ext/dist-ext.ext-ocaml.mc
+++ b/stdlib/ext/dist-ext.ext-ocaml.mc
@@ -9,5 +9,16 @@ let distExtMap =
   [
     ("myExternalExp", [
       { ident = "Float.exp", ty = tyarrow_ tyfloat_ tyfloat_ , libraries = [] }
+    ]),
+    ("externalBinomialLogPmf", [
+      { ident = "Owl_stats.binomial_logpdf",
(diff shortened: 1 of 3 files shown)


dlunde (approved):

stdlib/ext/dist-ext.ext-ocaml.mc
```diff
@@ -0,0 +1,59 @@
+
+
+include "map.mc"
+include "ocaml/ast.mc"
+
+let distExtMap =
+  use OCamlTypeAst in
+  mapFromSeq cmpString
+  [
```
br4sco, line 9: The external implementation record has been extended with a field `cLibraries` to allow specifying c libraries that are needed during linking. I.e. the type has been changed to
```
type ExternalImpl = {
  ident : String,
  ty : Type,
  libraries : [String],
  cLibraries : [String]
}
```
so you should add a field `cLibraries = []` to all of these.

david-broman: Thanks. Fixed.

PR #467, opened 2021-09-21
Fix issues with lint and parallel test-all

dlunde:
This fixes the parallel `test-all` issues with `lint` encountered by me, @lingmar, and @larshum. It just moves `lint` from the prerequisite list to the recipe body. This forces lint to be run last.

Note: the problem seems to originate from building multiple dune targets in parallel through the makefile. For example, just adding `lint` as a prerequisite to the `boot` rule also solves this (but that disallows building `boot` without having ocamlformat installed).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,12 +30,12 @@
   test-boot-py\
   test-boot-ocaml\
   test-sundials\
-	test-ipopt\
+  test-ipopt\
   test-par
 
 all: build
 
-boot:
+boot: lint
 	@./make boot
 
 install-boot: boot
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -30,12 +30,12 @@
...
   test-boot-ocaml\
   test-sundials\
-	test-ipopt\
+  test-ipopt\
   test-par
 
 all: build
 
-boot:
+boot: lint
```
aathn: Having `lint` as a dependency for `boot` will force all users to run it, won't it? Shouldn't it be on `test-boot-base` instead?

dlunde: For some reason, it is needed as a prereq of `boot` specifically. If I put it on `test-boot-base` instead, it gives the error. I think the issue is building multiple dune targets in parallel through make.

PR #469, opened 2021-09-22
Optimize the graph library

lingmar:
The main purpose of this PR is to merge an improvement of the `digraph.mc` library that I did a while back. The `lamlift.mc` and some `accelerate` things are now using the graph library, so they will benefit from this change. The main change is that the graph uses map intrinsics instead of `assocMap`. I have not benchmarked the graph library properly, but the performance of the graph operations for tuning improved a lot after this change.

It turned out I had to include a lot of changes in the `tuning` folder that were made after the `digraph.mc` update, hence the large diff.

Note that this PR includes #462.


larshum (approved):
After having a quick look, I think this looks good.

I found that the performance of lambda lifting is significantly improved after the changes here. Running lambda lifting on the full compiler AST went from ≈27s to ≈15s after the changes here (it's still slower than it was before, but that's because we have a lot more code). 👍

br4sco:
Great! means my BLT sorting is faster as well

PR #471, opened 2021-09-24
Fix iteri

br4sco:
This PR fixes a bug in `iteri` where the iteration variable were incorrect for some concatenated sequences.

Before the changes in this PR, this test
```
utest
 let r = ref [] in
 let x = concat (concat (concat [] [0]) [1]) [2] in
 iteri (lam i. lam x. modref r (snoc (deref r) (i, x))) x;
 deref r
with [(0, 0), (1, 1), (2, 2)] in
```
fails with the output
```
    LHS: [(0,0),(1,1),(1,2)]
    RHS: [(0,0),(1,1),(2,2)]
```

### DIFF
--- a/src/boot/lib/rope.ml
+++ b/src/boot/lib/rope.ml
@@ -161,8 +161,9 @@ let iteri_array (f : int -> 'a -> unit) (s : 'a t) : unit =
         done ;
         len
     | Concat {lhs; rhs; _} ->
-        let n = iteri off lhs in
-        iteri (off + n) rhs
+        let n1 = iteri off lhs in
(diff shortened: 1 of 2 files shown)


lingmar:
Darn, I recently "fixed" the `iteri` indices for `Slice`, but apparently I didn't get it right...

br4sco:
> Darn, I recently "fixed" the `iteri` indices for `Slice`, but apparently I didn't get it right...

No I think `Slice` was correct actually. I should revert that change to yours.

larshum:
Looks good to me!

Issue #231, opened 2020-12-11
Change the syntax of the regular pattern operators (and, or, not) to be prefix


elegios:
The regular pattern operators use infix syntax (`&`, `|`, `!`) with precedence and associativity, when everything else in the language is prefix. This seems inconsistent, so it should probably be changed. We've decided to leave it as is until we have a Ragnar implementation, to lessen the amount of busy work right now.

Once we do this we need to decide on syntax. Assuming we pick words (e.g. `and`, `or`, `not`) we have essentially three options:

- Make them keywords. This clobbers the names for the standard library and/or intrinsics (unless we're ok with always using `#var"and"` instead of `and`).
- Special case them and only treat them as keywords in a pattern. This has at least two alternative approaches:
  - The words are always reserved, i.e., if we want to bind a value to the corresponding identifier we have to use `#var"and"`.
  - The words are only treated specially when they're followed by another pattern, e.g., `match v with and then ...` is equivalent with `let #var"and" = v in ...`, while `match v with and a b then ...` is what we would presently write as `match v with a & b then ...`. This starts looking a little bit like syntax that could allow first-class/custom patterns, if that's a thing we want to look into down the road, but it feels like it would be too easy to make a mistake and/or make it hard to follow.

Another alternative is to pick a word with some extra signifier, e.g., `#and`:

```
match v with #and a b then ...
```

This also leaves the door open for first-class/custom patterns, but without the special casing, and with the additional option of having zero arity first class patterns.

PR #475, opened 2021-09-28
Type checker

aathn:
Edit: The PR is now ready for merging, as soon as we agree on the syntax. I have removed the macros, but left the freeze syntax unchanged until we reach a decision.

This PR adds a type checker with [FreezeML](https://dl.acm.org/doi/abs/10.1145/3385412.3386003)-style annotations and inference. The type checker is relatively complete, but currently lacks support for
- Recursive lets
- Tuples / records and variants
- Algebraic data types and type aliases
- Tensors
- `match` forms

~This is a draft PR, since~ The new code currently lacks:
- ~Test cases~
- Proper error propagation ~and reporting~ (when unification fails, an error is reported for the "innermost" failure)

To try out the type checker, use the `typeCheck` function of `MExprTypeCheck`, located in `mexpr/type-check.mc`. It works in the same way as `typeAnnot`. For details on the type system, check the paper linked above.

A few examples which demonstrate the syntax for the new annotations:

    lam x. x                      : _a -> _a       -- _a denotes a monomorphic type var
    let f = lam x. x in f         : a -> a         -- f's quantifiers are instantiated
    let f = lam x. x in `f        : all a. a -> a  -- the backtick ` is the freeze operator
~${lam x. x}                   : all a. a -> a  -- ${...} is the generalization macro~
~let f = lam x. x in {`f}@ 5   : Int            -- {...}@ is the instantiation macro~

The implementation is based on the paper, as well as the description of efficient unification given at http://okmij.org/ftp/ML/generalization.html.


src/boot/lib/symbolize.ml
```diff
@@ -92,12 +93,14 @@ let rec symbolize_type env ty =
       ty
   | TyVariant _ ->
       failwith "Symbolizing non-empty variant types not yet supported"
-  | TyVar (fi, x, s) ->
+  | TyCon (fi, x, s) ->
+      (* TODO(aathn,2021-09-25): This should not be needed anymore, since
+         the unbound type variables are now TyVar and this is TyCon *)
```
elegios, line 98: I would agree with this. I'm guessing though that just removing it runs into cases where we have types on the toplevel that refers to `syn` types which aren't in scope, right? If that's the case we should probably open an issue for making a way to refer to such types from the toplevel and fixing the issues, then removing the `try` here.

stdlib/ext/ext-test.ext-ocaml.mc
```diff
@@ -252,18 +252,18 @@ let extTestMap =
       impl
       {
         ident = "List.map",
-        ty = tyarrows_ [tyarrow_ (tyvar_ "a") (tyvar_ "b"),
-                        otylist_ (tyvar_ "a"),
-                        otylist_ (tyvar_ "b")]
+        ty = tyarrows_ [tyarrow_ (tycon_ "a") (tycon_ "b"),
```
elegios: I think these are actually supposed to be `tyvar_` with an `tyall_` before them, since it's supposed to be `List.map`.

stdlib/ext/ext-test.ext-ocaml.mc
```diff
@@ -252,18 +252,18 @@ let extTestMap =
...
       }
     ]),
     ("extTestListConcatMap", [
       impl
       {
         ident = "List.concat_map",
-        ty = tyarrows_ [tyarrow_ (tyvar_ "a") (otylist_ (tyvar_ "b")),
-                        otylist_ (tyvar_ "a"),
-                        otylist_ (tyvar_ "b")]
+        ty = tyarrows_ [tyarrow_ (tycon_ "a") (otylist_ (tycon_ "b")),
```
elegios: Also `tyvar_`.

src/boot/lib/ast.ml
```diff
@@ -346,8 +348,10 @@ and ty =
   | TyRecord of info * ty Record.t * ustring list
   (* Variant type *)
   | TyVariant of info * (ustring * Symb.t) list
+  (* Type constructors *)
+  | TyCon of info * ustring * Symb.t
```
elegios, line 356: I vaguely feel like `TyCon` shouldn't need to be symbolized either, since `boot` shouldn't care about types at all, but I'm guessing I'm wrong? Something to do with `mlang` I guess?

stdlib/mexpr/ast.mc
```diff
@@ -1188,16 +1189,75 @@ lang VariantTypeAst = Ast
...
+  | TyFlex t -> TyFlex {t with info = info}
 
   sem infoTy =
-  | TyVar r -> r.info
+  | TyVar t -> t.info
+  | TyFlex t -> t.info
+
+  sem smapAccumL_Type_Type (f : acc -> a -> (acc, b)) (acc : acc) =
+  | TyFlex t & ty1 ->
+    match deref t.contents with Link ty2 then
```
elegios, line 1252: I'm not entirely sure that `smapAccumL` should process the contents of a `Link`. The current hypothesis is that `smapAccumL` should give a simple syntactic recursion, but I suspect that the link is rather an annotation "this is the same as this other thing". C.f. `smapAccumL_Expr_Type` that doesn't examine the `ty` field, since it's an annotation.

If it's supposed to behave exactly like the `Link`ed type then I guess all of the helpers here (`ty`, `info`, etc.) should fully resolve the link before doing their work?

stdlib/mexpr/cmp.mc
```diff
@@ -392,9 +392,29 @@ lang VariantTypeCmp = Cmp + VariantTypeAst
...
 lang VarTypeCmp = Cmp + VarTypeAst
+  sem cmpTVar =
+  | (Unbound {ident = n}, Unbound {ident = m}) -> nameCmp n m
+  | (Link ty1, Link ty2) -> cmpType ty1 ty2
+  | (tv1, tv2) ->
+    subi (constructorTag tv1) (constructorTag tv2)
+
   sem cmpTypeH =
   | (TyVar t1, TyVar t2) -> nameCmp t1.ident t2.ident
+  | (TyFlex t1, TyFlex t2) -> cmpTVar (deref t1.contents, deref t2.contents)
```
elegios: I think you mentioned some uncertainty on whether comparison was correct, I'm guessing this was it, because it's not obvious to me how it should be :P

In general this comparison is syntactic, it ignores `info` and `ty`. I'm assuming the idea behind `TyFlex` and `Link` is that the type is the same as the one pointed to by `Link`, and that they take the place of explicit substitution, so I'm guessing that a `TyFlex` should be fully resolved before the comparison?

stdlib/mexpr/eq.mc
```diff
@@ -97,7 +100,12 @@ lang Eq
   sem eqExprH (env : EqEnv) (free : EqEnv) (lhs : Expr) =
   -- Intentionally left blank
 
-  sem eqType (typeEnv : EqTypeEnv) (lhs : Type) =
+  sem eqType (lhs : Type) =
```
elegios, line 106: I think the bidirectional map for free variables needs to be passed along through everything, similarly to how it's done for expressions. Otherwise these two would be equal:

```
type A = a in
type B = a in
type C = b in
()
```
```
type A = a in
type B = b in
type C = b in
()
```

stdlib/mexpr/eq.mc
```diff
@@ -480,103 +488,135 @@ end
...
+  | TyVar t1 ->
+    match lhs with Some (TyVar t2) then
+      if nameEq t1.ident t2.ident then true
+      else
+        match biLookup (t1.ident, t2.ident) typeEnv.tyVarEnv with Some _ then true
+        else false
+    else false
+  | TyFlex t1 ->
+    match lhs with Some (TyFlex t2) then
+      eqTVar typeEnv (deref t1.contents, deref t2.contents)
```
elegios: I think the same thing applies here as for `cmp`, whatever reasoning ends up being correct there should also apply here.

stdlib/mexpr/pprint.mc
```diff
@@ -1310,7 +1337,7 @@ in
 --   match o with !(None ()) & Some _ then true else false
 let func_pedanticIsSome =
   ulet_ "pedanticIsSome" (
-    lam_ "s" (tyapp_ (tyvar_ "Option") (tyvar_ "a")) (
+    lam_ "s" (tyapp_ (tycon_ "Option") (tycon_ "a")) (
```
elegios: I guess this type is just wrong, it should be `all a. Option a -> Bool` but it's written as just `Option a`. Could you fix it on the way?

stdlib/mexpr/symbolize.mc
```diff
@@ -94,13 +98,18 @@ lang LamSym = Sym + LamAst + VarSym
...
-lang LetSym = Sym + LetAst
+lang LetSym = Sym + LetAst + AllTypeAst
   sem symbolizeExpr (env : SymEnv) =
   | TmLet t ->
     match env with {varEnv = varEnv} then
       let tyBody = symbolizeType env t.tyBody in
       let ty = symbolizeType env t.ty in
-      let body = symbolizeExpr env t.body in
+      let body =
+        match stripTyAll tyBody with (vars, _) then
```
elegios, line 108: I'm assuming this has the effect of, e.g., bringing `a` and `b` into scope in `body` here:

```
let f : all a. all b. a -> b = body
```

but not `c` in `body` here:

```
let f : Int -> all c. c -> Int = lam arg. body
```

That's probably a pretty reasonable way to do it. It might be ideal if we carried the type along and opened top-most `all`s as we enter various constructors (of which I count `lam` one for the moment), but then `symbolize` would need a very good understanding of the typesystem, so it's probably not a good idea.

stdlib/mexpr/symbolize.mc
```diff
@@ -255,22 +264,48 @@ lang VariantTypeSym = VariantTypeAst
...
+lang VarTypeSym = VarTypeAst + UnknownTypeAst
+  sem symbolizeType (env : SymEnv) =
+  | TyVar t & ty ->
+    if nameHasSym t.ident then ty
+    else
+      let str = nameGetStr t.ident in
+      match mapLookup str env.tyVarEnv with Some ident then
+        TyVar {t with ident = ident}
+      else
+        -- NOTE(aathn, 2021-09-29): Unknown type variables are symbolized
```
elegios, line 292: I assume this is a placeholder for now since we'd break a lot of code otherwise, and later we'd produce an error here?

stdlib/mexpr/type-check.mc
```diff
@@ -0,0 +1,436 @@
...
+  sem unificationError =
+  | (ty1, ty2) ->
+    let msg = join [
+      "Type check failed: unification failure\n",
+      "LHS: ", _pprintType ty1, "\n",
+      "RHS: ", _pprintType ty2, "\n"
+    ] in
+    infoErrorExit (infoTy ty1) msg
+
+  sem occurs (tv : TVar) =
```
elegios: Following the documentation discussion last meeting it's probably a good idea to put documentation on the semantic functions in this fragment (not the ones that just add cases, but the "root" definition).

stdlib/mexpr/type-check.mc
```diff
@@ -0,0 +1,436 @@
...
+    unifyWithNames (biInsert (t1.ident, t2.ident) names) (t1.ty, t2.ty)
+
+  sem occurs (tv : TVar) =
+  | TyAll t ->
+    match tv with Unbound {weak = true} then
+      let msg = join [
+        "Type check failed: unification failure\n",
+        "Attempted to unify monomorphic type variable with polymorphic type!\n"
+      ] in
+      infoErrorExit t.info msg
```
elegios, line 139: This feels like a weird place for this check, shouldn't it be directly in a `unify` case somewhere? This would preclude unifying `_a` with `[all b. b -> b]`, right?

stdlib/mexpr/type-check.mc
```diff
@@ -0,0 +1,436 @@
...
+lang VarTypeGeneralize = Generalize + VarTypeAst
+  sem instBase (subst : Map Name Type) =
+  | TyFlex t & ty1 ->
+    match deref t.contents with Link ty2 then
+      instBase subst ty2
+    else
+      ty1
+  | TyVar {ident = n} & ty ->
+    match mapLookup n subst with Some tyvar then tyvar
+    else ty
```
elegios, line 208: Is this `else` case possible given that code has passed `symbolize` properly (once we error on unbound type variables)? I'm guessing it is, because of the handling of `all a. ...` annotations on `let`, which aren't propagated in the typecheck case for `TmLet`?

stdlib/mexpr/type-check.mc
```diff
@@ -0,0 +1,436 @@
...
+
+lang FunTypeGeneralize = Generalize + FunTypeAst
+  sem instBase (subst : Map Name Type) =
+  | TyArrow r ->
+    TyArrow {{r with from = instBase subst r.from}
+                with to = instBase subst r.to}
+
+  sem genBase (lvl : Level) =
+  | TyArrow r ->
+    match genBase lvl r.from with (vars1, fromNew) then
```
elegios: I suspect `smapAccumL` would be useful here and for some other cases.

stdlib/mexpr/type-check.mc
```diff
@@ -0,0 +1,436 @@
...
+lang LetTypeCheck = TypeCheck + LetAst
+  sem typeCheckBase (env : TCEnv) =
+  | TmLet t ->
+    let lvl = env.currentLvl in
+    let body = typeCheckBase {env with currentLvl = addi 1 lvl} t.body in
+    let inexpr =
+      match t.tyBody with TyUnknown _ then
+        -- No type annotation: generalize the inferred type
+        let genBody = gen lvl (ty body) in
+        typeCheckBase (_insertVar t.ident genBody env) t.inexpr
```
elegios: There's an interesting thing here for later code that might want to traverse the tree and check the types of things, it can't easily recover the type of the value bound to `t.ident`, since it's neither in `t.tyBody` nor `ty t.body`. I feel like it should be recorded somewhere, probably by just updating `t.tyBody`.

stdlib/mexpr/type-check.mc
```diff
@@ -0,0 +1,436 @@
...
+              with ty = ityseq_ t.info elemTy}
+end
+
+lang UtestTypeCheck = TypeCheck + UtestAst
+  sem typeCheckBase (env : TCEnv) =
+  | TmUtest t ->
+    let test = typeCheckBase env t.test in
+    let expected = typeCheckBase env t.expected in
+    let next = typeCheckBase env t.next in
+    let tusing = optionMap (typeCheckBase env) t.tusing in
```
elegios, line 336: This should also do some unification:
- If `using` is present unify it with `a -> b -> Bool` where `a` and `b` are types of `test` and `expected`.
- If not, unify the types of `test` and `expected`.

stdlib/mexpr/type-check.mc
```diff
@@ -0,0 +1,436 @@
...
+    TmUtest {{{{{t with test = test}
+                   with expected = expected}
+                   with next = next}
+                   with tusing = tusing}
+                   with ty = ty next}
+end
+
+lang NeverTypeCheck = TypeCheck + NeverAst
+  sem typeCheckBase (env : TCEnv) =
+  | TmNever t -> TmNever {t with ty = ityunknown_ t.info}
```
elegios: I think this should be a new weak variable. Conceptually we have `never : all a. a`.

elegios (changes requested):
This seems really nice, and on the whole I think this is a really good step to merge (despite the wall of comments, sorry about that...).

Besides those comments I'd still like to change the syntax a little bit. Even if we don't go as far as I suggested during the meeting, I still feel like it's unnecessary to have three new sigils, two seem sufficient; one for generalization, one for instantiation.

In particular I'd suggest using `$` both for freezing a variable and generalizing an expression. It's still easy to talk about them as one thing being implemented using the other, and it feels reasonable that `$foo` and `${lam x. x}` do similar things to `foo` and `lam x. x` respectively. There's also the additional benefit of avoiding backtick, which is annoying to write in inline code in markdown :)

Putting `$` in prefix form and `@` in postfix feels inconsistent to me, but I don't have a strong sense of how it should be, whether one would prefer one form or another, so I guess following the paper is a reasonable default.

As a sidenote, do you have any sense of how well it would work if every expression instantiated itself? The difference between identifiers and expressions feel clunky, and is the reason why `@` is necessary at all, but I don't know if it, e.g., would make code that passes these polymorphic values around worse to work with, or if it would break some other property of the typesystem.

src/boot/lib/symbolize.ml
aathn: That's right. I agree.

stdlib/ext/ext-test.ext-ocaml.mc
aathn: Yeah, it seems so. The previous `TyVar` was used both as type variables and type constructors,so just replacing them all with `TyCon` was a bit sloppy of me. There's a number other type signatures as well in e.g. `const-types.mc` which I haven't updated yet, which have `TyUnknown` instead of type variables. I'll try to fix the stuff directly touched by this PR, but leave the rest for a new one.

src/boot/lib/ast.ml
aathn: Right, I just kept the machinery which was already there, but I don't think it's needed. For the record, the parsing functions in `bootparser` actually symbolize the AST, but then discard the symbols.

stdlib/mexpr/ast.mc
aathn: My understanding is that the link should behave as the `Link`ed type. When unifying an unbound `TyFlex` with another type, we would really like to replace the `TyFlex` with the other type, but that would incur a performance penalty, so we use references to `Link`s instead.  So, I should probably make the other helpers resolve the link as well, as you say.

stdlib/mexpr/cmp.mc
aathn: I think what I was referring to was the implementation of `eqType` for `TyVar` and `TyAll`, but I have since updated it to be correct.

You're right, I should resolve the links here as well.

stdlib/mexpr/eq.mc
aathn: Would they? Currently free variables are checked for symbol equality in `eqType`, so the `a` and the `b` would be different, wouldn't they?

As an aside, there doesn't actually seem to be a case in `eqExpr` for `TmType` currently.

stdlib/mexpr/pprint.mc
aathn: Sure!

stdlib/mexpr/symbolize.mc
aathn: That's right. This is exactly the way it's described in the FreezeML paper (but I guess we could also extend it as you say).

stdlib/mexpr/symbolize.mc
aathn: Yep.

stdlib/mexpr/type-check.mc
aathn: `occurs` currently performs three tasks, all of which require traversing the type to unify with, in a single traversal.
- Performing an occurs check
- Updating `level` fields of type variables
- Checking that monomorphic type variables are not unified with polymorphic types

I should probably rename the function, and add a comment explaining this.

It would not be possible to unify `_a` with `[all b. b->b]`, correct. The way I interpreted the paper, this is to prevent phenomena like the following,  where

    bad1 = lam f. (poly `f, (f 42) + 1)

type checks, but

    bad2 = lam f. ((f 42) + 1, poly `f)

does not (with `poly : (all a. a -> a) -> (Int, Bool)`).

stdlib/mexpr/eq.mc
elegios: Aren't they assumed to be equal when they're found (and both unbound) so that the equality works modulo alpha-conversion (i.e. `lam x. y` = `lam z. w`, we assume `y = w` when we find them unbound and see if we later find an inconsistent assumption)? I believe that's the case for `TmVar` (that's why it returns an `Option` and not just a `Bool`) and figured it was the same idea for free `TyVar`, in which case the "these are equal" assumption must carry through so we don't get inconsistent assumptions.

stdlib/mexpr/symbolize.mc
elegios: Nice, I didn't realize that. Certainly wait with that extension, if we ever do it :)

stdlib/mexpr/eq.mc
aathn: Right, I didn't implement that yet because of laziness and not wanting to add too much stuff to the PR, but I probably should.

stdlib/mexpr/type-check.mc
aathn: It is, yes. Suppose we have

    f : all a. a -> (all b. b -> a)

Then, instantiating `f`'s type requires going under the `all b.` quantifier and passing over the `b`, triggering the `else` branch

stdlib/mexpr/type-check.mc
aathn: I thought so too, I'll probably make the switch in the next PR. I also think that if we change `eqTypeH` to return a pair `(ty1, ty2)` whenever the equivalence check fails, we could probably remove a good part of the unification code as well, and only keep the crucial cases.

stdlib/mexpr/type-check.mc
aathn: Good catch. Yeah, updating `t.tyBody` seems sensible. Should I also update `t.tyIdent` for lambdas then, for consistency? That would make the pretty-printed AST display the types of all bindings after type-checking, which would be nice. The potential downside is that we forget whether a given annotation was provided by the programmer or if we added it afterwards, but I guess we don't really care about that.

stdlib/mexpr/type-check.mc
aathn: Alright. I was unsure of how we want to handle `never` and opted to go for the way it was in `type-annot.mc`.

aathn:
Thanks for the extensive feedback!

I agree with you on the syntax, changing the backtick for `$` seems sane. I also think it would be nicer to have `@` in prefix position, the reason I put it in postfix position was simply to follow the paper. Changing `$` and `@` to act more like operators, not requiring the braces, is probably a good idea as well.

Regarding instantiation, the following is from section 3.2 in the paper. I guess we should consider the pros and cons of each approach.
> Instantiation strategies. In FreezeML (and indeed ML) the only terms that are implicitly instantiated are variables. Thus (head ids) 42 is ill-typed and we must insert the instantiation operator @ to yield a type-correct expression: (head ids)@ 42. It is possible to extend our approach to perform eliminator instantiation, whereby we implicitly instantiate terms appearing in monomorphic elimination position (in particular application position), and thus, for instance, infer a type for bad5 without compromising completeness. Another possibility is to instantiate all terms, except those that are explicitly frozen or generalised. Here, it also makes sense to extend the ⌈−⌉ operator to act on arbitrary terms, rather than just variables. We call this strategy pervasive instantiation. Like eliminator instantiation, pervasive instantiation infers a type for (head ids) 42. However, pervasive instantiation requires inserting explicit generalisation where it was previously unnecessary. Moreover, pervasive instantiation complicates the meta-theory, requiring two mutually recursive typing judgements instead of just one. The formalism developed in this paper uses variable instantiation alone, but our implementation also supports eliminator instantiation. We defer further theoretical investigation of alternative strategies to future work.

stdlib/mexpr/type-check.mc
elegios: Ah, good point, did not think of that.

stdlib/mexpr/type-check.mc
elegios: But the issue in that example is about unifying the type of `f` with `all a. a -> a` vs `Int -> Int`, but `[all b. b -> b]` is guarded by the `[]` constructor (the difference between full monotypes and guarded types in the paper). I feel like that difference should matter, shouldn't it?

aathn: I don't think it matters: suppose `poly` has the type `[all a. a -> a] -> (Int, Bool)`. Then we can do

    bad1 = lam l. (poly l, (head l)@ 42 + 1)
    bad2 = lam l. ((head l)@ 42 + 1, poly l)

which has the same problem.

elegios:
I'm not entirely sure on using them as operators. I think it would definitely be correct in a higher-level language, but we're trying to avoid needing precedence and associativity to read code, to make the connection to the AST as clear as possible, so there is actually a point to having them not be unary operators syntactically. 

(It's worth mentioning that I'm being a little bit hypocritical here, I introduced syntactical operators for patterns, resulting in #231)

I vote we stick with the braces for now, if we want to change to operators later it shouldn't be too much work to do (just a matter of changing `{}` to `()`).

aathn:
Ah, makes sense. If we put `@` in prefix position, we could change the braces to parens regardless (doing it with the postfix variant produces a parser conflict). That would be a more lightweight syntax which is closer to the paper (and wouldn't require any change if we went the operator route), but wouldn't be as visually distinguishable. What do you think?

elegios:
I think stick with the braces for now, it’s possible that it’s confusing when they almost behave like operators but not quite.

aathn:
We discussed the syntax a bit during the meeting today. The conclusion was as follows:
> Our proposal is to add the keyword `freeze` as the only new syntax, since everything can be expressed with that. If we see that the programs become too verbose we can always add more syntax later. But we didn't take any decision about this, it's only a proposal.
> 
> Further, we think we should do "eliminator instantiation" to avoid explicitly having to instantiate expressions at function applications (later PR).

What's your take @elegios ?

aathn:
Personally, I don't mind removing the macros, especially if we have eliminator instantiation. But I don't think having a keyword for freeze is a good idea, I think it's worth dedicating a symbol such as `$`, `^`, or backtick to it.

david-broman:
> Personally, I don't mind removing the macros, especially if we have eliminator instantiation. But I don't think having a keyword for freeze is a good idea, I think it's worth dedicating a symbol such as `$` or backtick to it.

Anders, so you did not agree with it on the meeting? I thought you said in the meeting that you did not have any strong opinion, but maybe I misunderstood your point. Anyway, we should discuss it again on Tuesday. If there is no consensus, we need to make a vote at the meeting. This is usually how we handle it if there is no agreement.

src/boot/lib/ast.ml
david-broman: It would be good to give a type alias for `bool` in the `TmVar` case. Otherwise, it is hard to know what this boolean actually means (without trying to derive it from the code)

david-broman:
Maybe you can explain a bit more why you think it is not a good idea with a keyword? The reason why several of us at the meeting though it was a good idea was that (1) the overall design of MCore is to not introduce special symbols (there is not even a symbol for adding integer numbers), (2) we suspect that freezing variables will not be the most common usage in the language, compared to other standard operations that could use symbols.

Also, I think it is important to remember that MCore is not a programming language, but an intermediate language for the compiler, that you can program in. 

It would be great to know if you have some good argument that we might have missed at the meeting.

aathn:
> Anders, so you did not agree with it on the meeting? I thought you said in the meeting that you did not have any strong opinion, but maybe I misunderstood your point.

I changed my mind, having thought some more after the meeting. That said, I'm not too strongly opinionated about it.

> Maybe you can explain a bit more why you think it is not a good idea with a keyword?

One problem I see with having a `freeze` keyword, besides being verbose, is that `freeze` risks looking like a function. Supposing we use the natural syntax `freeze x`, freezing a variable looks just like function application, which may confuse inexperienced readers and give the impression that `freeze` can be used on arbitrary expressions.

In addition, it seems weird to have a keyword for `freeze` if we don't parse it as a separate syntactic construct. If the argument is that the concrete syntax should be close to the AST, then having `freeze` as a keyword seems further from the AST than using some lightweight symbol attached to the variable. Note that we do have special symbols for variable syntax already: `#` is used for `#var`, `#con`, et.c.

elegios:
I would have agreed with `freeze`, but I think @aathn has a good point here, `freeze x` suggests an AST with a node representing `freeze` with a child representing `a` (this is true both for functions and constructors, which are the other things that use this overloaded syntax), but in actuality the AST is just one node, a variable with an extra annotation (the `freeze` field is `true` rather than `false`).

Besides that I'm quite ok with not having the macros, at least not initially.

stdlib/mexpr/type-check.mc
aathn: For the record, I went ahead and changed `instBase`, `genBase` and `checkBeforeUnify` (previously `occurs`) to use `smapAccumL` and its relatives, which significantly simplified the code. The other change I mentioned above I'm not so sure about anymore; I'll leave it be for now.

stdlib/mexpr/type-check.mc
aathn: I implemented the approach above, but found a slight complication related to type variables. For instance, take the following snippet of code:
```
mexpr
lam x. let y = x in y
```
Typechecking the term and printing the result now produces the following:
```
lam x: _a.
  let y: _a =
    x
  in
  y
```
This cannot be parsed to produce the same term we started with, since it contains monomorphic type variables, for which there is no syntax. In addition, for the program
```
mexpr
let f = lam x. x in f
```
we get the result
```
let f: all a. (a) -> (a) =
  lam x: _a.
    x
in
f
```
where the parameter `x` is annotated with `_a` rather than `a`. I'm not quite sure what the best way of handling this is.

aathn:
On a different note, but related to the syntax of types: do we want a `#tyvar"..."` syntax for type variables? Right now they use the same `#var"..."` tag as ordinary variables.

stdlib/mexpr/type-check.mc
aathn: I have reverted this change now as discussed in the meeting. Instead, we probably want a separate field for the type of the let-bound variable.

PR #478, opened 2021-10-03
Restructure lambda lifting

larshum:
The lambda lifting implementation previously defined functions for constructing a call graph. These functions were however also used in another file (`stdlib/pmexpr/extract.mc`), and I later found a bug in the latter file which was caused by the call graph implementation. So in this PR the call graph functions are moved to a separate file, and the bugs are fixed.

In addition, I found a performance bug when adding the vertices to the call graph, which resulted in adding unnecessary amounts of vertices (the graph included the `inexpr` of a recursive binding). After this change, along with the latest performance improvements to the `digraph` library, lambda lifting on the `src/main/mi.mc` AST takes ≈2s on my computer (which is slightly faster than the boot parser on the same file).

### DIFF
--- a/stdlib/mexpr/call-graph.mc
+++ b/stdlib/mexpr/call-graph.mc
@@ -1,0 +1,109 @@
+-- Defines a language fragment for constructing a call-graph of the bindings in
+-- a recursive let-expression. The call graph contains an edge from a to b if
+-- the body of binding a contains a call to binding b.
+
+include "digraph.mc"
+include "mexpr/ast.mc"
+include "mexpr/ast-builder.mc"
+include "mexpr/symbolize.mc"
+include "mexpr/type-annot.mc"
+
+lang MExprCallGraph = MExprAst
+  sem constructCallGraph =
+  | TmRecLets t ->
+    let g : Digraph Name Int = digraphEmpty nameCmp eqi in
+    let g = _addGraphVertices g (TmRecLets {t with inexpr = unit_}) in
+    _addGraphCallEdges g t.bindings
+  | t ->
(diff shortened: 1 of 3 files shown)


stdlib/mexpr/call-graph.mc
```diff
@@ -0,0 +1,109 @@
...
+include "digraph.mc"
+include "mexpr/ast.mc"
+include "mexpr/ast-builder.mc"
+include "mexpr/symbolize.mc"
+include "mexpr/type-annot.mc"
+
+lang MExprCallGraph = MExprAst
+  sem constructCallGraph =
+  | TmRecLets t ->
+    let g : Digraph Name Int = digraphEmpty nameCmp eqi in
```
lingmar, line 14: What is the meaning of the `Int` labels? Unique identifiers of the labels, I guess? And why does it have to start with a `TmRecLets` node? I also have a call graph generation in my transformations, but the type of that one is `Digraph (Name, Info) (Name, Info)` (name and info of functions and call sites). I'm wondering if it's possible for us to use the same code, but I haven't looked too deeply if our use-cases are too different.

larshum: The labels are not used, I just set the label to `0` for all edges, so they might as well have type `()`. My approach should still work with the type you used.

I implemented it for `TmRecLets` because I only need call graphs for recursive let-expressions (to find an order in which to traverse all vertices, i.e. the bindings and their local functions). If all code except that in `constructCallGraph` is compatible with your approach, then it should work, because it could just be renamed and rewritten to call a more general call graph function.

PR #481, opened 2021-10-13
Adding a basic async library

david-broman:
This PR adds a small library for simple asynchronous programming using cooperative scheduling. The main primitives are `asyncBind`, `asyncRun`, `asyncSleep`, `asyncPrint`, and `asyncReturn`. The library is an external library that requires the `lwt` OPAM package to work. 

The library is available under `stdlib/ext/async-ext.mc`. There is also a simple test program available here: `test/examples/async/tic.mc`.


stdlib/ext/async-ext.ext-ocaml.mc
```diff
@@ -0,0 +1,45 @@
+
+include "ocaml/ast.mc"
+
+
+let asyncExtMap =
+  use OCamlTypeAst in
+  mapFromSeq cmpString
+  [
+    ("asyncSleep", [
+      { ident = "Lwt_unix.sleep",
```
br4sco: The field `ident` was renamed to `expr` in a previous PR

stdlib/ext/async-ext.mc
```diff
@@ -0,0 +1,27 @@
+
+
+
+
+
+type Promise
+
+external asyncSleep ! : Float -> Promise a
```
lingmar: It would be nice if naming reveals that the unit is seconds (and not ms, for example). What about `asyncSleepSec`?

stdlib/ext/async-ext.ext-ocaml.mc
david-broman: Thanks. Fixed.

stdlib/ext/async-ext.mc
david-broman: Good proposal. Fixed.

PR #483, opened 2021-10-18
Toml parser

lingmar:
This PR adds a TOML parser under `stdlib/ext/toml-ext.mc`. It uses the [ocaml-toml](https://github.com/ocaml-toml) parser via externals.

### DIFF
--- a/src/boot/lib/intrinsics.ml
+++ b/src/boot/lib/intrinsics.ml
@@ -535,6 +535,7 @@ module IO = struct
       else if Obj.tag v = Obj.double_tag then
         string_of_float (Obj.obj v) ^ "\n"
       else if Obj.tag v = Obj.closure_tag then "<closure>\n"
+      else if Obj.tag v = Obj.unaligned_tag then "<unaligned>\n"
       else
         let istr = String.make indent ' ' in
         let children =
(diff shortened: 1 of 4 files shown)


dlunde:
Cool, is this currently typed or untyped? It would be great to have the same thing for JSON as well.

lingmar:
> Cool, is this currently typed or untyped? It would be great to have the same thing for JSON as well.

You cannot see from the type `TomlValue` what the underlying type of the value is. E.g., the value in `key=1` must be converted with `tomlValueToIntExn : TomlValue -> Int` because `1` is an `Int`, and if you use the wrong  function you get a runtime error.

So untyped, depending on what you mean by typed/untyped in this setting.

dlunde:
Ah, I see. That makes sense. This should also mean that we can replace the TOML parser in miking-benchmarks with this (and no longer need to run that on boot due to the Python dep)? Maybe that was the main motivation for this PR.

lingmar:
Yes, I have started to replace the Python one in miking-benchmarks (it’s a bit annoying that the Python one can only use boot and this one needs to be compiled, so I have to replace everything at once). There’ll be a PR when I’m done.

PR #484, opened 2021-10-19
Syntax DSL tool skeleton + documentation

elegios:
This PR adds the syntax dsl tool in a first version. Currently this means:

- It can parse `.syn` files (and there is documentation and an example of the format). Note that only a subset is parsed at the moment, I'll cover the entire format once I can bootstrap the `.syn` parser, since it's a lot of boilerplate.
- It can generate language fragments for the AST, including `smap` functions. Presently the only thing missing here is generating accessor functions for the `info` field.
- I have not yet made sure that it compiles (interpreting it works though), since most of it was written before #558 was merged, and that was needed for this.

There are some questions around the interface for this:
-  Should it be a new subcommand of `mi` or its own thing?
-  What things should be configurable, and where?
- Should the output be one file or two? Conceptually it could be split into AST definition and a parser that produces an AST. It's probably also relevant that the parser probably won't typecheck with the new typechecker since it uses GADTs, but the AST definition shouldn't have any issues with that (once it outputs `all`s properly, which isn't there right now).

---
The previous description of this PR, from when it only contained documentation, can be found below:

This PR contains the WIP documentation for the syntax specification DSL that we intend to develop, that will be based on the `breakable` idea. It can be found in rendered form [here](https://github.com/miking-lang/miking/blob/2ad5fcb1c9d96e148398c72343c84c72f7c451f6/stdlib/parser/dsl.md).

Edit: updated version is rendered [here](https://github.com/miking-lang/miking/blob/4d57c6802dbf498a46808e22c81aa02d58de5fed/stdlib/parser/dsl.md).

Edit 2: new version [here](https://github.com/miking-lang/miking/blob/ac43033185a9363b2ff237375048371e64420d29/stdlib/parser/dsl.md)

### DIFF
--- a/stdlib/parser/dsl.md
+++ b/stdlib/parser/dsl.md
@@ -3,7 +3,7 @@
 This file gives a basic runthrough of the in-progress design of a DSL
 for syntax and AST specification.
 
-## Basics
+## Basic syntax and AST structure
 
 ```
 -- Line comments
@@ -13,8 +13,8 @@ Multiline comments
 -/
 ```
 
-We declare new types in the AST and non-terminals in
-the grammar with `type`.
+We declare new types in the AST and non-terminals in the grammar with
+`type`:
 
 ```
 type Expr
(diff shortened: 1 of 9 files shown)


stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+
+The syntax of a single production can contain regex style repetition
+operators:
+
+```
+-- Repeated field names are collected into a sequence in the
+-- order they appear in the source.
+prod TmTuple: Expr = "(" fields:Expr ("," fields:Expr)+ ")"
+-- Zero-or-more is expressed with `?` and produce an Option
+prod TmLet: Expr = "let" ident:LIdent (":" tyIdent:Type)? "=" value:Expr "in" body:Expr
```
david-broman: What will the generated abstract syntax be for this example? You would like to have one list with name "fields". Is that the case for the TmTuple example? As you did before, it would be nice to show what abstract syntax is generated.

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+```
+lang TmMatch
+  syn Expr =
+  | TmMatch { target : Expr, arms : [{pat : Pat, body : Expr}]}
+end
+```
+
+Explicit grouping (i.e. parentheses) should not be specified as a
```
david-broman: I think that you need to explain the whole concept of grouping in more detail, with more examples.

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+It should have the form `grouping` `<token>` `<type>` `<token>`.
+
+### Tokens
+
+Tokens can be either a literal (e.g. `"let"`) or a named token
+(e.g. `LIdent`). In the current system there is a number of builtin
+token types, currently specified in `lexer.mc`.
+
```
david-broman, line 145: Great. Can you list them here, in a bullet list?

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+precedence {
+  mul div;
+  add sub;
+  equal notEqual less greater;
+} except {
+  equal notEqual less greater;
+}
+```
+
+Each line in the `except` block makes precedence undefined between
```
david-broman: I do not understand this example. Sorry that the feedback is not more precise. Maybe easier that we talk about it.

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+
+A constructor in the AST and production in the grammar is introduced
+with `prod`.
+
+```
+prod TmInt: Expr = value:Int
+```
+
+Each constructor gets its own language fragment in the generated code,
+thus the above yields the following fragment:
```
david-broman, line 47: Maybe we optionally should make it possible to also just generate one language fragment. It will definitely be more readable.

david-broman (commented):
Looks really good. Please see my comments.

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+```
+infix TmMul: Expr = "*"
+
+-- Similarly for prefix and postfix
+prefix TmNot: Expr = "!"
+postfix TmFieldAccess: Expr = "." field:LIdent
+```
+
+Each production will be one of these things:
+- Simple/atomic if it has no direct recursion on either edge
```
lingmar: What is the difference between a simple and atomic production?

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+- Prefix if it only has direct recursion on the right edge
```
lingmar: For clarity it would be good to define what you mean by "edge"

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+- Postfix if it only has direct recursion on the left edge
+- Infix if it has direct recursion on both edges
+
+I suspect that productions that sometimes have direct recursion on the
+edge and sometimes not are a bad idea. Please let me know if such
+cases come up, so we can see if that is correct or not.
```
lingmar: I don't understand how this can happen. Do you have an example of such a production?

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+Infix productions can specify their associativity:
+
+```
+prod left TmSub: Expr = left:Expr "-" right:Expr
+infix left TmDiv: Expr = "/"
+```
+
+## Precedence
+
+Precedence is primarily specified with a precedence table:
```
lingmar: Primarily? Is there another way?

stdlib/parser/dsl.md
```diff
@@ -0,0 +1,197 @@
...
+  equal notEqual less greater;
+}
+```
+
+Each line in the `except` block makes precedence undefined between
+each possible pair of operators on the line. In this case, e.g.,
+`equal` and `notEqual` do not have defined relative precedence, nor
+does `notEqual` and `greater`.
+
+## Broken Productions
```
lingmar: I couldn't follow the "Broken Productions" section. I didn't understand if this was something that just happens "under the hood" or something that affects how you use the tool.

lingmar (commented):
I already gave most of my comments in the trellis file, but here are a few that were simpler to ask directly in the PR.

stdlib/parser/dsl.md
elegios: None, I'm just not sure what to call it. From the precedence point of view such a production is atomic, we can't split it into smaller parts (which we could for prefix/infix/postfix). However, it can still be made up of smaller parts, e.g., `"if" c:Expr "then" t:Expr "else" e:Expr "end"`, whereby "atomic" might be a bad name when it's presented to the user, since it's not indivisible to them.

Better suggestions would be quite welcome.

stdlib/parser/dsl.md
elegios: This production sometimes has a right-recursion, sometimes not:

```
prod bad: Expr = "bad" (e:Expr)?
```

stdlib/parser/dsl.md
elegios: Originally that was the thought, but I'm not so sure it's needed anymore, so I didn't add the other ways, but I forgot to change this sentence.

I'll take a pass at the precedence section in general and then we'll see if any other way to do it seems necessary.

stdlib/parser/dsl.md
elegios: The idea is that it should be something that:

1. Is only needed for a small subset of the productions used.
2. When it is needed, it is typically/always done automatically, entirely under the hood.
3. If that doesn't seem viable, then we'd need to provide a way to do it manually.

In general, the section is mostly there because the concept will have a place in the finished DSL, but I'm not quite sure of the full extent of it yet. At present all it means is that if a production has a regex repetition then this stuff will likely come into play, but for the moment it'll just be me figuring out what should be done, and if it can be done automatically, so it's just a placeholder/FYI.

PR #490, opened 2021-10-26
Type checker improvements

aathn:
This PR improves the type checker in several ways.

- Update the freeze syntax as discussed
- Add type checking for more terms:
  + Recursive lets
  + Pattern matching
  + Tensors
- Add correct types for primitive operations like `cons` (using type vars instead of `TyUnknown`)
- More tests
- Minor code improvements and bug fixes

Features still missing:
- Type checking of records and algebraic data types
- Improved error handling
- Value restriction

EDIT: I have also added a `--typecheck` flag to `mi` which can be used to run the typechecker on a program before evaluation / compilation. For instance, if `my_file.mc` contains
```
mexpr
let f = lam x. x in
[f true, 10]
```
then the output of `mi --typecheck my_file.mc` is
```
FILE "/path/to/miking/my_file.mc" 3:3-3:7 ERROR: Type check failed: unification failure
LHS: Bool
RHS: Int
```
Note that most practical programs still cannot be typechecked since unit and tuples are unsupported.

### DIFF
--- a/src/boot/lib/lexer.mll
+++ b/src/boot/lib/lexer.mll
@@ -75,7 +75,6 @@ let reserved_strings = [
   ("!",             fun(i) -> Parser.NOT{i=i;v=()});
   ("_",             fun(i) -> Parser.UNDERSCORE{i=i;v=()});
   ("->",            fun(i) -> Parser.ARROW{i=i;v=()});
-  ("`",             fun(i) -> Parser.BACKTICK{i=i;v=()});
 ]
 
 (* Info handling *)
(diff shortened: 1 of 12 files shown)


david-broman:
Great Anders! Would it be possible to add the type checker to `mi.mc` now, so that the type checker is enabled optionally for now with a flag `--typecheck`, but is not used by default? It would be good for testing purposes.

aathn:
Good idea! I have now added a `--typecheck` flag.

For now, it simply runs `typeCheck` on the AST before evaluating / compiling the program (so it does not for instance print the inferred types).

PR #494, opened 2021-11-01
Add `--output` option and use argument parser library for options

lingmar:
This PR
* Uses the argument parser library for parsing command line options (in `options.mc`). Sub commands (`compile`, `eval`, etc) are still parsed as before (not using the argument parser), as it didn't seem to fit within the argument parser framework.
* Adds the option `--output` for specifying destination file when compiling.


src/main/mi.mc
```diff
@@ -79,18 +75,19 @@ if lti (length argv) 2 then usage () else
...
   else
-    -- No, not a well known command
-    -- Parse options as far as possible. Does user require help?
-    let split = splitOptionPrefix (tail argv) in
-    let options = parseOptions split.first in
-    maybePrintHelp options;
+    -- No, not a well known command.
+    -- Does user require help?
+    let wantHelp = optionIsSome (find (eqString "--help") (tail argv)) in
+    maybePrintHelp {options with printHelp = wantHelp};
```
aathn: I guess this makes it impossible to pass `--help` to the evaluated program when using the short syntax. Is that preferable to what we had before?

lingmar: I believe that's what we had before as well?

lingmar: Or wait, you could pass it after the filename before: `mi foo.mc --help`. I should add a call here to `splitOptionPrefix` to get that behaviour, I guess.

PR #496, opened 2021-11-03
First version of 0-CFA framework for MExpr

dlunde:
## Major changes
- Add a modular and extensible (hopefully) 0-CFA framework to Miking, based on chapter 3 in Nielson et al. ("Principles of Program Analysis"). Example: for the simple program `(lam x. x) (lam y. y)`, the analysis currently generates the following:
   ```
   --- ORIGINAL PROGRAM ---
   (lam x.
      x)
     (lam y.
        y)
   
   --- ANF ---
   let t =
     lam x.
       x
   in
   let t1 =
     lam y.
       y
   in
   let t2 =
     t
       t1
   in
   t2
   
   --- FINAL CFA GRAPH ---
   *** WORKLIST ***
     []
   *** DATA ***
     t =
       lam x
     t1 =
       lam y
     t2 =
       lam y
     x =
       lam y
   *** EDGES ***
     t =
       {lam x} ⊆ t ⇒ t1 ⊆ x
       {lam x} ⊆ t ⇒ x ⊆ t2
       {lam y} ⊆ t ⇒ t1 ⊆ y
       {lam y} ⊆ t ⇒ y ⊆ t2
     t1 =
       {lam x} ⊆ t ⇒ t1 ⊆ x
       {lam y} ⊆ t ⇒ t1 ⊆ y
     x =
       {lam x} ⊆ t ⇒ x ⊆ t2
     y =
       {lam y} ⊆ t ⇒ y ⊆ t2
   ```
   Data-flow components are not included, and should instead be added separately for the various analyses that we implement. Finally, some parts of MExpr are not yet supported. Most importantly, analysis of control flow within/through records and variants is still missing. 
- *Needs to be discussed*: Add support for the syntactic sugar `match t1 with p in t2` for `match t1 with p then t2 else never`.
- ANF updates
  - Propagate term info in `bind`.
  - Add default case for `isValue`.
  - Lift and bind lambdas to names (they were previously not lifted). Sequences of lambdas are still only bound to one name. That is, we translate `(lam x. (lam y. (lam z. b)))` to `let t = (lam x. (lam y. (lam z. b))) in t`, and *not* `
  let t1 = (lam x. let t2 = (lam y. let t3 = (lam z. b) in t3) in t2) in t1`. This is similar to sequences of applications, for which each intermediate closure is *not* assigned a name.
  - ANF transformation of recursive lets now halts on encountering non-lambda bodies.

## Minor changes
- Add note for installing `owl` to README.
- Add `mapAllWithKey` to `stdlib/map.mc`
- Fix typo in `type-annot.mc` error message.
- Generalize the first argument of `eqSeq` from `a -> a -> Bool` to `a -> b -> Bool`.
- Add `setSubset` to `set.mc`.

### DIFF
--- a/stdlib/seq.mc
+++ b/stdlib/seq.mc
@@ -13,7 +13,7 @@ let init = lam seq. subsequence seq 0 (subi (length seq) 1)
 utest init [2,3,5] with [2,3]
 utest last [2,4,8] with 8
 
-let eqSeq = lam eq : (a -> a -> Bool). lam s1. lam s2.
+let eqSeq = lam eq : (a -> b -> Bool). lam s1. lam s2.
   recursive let work = lam s1. lam s2.
     match (s1, s2) with ([h1] ++ t1, [h2] ++ t2) then
       if eq h1 h2 then work t1 t2
--- a/README.md
+++ b/README.md
@@ -1484,6 +1484,9 @@ output of the above program might be:
 However, the values and order of the thread IDs might be different over
 different runs.
 
+### Probability distributions
+Externals for probability distributions are defined in `stdlib/ext/dist-ext.mc`. To use these, you must install the `opam` package `owl` (i.e., `opam install owl`)
+
 ## Profiling
 
 Profiling of an MExpr program can be enabled using the `--debug-profile` flag
--- a/src/boot/lib/parser.mly
+++ b/src/boot/lib/parser.mly
@@ -322,6 +322,9 @@ mexpr:
   | MATCH mexpr WITH pat THEN mexpr ELSE mexpr
       { let fi = mkinfo $1.i (tm_info $8) in
         TmMatch(fi,$2,$4,$6,$8) }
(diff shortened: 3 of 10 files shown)


stdlib/seq.mc
```diff
@@ -13,7 +13,7 @@ let init = lam seq. subsequence seq 0 (subi (length seq) 1)
 utest init [2,3,5] with [2,3]
 utest last [2,4,8] with 8
 
-let eqSeq = lam eq : (a -> a -> Bool). lam s1. lam s2.
+let eqSeq = lam eq : (a -> b -> Bool). lam s1. lam s2.
```
br4sco, line 16: The type should be `a -> a -> Bool` since `eqSet` is constrained by `eqMap` which requires maps with the same compare function (which constrains `eqSet` beyond the type on the equality function). Might be worth documenting this if you don't mind?

PR #504, opened 2021-11-12
Remove dependency on `timeout` command for tune tests

lingmar:
The tune tests depended on the command `timeout`, which caused the tests to fail if that command didn't exist.


lingmar:
Subsumed by #507

PR #509, opened 2021-11-18
Fix handling of type aliases in type checker

aathn:
This PR makes the typechecker capable of resolving type aliases properly.

For instance, now the following simple program typechecks, whereas previously it was unable to unify `MyType` with `r<:{x : Int}`.

```ocaml
mexpr

type MyType = {x : Int} in

let f =
  lam r : MyType. {r with x = 10}
in

f {x = 5}
```

Note that it still cannot handle aliases with parameters.


david-broman:
Fantastic! It works for me in my test cases. I will merge.

PR #510, opened 2021-11-18
Add support for record polymorphism in the type checker

aathn:
This PR extends the typechecker to handle polymorphism in records. Specifically, it adds bounded polymorphism for record types, so that a function like
```ocaml
let mapPair =
  lam f. lam r.
  match r with (x, y) in (x, f y)
```
is assigned the type `all a. all b. all c. all r<:(c,a). (a -> b) -> r -> (c,b)`. This lets us express a subset of the things we would be able to with row polymorphism, and was relatively straightforward to add given the approach used to infer record types. Note that I have not yet added syntax for writing bounds like this in type annotations : they only exist in the type checker.

I know I promised to work on better error messages, but I couldn't resist adding this while I was at it. I will work on the error messages next though.

With this, the only thing remaining to be able to typecheck all of the current codebase should be handling type aliases with parameters.

EDIT: I have now updated this PR so that the record polymorphism is disabled by default but can be enabled using a parameter to the `typeCheck` function. When record polymorphism is disabled, the behaviour is as before, so that an unannotated record parameter to a function is assumed to be monomorphic. Note that using annotations it is still possible to write functions on polymorphic records (just not polymorphic over the fields of the record):
```ocaml
let mapPair : all a. all b. all c. (a -> b) -> (c, a) -> (c, b) =
  lam f. lam r.
  match r with (x, y) in (x, f y)
```
typechecks fine.

In addition, this PR improves the type checker code somewhat and fixes a bug.

### DIFF
--- a/stdlib/mexpr/ast-builder.mc
+++ b/stdlib/mexpr/ast-builder.mc
@@ -94,13 +94,15 @@ let tyvar_ =
   lam s.
   ntyvar_ (nameNoSym s)
 
-let ntyall_ = use AllTypeAst in
-  lam n. lam ty.
-  TyAll {ident = n, info = NoInfo (), ty = ty, sort = TypeVar ()}
+let nstyall_ = use AllTypeAst in
+  lam n. lam sort. lam ty.
+  TyAll {ident = n, info = NoInfo (), ty = ty, sort = sort}
 
-let tyall_ =
+let styall_ = lam s. nstyall_ (nameNoSym s)
+
+let tyall_ = use VarSortAst in
   lam s.
-  ntyall_ (nameNoSym s)
+  styall_ s (TypeVar ())
 
 let tyalls_ =
   lam strs. lam ty.
--- a/stdlib/mexpr/ast-builder.mc
+++ b/stdlib/mexpr/ast-builder.mc
@@ -96,7 +96,7 @@ let tyvar_ =
 
 let ntyall_ = use AllTypeAst in
   lam n. lam ty.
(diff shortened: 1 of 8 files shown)


dlunde:
> With this, the only thing remaining to be able to typecheck all of the current codebase should be handling type aliases with parameters.

I think we also need to add support for defining regular types within language fragments in some way. A lot of our current code (at least mine, but I think I'm not alone in this) looks as follows right now:
```
type Env = {
  a: A -- A is not defined here
}

lang L
  syn A =
  | ...

  sem computation (env: Env) =
  | ...
end
```
I think we have discussed this before, but I did not find an open issue for this.

elegios:
One thing to note here is that our backend doesn't support this style of polymorphism on records yet, so we can't quite replace type-annot with this as is, at least not without a few extra type annotations to make things a bit less polymorphic. How hard would it be to either not have this style of bounded polymorphism, or to have it as an option in some way?

aathn:
I think this could easily be made an option. The typechecker could check after inferring a type whether there are any undetermined or polymorphic records in the inferred type and if so raise an error asking the user to provide annotations for those types.

aathn:
I have now updated the PR to make the record polymorphism optional and disabled by default. When disabled, instead of raising an error for unannotated record functions as I suggested above, I simply revert to the old behavior so that record arguments are inferred to be monomorphic.

Thus, this PR now does not change the behavior of the type checker, but it lays the groundwork for handling record-polymorphic functions, improves the code, and fixes a bug.

PR #517, opened 2021-12-02
Make acceleration part of `mi compile` with a separate flag

larshum:
This PR:
* Makes acceleration controlled by a flag `--accelerate` of `mi compile` instead of a separate command.
* Adds support for using `utest`s of specific shapes to express size coercions and size equality constraints on the program. See the README for further details and examples.
* Upates the [documentation on acceleration](https://github.com/larshum/miking/blob/d4daec7d156a0afe033ace26236b59acb7b54ac4/README.md#acceleration) in the README.


README.md
```diff
@@ -1562,106 +1562,222 @@ using the evaluator.
...
+* `accelerate` - marks expressions that should be accelerated.
+* `parallelFlatten` - flattens a two-dimensional sequence, producing a
+  one-dimensional sequence. The term `parallelFlatten s` is equivalent to
+  `foldl concat [] s`.
+* `parallelMap` - equivalent to the `map` keyword.
+* `parallelMap2` - performs a `map` operation over two sequences that are
+  assumed to be of the same length.
+* `parallelReduce` - performs a reduction over elements of a sequence. If the
+  applied function is associative and the initial accumulator value is the
+  neutral element of this function (so that the applications can be reordered),
```
elegios: "[...] so that the applications can be reordered [...]" sounds a bit like using a commutative property instead of an associative one, is that intended?

README.md
```diff
@@ -1562,106 +1562,222 @@ using the evaluator.
...
+elaborate examples.
+
+### Sequence sizes
+
+A significant difference between MExpr and Futhark is that the latter includes
+the size of a sequence as part of its type. This means that, for example, when
+a function returns a two-dimensional sequence, the inner sequences must all
+have the same size (otherwise they would have different types). As this is not
+required by MExpr, there are situations where one may need to provide
+additional information in PMExpr to help the Futhark compiler.
```
elegios: PMExpr -> MExpr?

README.md
larshum: Yes, this is intentional.

If `f` is an associative function, we can reorder the expression `f (f a b) c` as `f a (f b c)`. We also assume that the expression `0` in `parallelReduce f 0 s` is the neutral element of `f`, i.e. an element such that `f 0 b = f b 0 = b` for all `b`. By these assumptions, we can apply `f` on the elements of `s` in any order and get the same result. For example,
```
f (f (f 0 s[0]) s[1]) s[2] = f (f 0 s[0]) (f (f 0 s[1]) s[2])
```

README.md
larshum: Ah yes, I never introduced the MExpr with parallelization as PMExpr here, so I suppose that makes more sense.

README.md
elegios: But it's not quite in any order though, if the original order is `f (f 0 s[0]) s[1]`, then `f (f 0 s[1]) s[0]` is a reorder of the applications, but it's not ok, right? We must maintain the order, but we're free to change the grouping. I feel like this tends to be phrased differently somehow, but I'm not coming up with/remembering how.

larshum: Ah I see what you mean, good point. No, that would only be okay if `f` is both associative and commutative.

larshum: Perhaps it would be best to just remove that part?

elegios: Personally I would find it sufficient to just state the assumptions and that it uses those assumptions to run the computations in parallel, I don't think there's a need to mention more of how it does it.

PR #518, opened 2021-12-06
Add graph equality to digraph library

gizemcaylak:
This PR adds equality comparison between two graphs having the same vertex and label comparisons


stdlib/digraph.mc
```diff
@@ -106,6 +106,18 @@ let digraphHasEdge = lam e. lam g.
...
   forAll (lam e. digraphHasEdge e g) es
 
+-- Check whether graph g1 is equal to graph g2.
+let digraphGraphEq =
+  lam g1: Digraph v l. lam g2: Digraph v l.
+    if eqi (digraphCountEdges g1) (digraphCountEdges g2) then
+      if eqi (digraphCountVertices g1) (digraphCountVertices g2) then
+        mapEq (lam ds1. lam ds2.
+          let zipped = zipWith (lam d1. lam d2. (d1, d2)) ds1 ds2 in
+          forAll (lam d:((v,l),(v,l)). and (digraphEqv g1 (d.0).0 (d.1).0) (digraphEql g1 (d.0).1 (d.1).1)) zipped
```
lingmar: This check takes the order of the edges into account, so it will fail if you have two identical graphs where you added the edges in different orders. I think it will work if you instead use `eqSeq` on the `ds1` and `ds2` sequences.

PR #520, opened 2021-12-07
Wrap relevant intrinsics in try-with expressions

larshum:
This PR wraps relevant intrinsics in the `boot` library in try-with expressions which provide a more helpful error message in case an exception is raised by the underlying `Rope` operation. The error message states that one should use the `--runtime-checks` flag for better error messages. This flag is not enabled by default because it has a non-negligible impact on performance, unlike the try-with expressions added here.

### DIFF
--- a/src/boot/lib/intrinsics.ml
+++ b/src/boot/lib/intrinsics.ml
@@ -3,7 +3,7 @@ open Ustring.Op
 module Mseq = struct
   type 'a t = List of 'a List.t | Rope of 'a Rope.t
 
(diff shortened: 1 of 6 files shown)


lingmar:
Would it make sense to always catch any exception from OCaml, not just out-of-bounds from `Rope`? Maybe it's complicated, but just a thought.

larshum:
I think we did that previously, by wrapping the entire OCaml AST (except the type declarations) in a try-with expression, so it should be doable.

PR #523, opened 2021-12-15
Combine dependency analysis with context analysis for holes

lingmar:
This PR extends the dependency analysis for holes by combining it with the existing context analysis for holes. The analysis now considers base holes with different contexts as two different data. 

It is based on #519 ([diff](https://github.com/miking-lang/miking/pull/523/files/75c91ccc75f33a74e2929e96554353a62a323c04..8e805f76ab678013586e39412006d4bee9fd7a73)).

Changes made to `cfa.mc` are as follows (great if @dlunde would like to review):
- Add a field `graphData` in the CFA graph. One can store custom data in it by extending the `syn` called `GraphData` and pattern matching on it when using it.
- Change the signature of the semantic function `initGraph` so that it takes the `graphData` as argument. This is because I need this data when generating the constraints.
- A semantic function `isDirect : AbsVal -> Bool` for enabling/disabling direct constraints on abstract values (default is `true`).
- A semantic function `directTransition : CFAGraph -> Name -> AbsVal -> AbsVal` which describes how an abstract value should be updated when transferred to a name (default identity function). This and the previous function could possibly be merged into one.

Other changes in `stdlib`:
- Adds a `setIntersect` function to `set.mc`


stdlib/mexpr/cfa.mc
```diff
@@ -260,11 +270,21 @@ lang DirectConstraint = CFA
...
   sem initConstraint (graph: CFAGraph) =
   | CstrDirect r & cstr -> initConstraintName r.lhs graph cstr
 
   sem propagateConstraint (update: (Name,AbsVal)) (graph: CFAGraph) =
-  | CstrDirect r -> addData graph update.1 r.rhs
+  | CstrDirect r ->
+    if isDirect update.1 then
+      let av = directTransition graph r.rhs update.1 in
+      addData graph av r.rhs
+    else graph
```
dlunde: I may be wrong, but I believe you need to do an equivalent change to this part of `cfa.mc`:
```
lang NamedPatCFA = MatchCFA + NamedPat
  sem propagateMatchConstraint (graph: CFAGraph) (id: Name) =
  | (PatNamed { ident = PName n }, av) -> addData graph av n

```

lingmar: Yes, you're right, thanks! I put the logic in a separate language fragment `DirectConstraintOptions` that is now used both by `DirectConstraint` and `NamedPatCFA`.

PR #528, opened 2022-01-10
Topological sort, efficient graph comparison in digraph

gizemcaylak:
This PR adds:

- Topological sort function on a DAG
- Update of the graph equivalence comparison function to a more efficient form using maps
- `digraphMaybeAddEdges` [and `digraphMaybeAddEdge`] function which does not change the graph if an edge exists.
- `digraphAddUpdateVertex` function which updates a vertex v in graph g if v already exists in g, otherwise add the vertex to the graph.


stdlib/digraph.mc
```diff
@@ -153,6 +160,15 @@ let digraphAddVertex = lam v. lam g.
 let digraphMaybeAddVertex = lam v. lam g.
   digraphAddVertexCheck v g false
 
+-- Add a vertex v to g. Updates the vertex v in graph g if v already exists in g.
+let digraphAddUpdateVertex = lam v. lam g : Digraph v l.
+  if digraphHasVertex v g then
+    let edgeList = mapLookupOrElse (lam. error "Lookup failed") v g.adj in
+    let m = mapRemove v g.adj in
+    {g with adj = mapInsert v edgeList m}
```
lingmar: Maybe I'm missing something, but wouldn't it be simpler with just:
```suggestion
  if digraphHasVertex v g then g
```

stdlib/digraph.mc
```diff
@@ -283,6 +312,27 @@ lam g.
...
 
+-- Kahn's algorithm: Topological sorting of large networks
+-- https://dl.acm.org/doi/10.1145/368996.369025
+-- Returns a list of sorted vertices
+let digraphTopologicalOrder:Digraph v l -> [v] = lam g.
+  let indegrees = foldl (lam acc. lam v. mapInsert v (length (digraphEdgesTo v g)) acc) (mapEmpty (digraphCmpv g)) (digraphVertices g) in
+  let rootnodes = filter (lam v. eqi (length (digraphEdgesTo v g)) 0) (digraphVertices g) in
+  recursive
+    let order = lam ordering. lam indegrees. lam rootnodes.
+      if eqi (length (rootnodes)) 0 then ordering
```
lingmar:
```suggestion
      if null rootnodes then ordering
```
(Since `null` might be more efficient than `length`)

stdlib/digraph.mc
gizemcaylak: If the vertex v exists, I still want to update the v in the graph (maybe same ident [lets say comps looking to ident] but non-compared parts within v might be updated). Is the motivation clear now?

lingmar: Ah, I see!

PR #533, opened 2022-01-18
Type alias definitions in language fragments

larshum:
This PR adds support for defining type aliases inside language fragments, so that a type alias can be properly defined in terms of a type defined within a language fragment. For example, the `RecLetBinding` can now be defined as (this change is included in the PR) a type alias within the `RecLetsAst` language fragment in `mexpr/ast.mc` as follows
```
lang RecLetsAst = Ast + VarAst
  type RecLetBinding = {
    ident : Name,
    tyBody : Type,
    body : Expr,
    info : Info}
  ...
end
```
and this results in a type alias `RecLetsAst_RecLetBinding` being defined after all the types it depends on have been defined, but prior to the definition of any constructors that depend on it (if my understanding of the MLang code is correct).

Note that he changes to the parser results in a shift-reduce conflict because the `end` keyword of a language fragment is optional. This means that, for example, a language fragment that only defines one type alias could also be interpreted as a language fragment followed by a type definition. I can think of two options to fix this (I haven't implemented either yet):
* Require that an `end` token is used to terminate a language fragment.
* Use a different keyword for type aliases.


Makefile
```diff
@@ -85,6 +85,7 @@ uninstall:
 test-all:\
   test-boot-all\
   test-compile\
+	test-compile-type-checked\
```
elegios: strange indentation

src/boot/lib/mlang.ml
```diff
@@ -215,7 +233,24 @@ let merge_lang_data fi {inters= i1; syns= s1} {inters= i2; syns= s2} :
...
+        None
+    | None, Some a ->
+        Some a
+    | Some a, None ->
+        Some a
+    | Some (fi1, ty1), Some (fi2, _) ->
+        if eq_info fi1 fi2 then Some (fi1, ty1)
+        else
+          raise_error fi1
+            ( "Type alias '" ^ Ustring.to_utf8 name ^ "' was redefined at "
```
elegios: Might be better with something like "conflicting definitions", because this is the consequence of merging two fragments that both define an alias with the same name, right?

src/boot/lib/mlang.ml
```diff
@@ -718,14 +793,16 @@ let desugar_top (nss, langs, subs, syns, (stack : (tm -> tm) list)) = function
         (* TODO(vipa,?): this does not declare the type itself *)
         | Data (_, name, cdecls) ->
             List.fold_right (wrap_con name) cdecls tm
+        | Alias (fi, name, ty) ->
```
elegios, line 797: How do we know that this gives an appropriate order? (`type`s first, then `type` aliases, then `con`s)

test/mlang/type-alias.mc
```diff
@@ -0,0 +1,70 @@
...
+--
+-- Type-aliases defined within language fragments.
+
+lang L1
+  syn T1 =
+
+  type A1 = {a1 : T1, a2 : T2}
+  type A2 = {b1 : A1, b2 : Int}
+
+  syn T2 =
```
elegios, line 12: I believe we need some permutations of `syn`s using `type` aliases as well, all these tests use `syn`s in `type`, but not vice-versa.

elegios (changes requested):

Makefile
larshum: Fixed!

larshum: Weird, I'll fix.

src/boot/lib/mlang.ml
larshum: Good point, I'll update the message.

test/mlang/type-alias.mc
larshum: Ok, I'll add that as well.

src/boot/lib/mlang.ml
larshum: So in my understanding, the `type`s are always emitted at the top, and this is not part of `wrap_data`. Because `aliases` are placed before `syns` in `decls` when the `Lang` is created on line 302 as
```
  Lang (info, name, includes, aliases @ List.rev_append syns inters)
```
I think this should guarantee that they are always placed before the `cons`, because `wrap_data` is called from a `fold_right`. This initialization is probably inefficient, but if I did `List.rev_append` for `aliases` I think they ended up in the wrong order relative to each other.

test/mlang/type-alias.mc
elegios: I'd suggest also adding `syn`s using `type`s that are defined later, since that's allowed by the semantics we're aiming for. After that I think this looks good.

PR #536, opened 2022-01-27
Fix compiler performance issues

larshum:
This PR fixes a significant performance issue in the register allocation step of the OCaml compilation, by adding top-level generation of externals. I also sneaked in some refactoring.

Because the generation of top-level lets did not have a case for externals, all expressions following the first top-level external to be treated as part of the main expression instead. For the large files we generate, in the case an external was placed early on in the file, this resulted in a main expression covering hundreds of thousands of lines. As I assume the register allocator treats this entire expression as one function, that is probably what caused it to be this slow.

Thanks to @elegios for spotting this bug when we looked at some generated code during today's meeting.

### DIFF
--- a/stdlib/ocaml/generate.mc
+++ b/stdlib/ocaml/generate.mc
@@ -86,6 +86,49 @@ let lookupRecordFields = use MExprAst in
 type MatchRecord = {target : Expr, pat : Pat, thn : Expr,
                     els : Expr, ty : Type, info : Info}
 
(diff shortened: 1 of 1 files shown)


lingmar:
Nice! Does it also have an effect on boot strapping time?

larshum:
Yes, I forgot to mention this! The second and third steps are significantly faster now - on my machine they are faster now using the default register allocation than they are on `develop` when using linear scan (by setting the `--disable-optimizations). The first step should be unaffected by these changes though, and that seems to be the case.

PR #538, opened 2022-02-02
Improve performance of `mi eval`

larshum:
This PR improves the performance of `mi eval` by representing the environment using a `List` (from `createList`) instead of using a `Map` as before. It also collects all uses of the evaluation environment to a few functions to make it easier to update the environment representation in the future, should that be desireable.

I include the (currently unmerged) changes in #536 here because without them the bootstrapping times went bananas. This is also why I haven't made this PR until now.


larshum:
I forgot to provide any measurements of how much this change improves performance, so I measured the time it takes to run `<eval> src/main/mi-lite.mc -- 0 src/main/mi-lite.mc` (i.e. the first step of bootstrapping) on my computer, where <eval> is one of `boot eval`, `mi eval` (before) and `mi eval` (after):
* For `boot eval` this takes ≈22s.
* For `mi eval` before this PR, it takes ≈2m 45s.
* For `mi eval` after this PR, it takes ≈1m.

So this is a pretty significant improvement. We're however still a factor of 3 slower than `boot eval`, even though we are now using (more or less) the same underlying data structure for the evaluation environment.

PR #540, opened 2022-02-09
Add option to C compiler for toggling between 32- and 64-bit int/float

larshum:
This PR adds support for controlling whether the C compiler translates integers and/or floats to 32-bit or 64-bit values. This was discussed during a [recent meeting](https://github.com/miking-lang/miking/wiki/Miking-Meeting-Notes-2022#meeting-notes-2022-02-03).

I decided to use explicitly sized types for integers (`int32_t` and `int64_t`) so that their sizes are not platform dependent (the C standard only guarantees `int` is at least 16-bits, for example). It should be noted that these types are not required to be defined by the standard (https://www.cplusplus.com/reference/cstdint/). Perhaps using `int_least_32_t` and `int_least_64_t`, which must be defined but could be larger than the specified number of bits, would be a better option?


stdlib/c/compile.mc
```diff
@@ -148,7 +163,7 @@ lang MExprCCompile = MExprAst + CAst
   -- Intentionally left blank
 
   -- Entry point
-  sem compile (typeEnv: [(Name,Type)]) =
+  sem compile (typeEnv: [(Name,Type)]) (compileOptions : CompileCOptions) =
```
dlunde, line 166: The `(compileOptions : CompileCOptions)` addition is basically the only interface change, right?

dlunde (approved):
I've looked at the changes and think it all looks good. Thanks for taking care of my TODO as well. I'll create a compatibility update PR for `miking-dppl` early next week.

stdlib/c/compile.mc
larshum: Yes.

larshum: Or no, the `compileCEnvEmpty` function now also expects a `CompileCOptions` record as a parameter.

Issue #541, opened 2022-02-11
Ambiguous output from failed utests in included files


larshum:
This issue was identified by @gizemcaylak.

When a `utest` fails we only print the line at which the failed `utest` is defined in a file. This works fine when all `utest`s originate in the same file as we are running tests on. But currently, we don't remove all `utest`s from included files (I think we only remove those placed on top-level of the MLang, but I may be wrong here). This means that when an included `utest` fails, we don't know which of the included files contains this failing `utest` and we have to look this up manually. The problem applies to both `boot` and the bootstrapped evaluator and compiler.

I made a toy example, just to illustrate what the problem looks like. If we have a file `a.mc` defined as
```
include "string.mc"
let f = lam x.
  utest x with 0 in
  int2string x
```
and another file `b.mc` which includes `a.mc` as
```
include "a.mc"

utest 3 with 0
mexpr
let x = f 3 in
x
```
We get the following output when running `boot eval --test b.mc`
```
 ** Unit test FAILED on line 3 **
    LHS: 3
    RHS: 0

 ** Unit test FAILED on line 3 **
    LHS: 3
    RHS: 0
```

dlunde:
Isn't this simply a matter of adding the file name to the printout? I believe it is stored in the info type.

larshum:
Yeah, that's true. I'm just not sure whether we want to have included `utest`s at all, because even with the file name we still don't know where it was called from. Perhaps we should print a stack trace?

lingmar:
I wouldn't want to remove utests from included files. I often use utests as `assert` in C, and then I want them to hit when writing the proper tests, which could be in a different file.

To get a stack trace, I typically set an `error` by the `utest` that failed, turn on `--debug-stack-trace` and run via `boot`. It's not very elegant but it's been enough for my use-cases.

larshum:
OK, then I just think we should improve the error reporting so that one does not have to depend on `boot`.

larshum:
I'm closing this now, since it was fixed in the now merged PR #543.

PR #543, opened 2022-02-16
Add file path to failed utest output

larshum:
This PR updates the behaviour of failed utests in `boot` and `mi` so that the path to the file of the failed utest is printed, as discussed in #541. Prior to the changes in this PR, a failed utest would be reported as
```
 ** Unit test FAILED on line 252 **
```
while after this change it is reported as
```
 ** Unit test FAILED on line 252 of file <path>/<to>/<file>.mc **
```

### DIFF
--- a/src/boot/lib/mexpr.ml
+++ b/src/boot/lib/mexpr.ml
@@ -1652,7 +1652,7 @@ let shape_str = function
 let unittest_failed fi t1 t2 tusing =
   uprint_endline
     ( match fi with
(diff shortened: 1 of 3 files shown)


larshum:
Updated to use format `** Unit test FAILED: <info2str> **`. Also, I noticed that the printing of usings had not been added to `mi`, so I added that as well. It will now print `Using: <expr2str using>` if a using-expression is provided to a failing utest, just like in `boot`.

PR #545, opened 2022-02-20
Fix compilation when using `dune` version >= 3

larshum:
This PR fixes a compilation error in the generated dune project for recent versions of `dune` (3.0.0 and later), due to a change in its behaviour relative to earlier versions.

To solve this, I generated a minimal `dune-project` file in the temporary directory, as this is required in `dune` versions >= 3. The reason I used `2.0` in the `(lang dune 2.0)` stanza is because this is backwards-compatible, so it works independently of the installed `dune` version (nothing seemed to happen when I used `1.0`, but I don't think we want to support such old versions anyway).


larshum:
I should also mention that we cannot use `dune` version 3.0 yet because we currently use `ocamlformat` version `0.16.0`, and this does not work with a more recent version of `dune` than `2.9.3`.

PR #546, opened 2022-02-22
Use ocamlformat 0.20.1

lingmar:
Bumps the version of `ocamlformat` to `0.20.1`. After this PR will follow a PR that adds the commit hash of this hash to `.git-blame-ignore-revs`.


larshum:
The tests seem to be failing because the GitHub Actions configuration file still uses `ocamlformat` version `0.16.0`. Update it at `.github/workflows/main.yml` on line 36.

PR #551, opened 2022-03-03
Support for singleton tuple matching

johnwikman:
It is currently possible to define singleton tuples, but not to match on them in patterns. This adds support in the boot compiler for creating patterns on singleton tuples.

### DIFF
--- a/src/boot/lib/parser.mly
+++ b/src/boot/lib/parser.mly
@@ -486,6 +486,9 @@ pat_atom:
       { PatSeqEdge(mkinfo (fst $1) (fst $3), Mseq.empty, snd $1, $3 |> snd |> Mseq.Helpers.of_list) }
   | LPAREN pat RPAREN
       { $2 }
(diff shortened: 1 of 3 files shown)


johnwikman:
As discussed during the meeting on 2022-03-03, it would be interesting to see what happens in the pretty-printing of singleton tuples in MExpr. Constructed a test program for this purpose:

```
include "mexpr/ast-builder.mc"
include "mexpr/pprint.mc"

-- match (0,) with (_,) then 1 else 2
let tupmatch_ast =
  match_
    (utuple_ [int_ 0])
    (ptuple_ [pvarw_])
    (int_ 1)
    (int_ 2)

mexpr

use MExprPrettyPrint in
print (expr2str tupmatch_ast); print "\n"
```

This outputs the following:

```
match
  (0,)
with
  (_,)
then
  1
else
  2
```

So there is nothing to be fixed in the pretty printing. This outputs the same syntax as is now allowed by the boot parser.

PR #552, opened 2022-03-08
Fix bugs in lambda lifting

larshum:
This PR fixes a bug in lambda lifting where the inner lambdas of nested anonymous lambdas would not be lifted. For example, if we add one to each element of a two-dimensional sequence as
```
map (lam s2 : [Int]. map (lam x. addi x 1) s2) s1
```
then both anonymous lambdas should be lifted. But prior to this PR, only the outermost lambda would be lifted because parts of the lambda lifting did not recurse into the body of an anonymous lambda.

After I opened this PR, I found and fixed two additional bugs in the lambda lifting:
1. The bodies of let-expressions inside recursive bindings would not remove lifted lambdas inside its body, leading to the same identifier being defined in multiple places in the program.
2. Lambda lifting of a recursive binding within a let-expression would not continue performing lambda lifting of the `inexpr` of the recursive binding.

### DIFF
--- a/stdlib/mexpr/lamlift.mc
+++ b/stdlib/mexpr/lamlift.mc
@@ -40,8 +40,13 @@ lang LambdaLiftNameAnonymous = MExprAst
 
   sem nameAnonymousLambdas =
   | TmLam t ->
+    recursive let getLambdaBody = lam t : Expr.
+      match t with TmLam tt then getLambdaBody tt.body
+      else t
+    in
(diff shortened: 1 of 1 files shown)


larshum:
I have found issues when using the lambda lifting even after these fixes, so this PR is not ready in its current state.

larshum:
It's ready now, but I'll miss the meeting today.

PR #553, opened 2022-03-10
Add type parameter support and improve --typecheck

aathn:
This PR includes the following changes. 

- Add parsing of type parameters
- Add support for parametrized type aliases to the type checker
- Make `--typecheck` perform `typeCheck` instead of `typeAnnot` rather than running both
- Make `--typecheck` be strict when symbolizing type variables and type constructors
- Make utesttrans.mc and type-lift.mc compatible with type-check.mc
- Update type signatures in `option.mc`, `seq.mc` and `assoc.mc` to make them pass `--typecheck`

With this, the type checker should be ready for general use. There are still things to improve, such as checking for invalid type constructor applications (i.e. basic kind checking), implementing a value restriction, improving error messages, and testing it thoroughly so that there are no bugs.

I have not yet added type checking to `make test`, but plan to do so soon (adding a list of all files, to be gradually uncommented as more and more of them pass the type checker).

Discussion point: Currently the type checker / symbolizer requires all type variables to be explicitly quantified, including top-level type variables and in constructor definitions. Is this OK or unnecessarily verbose?


aathn:
As discussed during the meeting last time, I encountered a bug when compiling with the `--typecheck` flag, which turned out to be because I had accidentally disabled the post-`utesttrans` run of `typeAnnot` (thanks @elegios for pointing that out!). Unfortunately, I encountered further problems when trying to replace that invocation with `typeCheck`, which I was not immediately able to solve.

For now, I let `--typecheck` replace `typeAnnot` everywhere _except_ for this post-`utesttrans` run, and now everything works. I will need to investigate what caused the problems when running `typeCheck` after `utestGen` further.

PR #555, opened 2022-03-14
Add microbenchmark for normally distributed float generation

johnwikman:
Adds support for floats, both for normal and for uniform distributions. There are no bound arguments to the functions per default as these transformations can easily be done afterwards.


david-broman:
There is already a library under `stdlib/ext/dist-ext.mc` that includes many distributions (both sampling functions and pdf functions). It is using an external library.

david-broman:
But, maybe you have another use case?

johnwikman:
No, I simply did not find it when looking around for a function that can generate from a normal distribution :slightly_smiling_face:

PR #557, opened 2022-03-23
Add compilation to javascript

WilliamRagstad:
This PR adds an _initial_ version of a MExpr to JavaScript back end.
Enable using the `--to-js` flag on the `compile` command to output a `.js` file in the same directory as the source/input file.

### Additions
* `JSExprAst`, `JStmtAst` and combined `JSAst`.
* `JSProgAst` with future support for imports. Acts right now as a "top level" abstraction.
* Support for basic data types and structures
* Convertion from Char sequence to String
* `CPrint` transformed to `console.log` using `JSEApp` and `JSEMember`.
* Pretty printing
* Test files for compilation from MExpr to JS

### Changes
* Renamed `ocamlCompileAstWithUtests` to `compileWithUtests` for general usage.

> **Disclaimer:**
> Lots of things does not work, are not yet implemented, are wrongly implemented and not tested completely.
> Some things that are important to fix have been listed below.

### Todo
* Implement statements
* Ensure that operators work

### Discussion
Is this the way to go when generating JavaScript code? Would it be better to generate https://github.com/estree/estree markup and use the [babel-parser](https://babeljs.io/docs/en/babel-parser) + [babel-generator](https://babeljs.io/docs/en/babel-generator) to emit more syntactically pleasing and maybe more optimized JS code? Is a manual implementation sufficient?

Lastly, should variable declarations use `const` for immutability?
JavaScript also have support for multiple declarations on one line (`const a = 0, b = 1, c;`), should this be considered?


stdlib/javascript/ast.mc
```diff
@@ -0,0 +1,84 @@
...
+
+  syn JSUnOp =
+  | JSONeg       {} -- -arg
+  | JSONot       {} -- !arg
+
+end
+
+
+------------------
+-- C STATEMENTS --
```
david-broman: Why does it say C statements?

david-broman (commented):
Looks like a good start!

WilliamRagstad:
An additional possible optimization would be to use the [Google Closure Compiler](https://developers.google.com/closure/compiler).

PR #559, opened 2022-03-29
Add tensor intrinsics for linear indexing

johnwikman:
Adds the intrinsics

* `tensorLinearGetExn: Tensor[a] -> Int -> a`
* `tensorLinearSetExn: Tensor[a] -> Int -> a -> ()`

which gets or sets elements based on a row-major linear index. I.e. the following statements (1) and (2) are equivalent:

```
let t = tensorCreateDense [3,3] (lam idx. ...) in
let #var"(1)" = tensorGetExn t [1,2] in
let #var"(2)" = tensorLinearGetExn t 5 in
...
```

The goal of this is allow for optimization in the CUDA backend, where there is currently a lot of computations just to work out indices for simple operations such as pair-wise addition of matrices.

### DIFF
--- a/src/boot/lib/tensor.ml
+++ b/src/boot/lib/tensor.ml
@@ -111,7 +111,7 @@ let transpose create shape get_exn t dim0 dim1 =
 
 module Generic : GENERIC = struct
   type ('a, _) t =
-    {data: 'a array; shape: int array; rank: int; stride: int; size: int}
+    {data: 'a array; shape: int array; rank: int; offset: int; size: int}
 
   let rank t = t.rank
 
@@ -126,24 +126,24 @@ module Generic : GENERIC = struct
 
   let get_exn t idx =
     if Array.length idx = rank t && is_valid_index t.shape idx then
-      let linear_idx = cartesian_to_linear_idx t.shape idx + t.stride in
+      let linear_idx = cartesian_to_linear_idx t.shape idx + t.offset in
       t.data.(linear_idx)
     else raise (Invalid_argument "Tensor.Op_mseq_generic.get_exn")
 
   let set_exn t idx v =
     if is_valid_index t.shape idx then
-      let linear_idx = cartesian_to_linear_idx t.shape idx + t.stride in
+      let linear_idx = cartesian_to_linear_idx t.shape idx + t.offset in
       t.data.(linear_idx) <- v
     else raise (Invalid_argument "Tensor.Op_mseq_generic.set_exn")
 
   let linear_get_exn t linear_idx =
     if linear_idx >= 0 && linear_idx < t.size then
-      t.data.(linear_idx + t.stride)
+      t.data.(linear_idx + t.offset)
     else raise (Invalid_argument "Tensor.Op_mseq_generic.linear_get_exn")
 
   let linear_set_exn t linear_idx v =
     if linear_idx >= 0 && linear_idx < t.size then
-      t.data.(linear_idx + t.stride) <- v
+      t.data.(linear_idx + t.offset) <- v
     else raise (Invalid_argument "Tensor.Op_mseq_generic.linear_set_exn")
 
   let reshape_exn t shape =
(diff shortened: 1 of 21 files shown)


src/boot/lib/tensor.ml
```diff
@@ -121,6 +136,16 @@ module Generic : GENERIC = struct
       t.data.(linear_idx) <- v
     else raise (Invalid_argument "Tensor.Op_mseq_generic.set_exn")
 
+  let linear_get_exn t linear_idx =
+    if linear_idx >= 0 && linear_idx < t.size then
+      t.data.(linear_idx + t.stride)
```
br4sco: Do you mind renaming `stride` to lets say `offset` in this PR as well? It would better describe what it really is.

src/boot/lib/tensor.ml
```diff
@@ -79,6 +83,17 @@ let cartesian_to_linear_idx shape idx =
   done ;
   !tmp_ofs
 
+let linear_to_cartesian_idx shape linear_idx =
```
br4sco, line 86: We could perhaps keep a [Array1](https://ocaml.org/api/Bigarray.Array1.html) representation for all tensors next to the [Genarray](https://ocaml.org/api/Bigarray.Genarray.html) representation to avoid having to make this conversion in the future?

johnwikman: [`array1_of_genarray`](https://github.com/ocaml/ocaml/blob/4.12/stdlib/bigarray.ml#L384) only appears to work if the Genarray is 1-dimensional to begin with. Tested it locally:
```ml
# open Bigarray;;
# let t = Genarray.create float32 c_layout [|3;4;5|];;
val t : (float, Bigarray.float32_elt, Bigarray.c_layout) Bigarray.Genarray.t =
  <abstr>
# t;;
- : (float, Bigarray.float32_elt, Bigarray.c_layout) Bigarray.Genarray.t =
<abstr>
# Genarray.get t [|0;0;0|];;
- : float = 0.
# array1_of_genarray t;;
Exception: Invalid_argument "Bigarray.array1_of_genarray".
# array3_of_genarray t;;
- : (float, Bigarray.float32_elt, Bigarray.c_layout) Bigarray.Array3.t =
<abstr>
```

I suppose the option here would be to do [`reshape`](https://github.com/ocaml/ocaml/blob/4.12/stdlib/bigarray.ml#L394), but that [appears to have quite a bit of overhead](https://github.com/ocaml/ocaml/blob/4.12/runtime/bigarray.c#L1203) (specifically looking at the `caml_ba_alloc`):
```c
CAMLprim value caml_ba_reshape(value vb, value vdim)
{
  CAMLparam2 (vb, vdim);
  CAMLlocal1 (res);
#define b (Caml_ba_array_val(vb))
  intnat dim[CAML_BA_MAX_NUM_DIMS];
  mlsize_t num_dims;
  uintnat num_elts;
  int i;

  num_dims = Wosize_val(vdim);
  /* here num_dims is unsigned (mlsize_t) so no need to check (num_dims >= 0) */
  if (num_dims > CAML_BA_MAX_NUM_DIMS)
    caml_invalid_argument("Bigarray.reshape: bad number of dimensions");
  num_elts = 1;
  for (i = 0; i < num_dims; i++) {
    dim[i] = Long_val(Field(vdim, i));
    if (dim[i] < 0)
      caml_invalid_argument("Bigarray.reshape: negative dimension");
    num_elts *= dim[i];
  }
  /* Check that sizes agree */
  if (num_elts != caml_ba_num_elts(b))
    caml_invalid_argument("Bigarray.reshape: size mismatch");
  /* Create bigarray with same data and new dimensions */
  res = caml_ba_alloc(b->flags, num_dims, b->data, dim);
  /* Copy the finalization function from the original array (PR#8568) */
  Custom_ops_val(res) = Custom_ops_val(vb);
  /* Create or update proxy in case of managed bigarray */
  caml_ba_update_proxy(b, Caml_ba_array_val(res));
  /* Return result */
  CAMLreturn (res);

#undef b
}
```

In this case I would prefer to have the index conversion in tensor.ml, and then just piggy-back on whatever `get_exn` and `set_exn` uses.

src/boot/lib/tensor.ml
johnwikman: Done!

src/boot/lib/tensor.ml
br4sco: I see. I was thinking that we should maintain a tuple with two views on the same data for each tensor. One genarray of high rank (the same as the tensor) and one Array1. Then we would of course update slice, reshape, etc. to update both of these views which would lead to an overhead for these operations. The upside would then be that `linear_set` and `linear_get`, which are more performance critical, would be made faster as we could then use `Array1.set` and `Array1.get` to set and get elements in the data. It is just an idea and I don't think we should do it now (if we should at all).

johnwikman: I see what you're getting at. Didn't consider the ocaml case from a performance perspective here. Either, I would prefer that we have the option to create different underlying representations (similar to what we have today) depending on the kind of performance we want, or that we somehow exploit the underlying C-implementation of `get` to the [offset accessing that they do](https://github.com/ocaml/ocaml/blob/4.12/runtime/bigarray.c#L576-608). The latter shouldn't be too much issue since we always enable the `c_layout` for `Genarray` anyway.

PR #567, opened 2022-04-07
Make the intrinsics that take two sequences handle the case where their representation differ

elegios:
This PR makes the `MSeq` intrinsics that take two sequences (`concat`, `equal`, `combine`, and `fold_right2`) handle the case where one sequence is a `List` and one is a `Rope`. If the result is also a sequence then it will have a `List` representation, with the reasoning that you had to opt-in to get a `List`, thus you probably want to keep working with a list.

The implementation uses OCaml's `Seq.t` which are lazy sequences/iterators, which are implemented using anonymous functions. This means that we presumably allocate a lot of anonymous functions, but my hope is that this is cheaper than allocating a full list (since the functions can be discarded as soon as they've been run) without making the code any more annoying.

### DIFF
--- a/src/boot/lib/intrinsics.ml
+++ b/src/boot/lib/intrinsics.ml
@@ -31,8 +31,10 @@ module Mseq = struct
         Rope (Rope.concat_array s1 s2)
     | List s1, List s2 ->
         List (s1 @ s2)
-    | _ ->
-        raise (Invalid_argument "Mseq.concat")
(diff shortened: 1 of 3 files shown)


lingmar (commented):
Do we want to add some `utest`s to e.g. `test/mexpr/seq-test.mc` combining sequences of different representations and pattern matching on lists?

elegios:
probably yes, I'll do that a bit later

PR #575, opened 2022-04-28
Remove order of labels in records

johnwikman:
Use use implicit shortlex ordering of labels with precedence for tuple labels (natural numbers)


johnwikman:
Waiting for a fix from @br4sco on how to handle external records with OCaml before this can be merged. That the tests currently pass is by coincidence of the labels being ordered in the ext-test.mc and exttest.ml files.

PR #580, opened 2022-05-10
Make mlang aliases order-dependent

elegios:
This patch changes type aliases in mlang language fragments to remember their declaration order, which makes it actually reliable to have aliases that depend on each other. I also changed how it works to have multiple aliases of the same name, it's now always an error, instead of an error only if the right-hand side differs (this only applies to inside language fragments, other places still shadow as before).

I also made some minor fixes on how `desugar_ty` works, so it handles shadowing of aliases in normal code properly.


larshum:
I think this looks good. 👍

PR #581, opened 2022-05-10
Thread info throughout `unify` in the typechecker

elegios:
This includes #578 and should be merged after. Only the last commit is new.

This PR changes type-check.mc such that it threads around an `Info` value throughout unification, instead of setting it through a mutable `ref`. This should make the presented `Info` a bit more precise in some cases. The `Info` used is derived from the expression or pattern that "triggers" the unification (e.g., the argument to a `TmConApp`). In cases where there are multiple things triggering (e.g., `then` and `else` branches of a `match`) I've arbitrarily chosen the thing that occurs later in the code, with the assumption that everything else about the type-checker is left-to-right, thus it'd be consistent if the error points to the later thing.

### DIFF
--- a/stdlib/mexpr/type-check.mc
+++ b/stdlib/mexpr/type-check.mc
@@ -55,20 +55,19 @@ let _insertTyCon = lam name. lam ty. lam env : TCEnv.
   {env with tyConEnv = mapInsert name ty env.tyConEnv}
 
 type UnifyEnv = {
+  info: Info,  -- The info of the expression triggering the unification
   names: BiNameMap,
   tyConEnv: Map Name ([Name], Type)
 }
 
-let errInfo = ref (NoInfo ())
-
 let unificationError =
-  lam lhs. lam rhs.
+  lam info. lam lhs. lam rhs.
   let msg = join [
     "Type check failed: unification failure\n",
     "LHS: ", lhs, "\n",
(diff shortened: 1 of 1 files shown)


david-broman:
Are there any performance implications for the type checker because of this change? Would be good to do some simple tests. 

There is now also a conflict that needs to be resolved.

elegios:
It would surprise me, and it doesn't seem like it (time to typecheck `type-check.mc` before and after, respectively):

```
Benchmark 1: mi compile --typecheck --keep-dead-code --exit-before stdlib/mexpr/type-check.mc
  Time (mean ± σ):     13.952 s ±  0.097 s    [User: 13.198 s, System: 0.698 s]
  Range (min … max):   13.854 s … 14.098 s    5 runs

Benchmark 1: mi compile --typecheck --keep-dead-code --exit-before stdlib/mexpr/type-check.mc
  Time (mean ± σ):     13.917 s ±  0.131 s    [User: 13.225 s, System: 0.648 s]
  Range (min … max):   13.721 s … 14.085 s    5 runs

```

(I compiled both versions, then ran them with the repository in the state it was before this PR, so it's fair in that sense)

The previous PR that's already merged probably has more of an impact (haven't measured though), but those were correctness fixes, so just reverting isn't ideal.

PR #583, opened 2022-05-12
Add 0-CFA for sequence intrinsics

lingmar:
Adds 0-CFA support for a subset of the sequence intrinsics. Also adds some type annotations for semantic functions.

The sequence intrinsics that remain are the higher-order ones (`map`, `create` and the likes) as well as `splitAt`. 

One potential problem is performance. The time for running 0-CFA on `mi-lite.mc` increased from ~7.5 to ~23 seconds.

Update: after reducing the number of generated constraints for constants, the runtime is now ~10 s, which seems much more reasonable.


stdlib/mexpr/cfa.mc
```diff
@@ -363,14 +374,61 @@ lang LamCFA = CFA + InitConstraint + LamAst
...
+        CstrInit { lhs = av, rhs = b.ident }
+      else infoErrorExit (infoTm b.body) "Not a lambda in recursive let body"
+    ) bindings
+
+end
+
+lang ConstCFA = CFA + ConstAst + InitConstraint
+
+  syn AbsVal =
+  | AVConst { const : Const, info : Info, args : [Name] }
```
dlunde: This definition makes sense to me. Is there a specific reason for including an `Info` here? It's probably a good idea, but we do not include this for the other `AbsVal`s currently

stdlib/mexpr/cfa.mc
```diff
@@ -490,6 +516,66 @@ lang RecordCFA = CFA + InitConstraint + RecordAst
...
+  syn AbsVal =
+  -- Abstract representation of sequences. Contains a set of names that may
+  -- flow to the sequence.
+  | AVSeq { names: Set Name }
+
+  syn Constraint =
+  -- [{names}] ⊆ lhs ⇒ ∀n ∈ names: {n} ⊆ rhs
+  | CstrSeq {lhs : Name, rhs : Name}
+  -- [{names}] ⊆ lhs ⇒ [{names} ∪ {rhs}] ⊆ res
+  | CstrSeqUnion {lhs : Name, rhs : Name, res : Name}
```
dlunde: These constraints are actually not generated in this fragment. Is this the right place to define them?

stdlib/mexpr/cfa.mc
```diff
@@ -363,14 +374,61 @@ lang LamCFA = CFA + InitConstraint + LamAst
...
+  | (AVConst lhs, AVConst rhs) ->
+    use ConstCmp in
+    let cmp = cmpConst lhs.const rhs.const in
+    if eqi 0 cmp then subi (length lhs.args) (length rhs.args)
+    else cmp
+
+  sem generateConstraints =
+  | TmLet { ident = ident, body = TmConst t } ->
+    let av = AVConst {const = t.val, info = t.info, args = []} in
+    [ CstrInit {lhs = av, rhs = ident} ]
```
dlunde: I think doing this general generation for const constraints is good, but we do lose the ability to exit with an error on unsupported constraints already at the constraint generation step (I see that you added errors during propagation instead). This is probably fine though.

stdlib/mexpr/cfa.mc
```diff
@@ -587,249 +676,282 @@ end
...
+    | CMapi _
+    | CIter _
+    | CIteri _
+    | CFoldl _
+    | CFoldr _
+    | CCreate _
+    | CCreateList _
+    | CCreateRope _
+    | CSplitAt _
+    ) -> infoErrorExit info "Sequence intrinsic not supported in CFA"
```
dlunde: Looks good! But, I think that we should maybe define `CstrSeq` and `CstrSeqUnion` in this fragment instead, as they are only used here. What do you think?

dlunde (approved):
The overall approach to handling constants looks very good. I only have a few minor comments (see below).

stdlib/mexpr/cfa.mc
lingmar: The reason is to propagate the `info` field of the constant to use in `infoErrorExit` in the `propagateConstraintConst` function (for erroring on unsupported constants). The `info` field was previously passed via `generateConstraintsConst`, which was called from `generateConstraints`.

stdlib/mexpr/cfa.mc
lingmar: True. The only impact would be on performance, right? And I guess we would not catch the case when an unsupported constant is used partially (using currying). But that shouldn't really matter to the CFA result and I believe we are planning on not allowing currying on constants anyway?

stdlib/mexpr/cfa.mc
lingmar: Good idea, I'll move them.

stdlib/mexpr/cfa.mc
lingmar: Moving to the `SeqOpCFA` fragment as you suggested below.

stdlib/mexpr/cfa.mc
lingmar: Perhaps I should add a comment about it.

stdlib/mexpr/cfa.mc
dlunde: Yeah, let's leave it as it is now. The performance cost is in any case zero if we analyze only supported programs (which is really the only important case).

PR #590, opened 2022-05-17
Better errors

elegios:
This PR is based on #588 and should thus be merged after it.

I got tired of following purely textual `Info`s in error outputs, so I moved all the errors I could find to using my error highlighting library instead.

- This moves `stdlib/parser/error-highlight.mc` to `stdlib/error.mc`.
- There are now two helper functions for printing errors and immediately exiting, one for highlighting a single range (possibly with sub-ranges) and one for highlighting multiple ranges (that could be in different files). These do a cached `readFile` on the `filename` in the supplied `Info`s, so it should be as easy to use as possible.
- I also changed `type-check.mc` to use the multi-range highlighting when there are two clear things to highlight (e.g., the scrutinee and pattern in a `match`).
- One of the commits is essentially auto-generated, I used [Comby](https://comby.dev/) to rewrite calls to `infoErrorExit` into the new `errorSingle`. In the process I made a configuration file for it which is also in this PR, if we want it.

----

### `error.mc` documentation, copied here for convenience:

This module provides functions for highlighting sections of a file, primarily intended for error reporting in a compiler.

The underlying workhorse is `formatHighlights`, which takes the source of a file and a sequence of ranges to highlight (`Highlight`s, essentially `Info`s with some extra information).

There is also a high-level interface for reporting errors and immediately exiting. This centers around the `ErrorSection` type:

```
type ErrorSection = {msg : String, multi : String, info : Info, infos : [Info]}
let err : ErrorSection = {msg = "", multi = "", info = NoInfo (), infos = []}
```

An `ErrorSection` represents a single contiguous range in a file, possibly with multiple sub-ranges to highlight, and with an optional related message. Typically you would construct this through record updates on `err`, since most fields are optional:

- `infos`: The ranges to highlight. Default to `[info]` if absent.
- `info`: The complete section to display. Defaults to `foldl mergeInfo (NoInfo ()) infos` if absent.
- `msg`: The message to display above the highlighted section, if any.
- `multi`: Used instead of `msg` if present and `infos` contains at least two elements (useful for pluralizing messages).

There are two functions for displaying an error and immediately exiting:
- `errorSingle` takes a single `ErrorSection`.
- `errorMulti` takes a sequence of `ErrorSection`s and a record with two fields (both are optional):
  - `single`: The message to display if only one `ErrorSection` is given.
  - `multi`: The message to display if more than one `ErrorSection` is given. Defaults to `single` if absent.

### DIFF
--- a/stdlib/error.mc
+++ b/stdlib/error.mc
@@ -1,5 +1,6 @@
 include "seq.mc"
 include "char.mc"
+include "map.mc"
 include "mexpr/info.mc"
 
 type Highlight
@@ -164,6 +165,83 @@ let terminalHighlightErrorConfig: HighlightConfig =
   , relevant = lam str. concat (concat "[31m" str) "[0m"
   , added = lam str. concat (concat "[31m" str) "[0m"
   }
+
+type ErrorSection = {msg : String, multi : String, info : Info, infos : [Info]}
+let errorDefault : ErrorSection = {msg = "", multi = "", info = NoInfo (), infos = []}
+let infoToSection : Info -> ErrorSection = lam info. {errorDefault with info = info}
+
+let _cachedContent : Ref (Map String String) = ref (mapEmpty cmpString)
+let _readContent : String -> Option String = lam filename.
+  match mapLookup filename (deref _cachedContent) with c & Some content then c else
+  -- TODO(vipa, 2022-05-17): This is technically a race condition, the
+  -- file could be removed in-between the check and the read, but
+  -- there's no better way to handle it atm.
+  if fileExists filename then
+    let content = readFile filename in
+    modref _cachedContent (mapInsert filename content (deref _cachedContent));
+    Some content
+  else None ()
+
+let _emptyOrNewlineTerm : String -> String = lam str.
+  switch str
+  case "" then str
+  case _ ++ "\n" then str
+  case str then snoc str '\n'
+  end
+
+let _highlightSection
+  : ErrorSection -> (Info, String)
+  = lam section.
+    let info = match section.info with NoInfo ()
+      then foldl mergeInfo (NoInfo ()) section.infos
+      else section.info in
+    let infos = match section.infos with []
+      then [section.info]
+      else section.infos in
+    let infos = map (lam x. Relevant x) infos in
+    let infos =
+      match section.info with Info x then
+        let first = infoVal x.filename x.row1 x.col1 x.row1 x.col1 in
+        let last = infoVal x.filename x.row2 x.col2 x.row2 x.col2 in
+        snoc (cons (Irrelevant first) infos) (Irrelevant last)
+      else infos in
+    let msg = match section.infos with ![] & ![_]
+      then match section.multi with !"" then section.multi else section.msg
+      else section.msg in
+    let msg = _emptyOrNewlineTerm msg in
+    let msg =
+      match info with Info {filename = filename} then
+        match _readContent filename with Some content
+        then concat msg (formatHighlights terminalHighlightErrorConfig content infos)
+        else join [msg, "<Couldn't read '", filename, "', no highlight available>"]
+      else msg in
+    (info, _emptyOrNewlineTerm msg)
+
+let errorMsg
+  : [ErrorSection] -> {single: String, multi: String} -> (Info, String)
+  = lam sections. lam msg.
+    switch map _highlightSection sections
+    case [(info, inner)] then (info, concat (_emptyOrNewlineTerm msg.single) inner)
+    case sections then
+      let msg = match msg.multi with !"" then msg.multi else msg.single in
+      match unzip sections with (infos, inners) in
+      let info = foldl mergeInfo (NoInfo ()) infos in
+      let msg = strJoin "\n" (cons (_emptyOrNewlineTerm msg) inners) in
+      (info, msg)
+    end
+
+let _die : all a. (Info, String) -> a = lam msg.
+  printError (join ["\n", infoErrorString msg.0 msg.1, "\n"]);
+  flushStderr ();
+  exit 1
+let errorGeneral : all a. [ErrorSection] -> {single: String, multi: String} -> a
+  = lam sections. lam msg. _die (errorMsg sections msg)
+let errorSingle : all a. [Info] -> String -> a
+  = lam infos. lam msg.
+    _die (errorMsg [{errorDefault with infos = infos, msg = msg}] {single = "", multi = ""})
+let errorMulti : all a. [(Info, String)] -> String -> a
+  = lam sections. lam msg.
+    _die (errorMsg (map (lam sec. {errorDefault with info = sec.0, msg = sec.1}) sections) {single = msg, multi = ""})
 
 mexpr
 
--- a/stdlib/error.mc
+++ b/stdlib/error.mc
@@ -1,3 +1,48 @@
+/-
+
+This module provides functions for highlighting sections of a file,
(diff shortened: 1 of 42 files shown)


stdlib/error.mc
```diff
@@ -165,6 +207,80 @@ let terminalHighlightErrorConfig: HighlightConfig =
...
+    case sections then
+      let msg = match msg.multi with !"" then msg.multi else msg.single in
+      match unzip sections with (infos, inners) in
+      let info = foldl mergeInfo (NoInfo ()) infos in
+      let msg = strJoin "\n" (cons (_emptyOrNewlineTerm msg) inners) in
+      (info, msg)
+    end
+
+let _die : all a. (Info, String) -> a = lam msg.
+  print (join ["\n", infoErrorString msg.0 msg.1, "\n"]);
```
johnwikman: Add a `flushStdout ()` after this print statement. Otherwise, depending on backend, there is a risk of the program exiting without the print being shown.

Ideally this would be printed on stderr instead of stdout. Maybe add this as a todo?

johnwikman:
So I promised during the meeting on 2022-05-17 that I would comment here on the use of `err` as a top-level name. Since `err` is a nifty ad-hoc name that I frequently use (similar to `i` in loops), I would prefer to have something a bit more verbose on the top-level. Such as:

```
-- change name from "err" to "errorEmpty"
let errorEmpty : ErrorSection = {msg = "", multi = "", info = NoInfo (), infos = []}
```

However, causes the syntax to be a bit more verbose, which makes writing good errors (to some degree) more tedious:

```
-- the current way:
errorSingle {{err with info = infoTm t} with msg = "It doesn't work!"}

-- the alternative way:
errorSingle {{errorEmpty with info = infoTm t} with msg = "It doesn't work!"}
```

An solution could be to add more helper functions to assist the regular cases with an info field and a single error more easy to write. Such as:

```
-- definitions
let errorInfo = lam fi : Info. lam msg : String.
  errorSingle {{errorEmpty with info = fi} with msg = msg}

let errorMultiInfo = lam fis : [Info]. lam msg : String.
  errorMulti {{errorEmpty with infos = fis} with multi = msg}


-- usages
errorInfo (infoTm t) "It doesn't work!"

errorMultiInfo [infoTm parent, infoTm t] "Nothing works!"
```

By this, I also think that we should outlaw the usage of the intrinsic `error` outside of the error.mc function, or just plain an simple remove the `error` intrinsic entirely. The most convenient way to write errors in MCore should be to write good errors! (Stack traces can probably be solved in some other way?)

johnwikman:
So I promised during the meeting on 2022-05-17 that I would comment here on the use of `err` as a top-level name. Since `err` is a nifty ad-hoc name that I frequently use (similar to `i` in loops), I would prefer to have something a bit more verbose on the top-level. Such as:

```
-- change name from "err" to "errorEmpty"
let errorEmpty : ErrorSection = {msg = "", multi = "", info = NoInfo (), infos = []}
```

However, causes the syntax to be a bit more verbose, which makes writing good errors (to some degree) more tedious:

```
-- the current way:
errorSingle {{err with info = infoTm t} with msg = "It doesn't work!"}

-- the alternative way:
errorSingle {{errorEmpty with info = infoTm t} with msg = "It doesn't work!"}
```

An solution could be to add more helper functions to assist the regular cases with an info field and a single error more easy to write. Such as:

```
-- definitions
let errorInfo = lam fi : Info. lam msg : String.
  errorSingle {{errorEmpty with info = fi} with msg = msg}

let errorMultiInfo = lam fis : [Info]. lam msg : String.
  errorMulti {{errorEmpty with infos = fis} with multi = msg}


-- usages
errorInfo (infoTm t) "It doesn't work!"

errorMultiInfo [infoTm parent, infoTm t] "Nothing works!"
```

By this, I also think that we should outlaw the usage of the intrinsic `error` outside of the error.mc function, or just plain an simple remove the `error` intrinsic entirely. The most convenient way to write errors in MCore should be to write good errors! (Stack traces can probably be solved in some other way?)

elegios:
On naming of `err`: maybe `errorDefault` is a better name? This assumes that we do something where simple errors don't need to touch it, since it's even longer.

My thought originally was to have the distance between the simplest use and the most advanced use be short, to minimize the additional cost of providing a better error, and the amount of reading required to figure out how to use the more advanced version. Ultimately it would be nice if we had optional arguments for this, that's what I'm emulating with the record updates on `err`. Maybe we could just enumerate the simple versions with a consistent naming scheme:

```
         Number of sections
         |Number of ranges per section
         ||
         vv
let error11 : all a. Info -> String -> a
let error1N : all a. [Info] -> {one : String, many : String} -> a
let errorN1: all a. [Info] -> {one : String, many : String} -> a
let errorNN : all a. [[Info]] -> {one : String, many : String} -> a
```

Design knobs in this idea:
- Markers for plural, `11` looks like eleven which might not be ideal.
- Should the last three functions take just a `String`? It's easier to use, but it makes it a bunch more effort to support proper pluralization of the message. Then again, many uses of these functions might be in cases where we *know* that we have more than one range/section.
- I think a large fraction of uses of `errorN1` and `errorNN` should want a message per section, so it may be the case that `errorN1 : all a. [(Info, String)] -> {one : String, many: String} -> a` is a better type, and something similar for `errorNN`.

After writing this and thinking a bit more on it I think my current stance is this:
- `errorXX` functions are for when you know you *need* the plurality implied by the function. We could even have a warning if one of the plural versions is called with a singleton list if we want to, at least in some form of debug mode, to make it less likely that we provide bad errors. I think I would argue against adding such a warning, and instead just argue for vigilance against bad messages, but it seemed reasonable to mention for completeness.
- The `errorNN` case might not be useful to have a helper for, because it's fairly likely that the inner list might be a singleton (at least that's what my very limited experience suggests). Having a helper that suggests always pluralizing in this case might encourage not pluralizing properly, so there might be a point in not having it at all (or having a deprecated warning, once we eventually have those).
- The types should thus probably be as follows:
  - `error11 : all a. Info -> String -> a`
  - `error1N : all a. [Info] -> String -> a`
  - `errorN1 : all a. [(Info, String)] -> String -> a`

johnwikman:
Yes `errorDefault` is a better name.

Although sound, I do find the `errorXX` naming a bit cryptic. I am more comfortable with spelling the whole this out:
```
-- This is longer...
errorSingle {errorDefault with info = infoTm t, msg = "My Error Message"}

-- ... but this is harder to understand at first glance.
error11 (infoTm t) "My Error Message"
```

I find these alternatives below to be easier to understand, but they might be less expressive. I am not entirely sure how the `errorN1` case is supposed to be used.
```
let errorS : all a. [Info] -> String -> a = lam fis. lam msg.
  match fis with [fi] then
    error11 fi msg
  else
    error1N fis msg

let errorN : all a. [(Info, String)] -> String -> a = errorN1

errorS [infoTm t] "My Error Message"
errorS [infoTm parent, infoTm t] "Another message with multiple infos."

errorN [(infoTm parent, "Parent error."),
        (infoTm t, "Child error.")]
       "Multiple error messages?"
```

elegios:
It's intended for cases when there are multiple things in the same contiguous code section to highlight, instead of giving multiple sections. The difference between

```
ERROR: unification failed
<...>
let x : () = true in   -- `()` and `true` are highlighted
```
vs
```
ERROR: unification failed
<...>
let x : () = true in   -- only `()` is highlighted

let x : () = true in   -- only `true` is highlighted
```
(using comments to show the highlights, since I don't know how to apply arbitrary coloring in a code block)

PR #592, opened 2022-05-18
Syntactic sugar for multiple record updates

johnwikman:
This PR adds syntactic sugar in the parser for record updates of multiple fields, to avoid the reverse lisp-hell of having to prefix record updates with brackets:

```
-- currently there is a lot of code that looks like this:
let newState = {{{state
  with env = newEnv}
  with pos = incr state.pos}
  with info = newInfo}
in
...

-- this could now be rewritten (with equivalent AST) as
let newState = {state with
  env = newEnv,
  pos = incr state.pos,
  info = newInfo
} in

-- or maybe even neatly inlined
let newState = {state with env = newEnv, pos = incr state.pos, info = newInfo} in
```

Here is an actual example from the miking-ml repo that could be tidied up ([link](https://github.com/johnwikman/miking-ml/blob/84aee7558f1045aa80907daf89788bbd9ede48a0/benchmarks/mnist/bm-mnist.mc#L23-L32)):

```
-- currently:
let params = {{{{{{{nnVanillaSGDParameters
  with batchsize = 128}
  with epochs = 100}
  with init_alpha = 0.9}
  with init_lambda = 0.0}
  with printStatus = true}
  --with printStatus = false}
  with evaluateBetweenEpochs = true}
  with evaluateBeforeFirstEpoch = true}
in

-- with sugar:
let params = {nnVanillaSGDParameters with
  batchsize = 128,
  epochs = 100,
  init_alpha = 0.9,
  init_lambda = 0.0,
  printStatus = true,
  --printStatus = false,
  evaluateBetweenEpochs = true,
  evaluateBeforeFirstEpoch = true
} in
```


dlunde:
I believe we decided against having such syntactic sugar for record updates at one point. But with various other syntactic sugar that has been introduced (e.g., `match .. in` and `switch`), I see no reason why this shouldn't also be added. It's still obvious from the syntax what the underlying AST looks like.

johnwikman:
Yes, I also remembered that discussion from a while back. But compared to the `mexpr ; mexpr` syntax, I think this is quite uncontroversial :slightly_smiling_face:

david-broman:
I agree. It follows the rule we have now with "obvious AST" concrete syntax.

PR #595, opened 2022-05-23
JavaScript backend improvements

WilliamRagstad:
Further development on the JavaScript backend.
Changed method to only compile and generate expression based JS code.

## Added
* `--js-target [web|node|=generic]` flag that import different pre-written JS intrinsic used by the compiler
* Support for the spread operator for iterators (`[...value]`)
* Pattern matching
* Inline (flatten) nested block expressions
* Compile block expressions to IIFE when needed
* Support for recursive lets
* Support for references

## Changed
* Removed statements AST and replaced with expression alternatives (if-else -> ternary)
* Replaced `var` with `let` in generated code (use latest ES6 syntax, `var` also doesn't bind to scope)


stdlib/javascript/intrinsics.mc
```diff
@@ -0,0 +1,59 @@
...
+        args = args,
+        curried = true
+      }
+    else -- No arguments, return the function itself
+      JSEMember {
+        expr = JSEVar { id = runtime },
+        id = name
+      }
+
+let optimizedOpIntrinsic : Name -> Const -> String -> [JSExpr] -> ([JSExpr] -> JSEBinOp) -> JSExpr =
```
larshum: This signature does not type-check. The `JSEBinOp` is not a type, but an instance of `JSExpr`.

larshum: Also, you could use `getConstStringCode 0 c` from `MExprPrettyPrint` to get a `String` representing the name of the intrinsic. I think this is preferable since right now you're essentially defining a custom pretty-printing of the constants when calling this function.

stdlib/javascript/intrinsics.js
```diff
@@ -0,0 +1,34 @@
...
+  ne: lhs => rhs => lhs !== rhs,
+  lt: lhs => rhs => lhs < rhs,
+  le: lhs => rhs => lhs <= rhs,
+  gt: lhs => rhs => lhs > rhs,
+  ge: lhs => rhs => lhs >= rhs,
+  // and: lhs => rhs => lhs && rhs, // Unused
+  // or: lhs => rhs => lhs || rhs,  // Unused
+  // not: lhs => !lhs,              // Unused
+  neg: lhs => -lhs,
+  // abs: lhs => Math.abs(lhs),     // Unused
```
larshum: I think you should remove the commented-out functions, like `abs`.

stdlib/javascript/compile.mc
```diff
@@ -43,73 +46,159 @@ let _charSeq2String = use MExprAst in lam tms.
 let compileJSEnvEmpty = { externals = mapEmpty nameCmp, allocs = [] }
 
 
+
+
 -- Names used in the compiler for intrinsics
-let _consoleLog = use JSAst in
-  JSEMember { expr = JSEVar { id = nameSym "console" }, id = nameSym "log" }
+let _consoleLog = use JSExprAst in
+  JSEMember { expr = JSEVar { id = nameSym "console" }, id = "log" }
```
larshum: If a user had defined a variable named `console`, and it is used before the first `console.log` in the program, this might be pretty-printed as `console1` which would not work.

A solution could be to define the `console` name in `pprint.mc`. Then you can include this name in the initial pprint environment `env` in `printJSProg` so that it becomes reserved (instead of setting it to `pprintEnvEmpty`).

stdlib/javascript/compile.mc
```diff
@@ -43,73 +46,159 @@ let _charSeq2String = use MExprAst in lam tms.
...
+  | CModRef _ -> intrinsicGen "modref" args
+  | CDeRef _ -> intrinsicGen "deref" args
 
   -- Not directly mapped to JavaScript operators
   | CPrint _ ->
-    JSEApp { fun = _consoleLog, args = args }
+    match opts.targetPlatform with CompileJSTP_Node () then intrinsicNode "print" args
+    else
+      -- Warning about inconsistent behaviour
+      printLn "Warning: CPrint might have unexpected behaviour when targeting the web or a generic JS runtime";
```
larshum: You could provide a better warning message by passing the `info` field of the `TmConst` to the `compileCOp` (which should be called `compileJSOp`, right?). Then you could use the `info2str` function from `stdlib/mexpr/info.mc` to show where in the file the CPrint was used.

stdlib/javascript/compile.mc
```diff
@@ -121,56 +210,104 @@ lang MExprJSCompile = MExprAst + JSProgAst
...
-      else error "Unsupported application in compileExpr"
+      else error "Unsupported application in compileMExpr"
     else never
 
   -- Anonymous function, not allowed.
-  | TmLam _ -> error "Anonymous function in compileExpr."
+  | TmLam { ident = arg, body = body } ->
+    let body = (compileMExpr opts) body in
+    -- dprintLn "Compiling lambda";
+    -- dprintLn body;
```
larshum: Please remove the commented out debug-prints.

stdlib/javascript/compile.mc
```diff
@@ -121,56 +210,104 @@ lang MExprJSCompile = MExprAst + JSProgAst
...
+      then (sidToString sid, (compileMExpr opts) expr)
+      else never in
+    JSEObject { fields = map compileField fieldSeq }
 
-  | TmSeq {tms = tms, ty = ty, info = info} & t ->
-    -- Special handling of strings
+  | TmSeq {tms = tms} ->
     -- Check if sequence of characters, then concatenate them into a string
     if _isCharSeq tms then
       match (_charSeq2String tms) with Some str then JSEString { s = str }
```
larshum, line 244: How come you are special casing on strings here?

stdlib/javascript/compile.mc
```diff
@@ -121,56 +210,104 @@ lang MExprJSCompile = MExprAst + JSProgAst
...
-  | TmExt _ -> error "External expressions cannot be handled in compileExpr."
+      JSEArray { exprs = map (compileMExpr opts) tms }
 
   -- Literals
   | TmConst { val = val } ->
     match val      with CInt   { val = val } then JSEInt   { i = val }
     else match val with CFloat { val = val } then JSEFloat { f = val }
     else match val with CChar  { val = val } then JSEChar  { c = val }
     else match val with CBool  { val = val } then JSEBool  { b = val }
+    else match compileCOp opts [] val with jsexpr then jsexpr -- SeqOpAst Consts are handled by the compile operator semantics
```
larshum: This `match` will always succeed since it matches on a variable. So it's probably better to just return `compileCOp opts [] val` directly instead of matching on it.

stdlib/javascript/compile.mc
```diff
@@ -121,56 +210,104 @@ lang MExprJSCompile = MExprAst + JSProgAst
...
+    ) in
+    match data with (exprs, ret) then
+      flattenBlock (JSEBlock {
+        exprs = exprs,
+        ret = ret
+      })
+    else never
+
+  | TmRecLets { bindings = bindings, inexpr = e } ->
+    -- Todo: Compile each binding as a let expression
```
larshum: We have a convention for comments like this: https://github.com/miking-lang/miking/wiki/Conventions.
It should be formatted as `TODO(<user>, <date>): ...`

stdlib/javascript/compile.mc
```diff
@@ -121,56 +210,104 @@ lang MExprJSCompile = MExprAst + JSProgAst
...
+        lam bind : RecLetBinding.
+        match bind with { ident = ident, body = body } then
+          let nop = TmConst { val = CFlushStdout { } } in
+          compileMExpr opts (TmLet { ident = ident, body = body, inexpr = nop })
+        else never
+        ) bindings,
+      ret = (compileMExpr opts) e
+    })
+  | TmType { inexpr = e } -> (compileMExpr opts) e -- no op (Skip type declaration)
+  | TmConApp _ -> error "Constructor application in compileMExpr."
```
larshum: In general, it's better to use `infoErrorExit` and pass the `info` field of the expression, when an `info` is available. This shows where in the file you're compiling the error originates from, instead of where in the compiler the error occured (which is pointless to a user). If #590 is merged before this PR, I think you could update this line to something like (I don't remember what `err` will be called though)
```suggestion
  | TmConApp t -> errorSingle {{err with info = t.info} with msg = "Constructor application in compileMExpr."}
```
Otherwise, I suggest you use the `infoErrorExit` (which is being replaced in that PR, but it's still better than using `error`) as follows
```suggestion
  | TmConApp t -> infoErrorExit t.info "Constructor application in compileMExpr."
```

larshum: Please do this for all places where you use `error` and it isn't too much work to get access to an `info` field (all `Expr` have an `info` field).

larshum (commented):
Here's my comments on your code. I think it mostly looks good apart from what I commented here.

stdlib/javascript/compile.mc
larshum: Since #590 has been merged now, I suggest that you use the `errorSingle` function:
```suggestion
  | TmConApp _ -> errorSingle [t.info] "Constructor application in compileMExpr."
```

stdlib/javascript/compile.mc
WilliamRagstad: It is actually unused and previously replaced with an intrinsic:
```js
print: msg => console.log(...)
```

stdlib/javascript/compile.mc
WilliamRagstad: Because I want hard-coded strings to become strings in the compiled JS as well. Let's say `printLn "nothing"` should become `printLn("nothing")` and not `printLn(['n', 'o', 't', 'h', 'i', 'n', 'g'])`.

WilliamRagstad:
This PR is contained in #607

WilliamRagstad:
Continued in https://github.com/miking-lang/miking/pull/617.

PR #601, opened 2022-06-08
Typechecker and printing improvements

aathn:
This PR fixes two bugs in the typechecker:

- The typechecker would not look through aliases in annotations on recursive let bindings.
- The type variable scope check was incorrect, which prevented applying functions to polymorphic arguments in some cases.

The second bug was discovered by @dlunde in connection with https://github.com/miking-lang/miking/pull/600.
For example, the following program would fail:
```
mexpr
let id : all a. a -> a = lam y. y in
id #frozen"id"
```
When unifying `a1` with `all a2. a2 -> a2`, the type checker would think that the `a2` in the right hand side is escaping its scope, even though the unification keeps `a2` inside its binder. The scope check is now updated to only check variables occurring free in the right-hand side.

EDIT:
This PR now also includes some general code improvements to type-check.mc and pprint.mc, and updates the pretty-printer to use precedence to reduce the number of parentheses. What was previously printed as
```
all r125. ((all r126. ((Bool) -> (r126)) -> ((Bool) -> (r126))) -> (r125)) -> ((Bool) -> (r125))
```
is now printed as
```
all r125. ((all r126. (Bool -> r126) -> Bool -> r126) -> r125) -> Bool -> r125
```
In addition, monomorphic type variables are no longer prefixed with underscore when printed, this is instead used for non-generalizable (weak) variables (which are currently only introduced by un-annotated records, but would occur more frequently if we implemented a value restriction). The monomorphic type variables are no longer referred to as 'weak' in the code base, since they are not generally 'weak' in this sense.

### DIFF
--- a/src/main/compile.mc
+++ b/src/main/compile.mc
@@ -33,6 +33,10 @@ let pprintMcore = lam ast.
   use MExprPrettyPrint in
   expr2str ast
 
+let pprintType = lam ty.
+  use MExprPrettyPrint in
+  type2str ty
+
 let generateTests = lam ast. lam testsEnabled.
   use MCoreCompile in
   if testsEnabled then
@@ -68,6 +72,8 @@ let compileWithUtests = lam options : Options. lam sourcePath. lam ast.
     in
 
     let ast = typeCheck ast in
+    (if options.debugTypeCheck then
+       printLn (join [pprintMcore ast, "\n : ", pprintType (tyTm ast)]) else ());
 
     -- If --runtime-checks is set, runtime safety checks are instrumented in
     -- the AST. This includes for example bounds checking on sequence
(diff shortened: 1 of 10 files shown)


dlunde:
Does this mean that the output of pretty printing a type-checked program is now always parsable? (as needed for https://github.com/miking-lang/miking-dppl/pull/96)

aathn:
It should always be parsable now as long as the program contains no unknown record types, maybe that's enough for your needs? Note that it might still contain unbound variables though, which would make the symbolizer fail.

I'm planning to look into a proper solution as well, where we have a dedicated field for the type inferred by the checker instead of updating the syntactic annotations on lambdas and lets

dlunde:
Not sure, it could very well be enough (I need to test first). In any case, it's not critical at all. I solve it now by just running the type checker and then discarding the resulting annotated program, instead printing the program as it was before type checking (which I'm now sure type checks). Thanks for your updates Anders!

PR #607, opened 2022-06-18
Tail call optimization for JS backend

WilliamRagstad:
All functions in a `recursive let` are automatically tail call optimized using trampolines. This supports both mutual and direct recursion without exploding the call stack.
Added check for optimized functions that verify full applications. This is because the recursive functions are not curried.
Added trampoline intrinsic:
```js
trampolineCapture: fun => args => ({ fun: fun, args: args, isTrampolineCapture: true }),
trampoline: val => { while(val.isTrampolineCapture) val = val.fun(...val.args); return val; }
```

Example:
```lua
mexpr
  recursive let fact_rect = lam acc. lam n.
    if eqi n 0 then acc
    else fact_rect (muli n acc) (subi n 1)
  in
  let fact = lam n. fact_rect 1 n in
  printLn (int2string (fact 5));
  recursive
      let isEven_rect = lam n.
        if eqi n 0 then "true"
        else isOdd_rect (subi n 1)
      let isOdd_rect = lam n.
        if eqi n 0 then "false"
        else isEven_rect (subi n 1)
  in
  printLn (isEven_rect 10);
  ()
```
Compiles to:
```js
let fact_rect = (acc1, n2) => ((n2 === 0) ? acc1 : MExpr_JS_Intrinsics.trampolineCapture(fact_rect)([(n2 * acc1), (n2 - 1)]));
let fact = n3 => MExpr_JS_Intrinsics.trampoline(fact_rect(1, n3));
printLn(int2string(fact(5)));
let isEven_rect = n4 => ((n4 === 0) ? "true" : MExpr_JS_Intrinsics.trampolineCapture(isOdd_rect)([(n4 - 1)]));
let isOdd_rect = n5 => ((n5 === 0) ? "false" : MExpr_JS_Intrinsics.trampolineCapture(isEven_rect)([(n5 - 1)]));
printLn(MExpr_JS_Intrinsics.trampoline(isEven_rect(10)));
{};
```
Note:
* tail-recursive call in `fact_rect` gets converted to a capture.
* All uses of `fact_rect` are fully and directly applied (not curried). The JS compiler ensures this.
* Both `isEven_rect` and `isOdd_rect` are mutually recursive and are optimized correctly.
* All uses of either `isEven_rect` or `isOdd_rect` gets automatically trampolined.
* `int2string` implementation is correctly tail call optimized as well.

Blocked by #595 and #606


src/main/compile.mc
```diff
@@ -82,7 +82,17 @@ let compileWithUtests = lam options : Options. lam sourcePath. lam ast.
...
+      let target = switch options.jsTarget
+        case Some ("node")    then runtimePrint "the node";       CompileJSTP_Node ()
+        case Some ("web")     then runtimePrint "a web browser";  CompileJSTP_Web  ()
+        case Some ("generic") then runtimePrint "a generic";      CompileJSTP_Normal ()
+        case Some (e) then error (join ["Invalid value for --js-target: '", e, "'"])
+        case _ then runtimePrint "a generic"; CompileJSTP_Normal ()
+      end in
+      let res = javascriptCompileFile {compileJSOptionsEmpty with targetPlatform = target} ast sourcePath in
+      printLn (join ["Successfully compiled to JavaScript in ", res]);
+      res
```
larshum, line 95: I would prefer it if this code was placed in a separate function, so that we keep the steps of the compilation relatively easy to follow (i.e. after the "preprocessing" steps, we either compile to JS or OCaml).

src/main/options-type.mc
```diff
@@ -23,6 +23,7 @@ type Options = {
   keepDeadCode : Bool,
   printHelp : Bool,
   toJavaScript : Bool,
+  jsTarget : Option String,
```
larshum, line 26: Since there is a known, finite number of valid values for this parameter, would it not make more sense to use an "enum-like" type (e.g. declare type `JSTargetType` and then define `con JSTargetWeb` etc)? Or perhaps this is due to a limitation of the current argparser library?

src/main/options.mc
```diff
@@ -25,7 +25,8 @@ let optionsDefault : Options = {
   use32BitFloats = false,
   keepDeadCode = false,
   printHelp = false,
-  toJavaScript = false, -- Compile MExpr to JavaScript
+  toJavaScript = false,
+  jsTarget = None (),
```
larshum, line 29: Even if my comment in `options-type.mc` does not apply, would it not make more sense to have this be a `Some x` where `x` is the default value? Surely the target isn't an option? Perhaps this is more of a general problem with the current argparser.

stdlib/javascript/intrinsics.mc
```diff
@@ -0,0 +1,66 @@
...
+include "mexpr/const-arity.mc"
+include "javascript/ast.mc"
+
+
+-- Fragments used by the intrinsic functions
+lang CompileJSOptimizedIntrinsics = JSExprAst + MExprAst + MExprArity + MExprPrettyPrint
+end
+
+-- Compile intrinsic function
+let intrinsicFromString : Name -> String -> [JSExpr] -> JSExpr =
```
larshum, line 12: I think this could be pretty cleanly expressed as a semantic function of `CompileJSOptimizedIntrinsics`. Something like
```suggestion
  sem intrinsicFromString : Name -> String -> [JSExpr] -> JSExpr
  sem intrinsicFromString runtime name =
  | [] -> ... -- the else branch
  | args -> ... -- the then-branch
```

larshum: Note that, as far as I know, we currently don't have a well-defined approach for how to do this. So I think you can leave the function as is if you prefer it that way. I just pointed it out because I prefer what I suggested, because it removes the need to have a `use`, which in turn makes it more extensible. However, that is irrelevant in this case, since the matching is exhaustive.

stdlib/javascript/optimizations.mc
```diff
@@ -0,0 +1,175 @@
...
+      (concat flatExprs retExprs, retRet)
+    else never
+  | expr -> ([], expr)
+
+  sem flattenBlock : JSExpr -> JSExpr
+  sem flattenBlock =
+  | JSEBlock _ & block ->
+    match flattenBlockHelper block with (exprs, ret) then
+      JSEBlock { exprs = exprs, ret = ret }
+    else never
```
larshum, line 43: You could use the `match ... in` pattern here, since you have an `else never`. This reduces the amount of nesting.
```suggestion
    match flattenBlockHelper block with (exprs, ret) in
    JSEBlock { exprs = exprs, ret = ret }
```

stdlib/javascript/optimizations.mc
```diff
@@ -0,0 +1,175 @@
...
+type JSTCOContext = {
+  expr: JSExpr,
+  foundTailCall: Bool
+}
+
+-- Tail Call Optimizations
+lang JSOptimizeTailCalls = JSExprAst
+
+  sem optimizeTailCall : Name -> Info -> CompileJSContext -> JSExpr -> (CompileJSContext, JSExpr)
+  sem optimizeTailCall (name: Name) (info: Info) (ctx: CompileJSContext) =
```
larshum, line 69: When you have a `sem` which declares the types, it's unnecessary to explicitly declare the types of the parameters.
```suggestion
  sem optimizeTailCall name info ctx =
```

stdlib/javascript/optimizations.mc
```diff
@@ -0,0 +1,175 @@
...
+      let ctx = { ctx with trampolinedFunctions = mapInsert name fun ctx.trampolinedFunctions } in
+      (ctx, fun)
+    else
+      -- Otherwise, return the function as is
+      (ctx, fun)
+  | _ -> errorSingle [info] "Non-lambda expressions cannot be optimized for tail calls when compiling to JavaScript"
+
+
+  sem wrapCallToOptimizedFunction : Info -> JSExpr -> Int -> JSExpr -> JSExpr
+  sem wrapCallToOptimizedFunction (info: Info) (fun: JSExpr) (nrArgs: Int) =
```
larshum, line 83: Same as above, you don't need the explicit types of the parameters.
```suggestion
  sem wrapCallToOptimizedFunction info fun nrArgs =
```

stdlib/javascript/optimizations.mc
```diff
@@ -0,0 +1,175 @@
...
+  sem foldFunc =
+  | JSEFun { params = params, body = body } ->
+    let body = foldFunc body in
+    match body with JSEFun { params = paramsNested, body = bodyNested } then
+      JSEFun { params = concat params paramsNested, body = bodyNested }
+    else JSEFun { params = params, body = body }
+  | e -> e
+
+  sem runOnTailPositional : (JSExpr -> JSExpr) -> JSExpr -> JSTCOContext
+  sem runOnTailPositional (action: (JSExpr -> JSExpr)) =
```
larshum, line 107:
```suggestion
  sem runOnTailPositional action =
```

stdlib/javascript/optimizations.mc
```diff
@@ -0,0 +1,175 @@
...
+  | JSEFun      t -> runWithJSTCOCtx action t.body (lam e. JSEFun { t with body = e })
+  | JSEIIFE     t -> runWithJSTCOCtx action t.body (lam e. JSEIIFE { t with body = e })
+  | JSEBlock    t -> runWithJSTCOCtx action t.ret (lam e. JSEBlock { t with ret = e })
+  | JSETernary  t -> runWithJSTCOCtx2 action t.thn t.els (lam e1. lam e2. JSETernary { t with thn = e1, els = e2 })
+  | t -> { expr = t, foundTailCall = false } -- No terms where tail calls can be located
+
+
+  sem runWithJSTCOCtx : (JSExpr -> JSExpr) -> JSExpr -> (JSExpr -> JSExpr) -> JSTCOContext
+  sem runWithJSTCOCtx action expr =
+    | constr ->
```
larshum, line 127: Weird indentation here
```suggestion
  | constr ->
```

stdlib/javascript/optimizations.mc
```diff
@@ -0,0 +1,175 @@
...
+      let res = runOnTailPositional action expr in {
+        expr = constr res.expr,
+        foundTailCall = res.foundTailCall
+      }
+
+  sem runWithJSTCOCtx2 : (JSExpr -> JSExpr) -> JSExpr -> JSExpr -> (JSExpr -> JSExpr -> JSExpr) -> JSTCOContext
+  sem runWithJSTCOCtx2 action expr1 expr2 =
+    | constr ->
```
larshum, line 135:
```suggestion
  | constr ->
```

stdlib/javascript/optimizations.mc
```diff
@@ -0,0 +1,175 @@
...
+lang JSOptimizePatterns = JSExprAst
+
+  sem optimizePattern : JSExpr -> JSExpr
+  sem optimizePattern =
+  | JSEBinOp {
+      op = JSOEq {},
+      lhs = JSEBool { b = true },
+      rhs = e
+    } -> e
+  | e -> e
```
larshum: Perhaps I'm mistaken here, but don't you want to apply this to subexpressions as well? I.e. you want to use `smap`on the expression in the second case.
```suggestion
  sem optimizePattern : JSExpr -> JSExpr
  sem optimizePattern =
  | JSEBinOp {
      op = JSOEq {},
      lhs = JSEBool { b = true },
      rhs = e
    } -> e
  | e -> smapJSExprJSExpr optimizePattern e
```
Assuming you define an `smapJSExprJSExpr` in the AST definition (you could take a look at other AST definitions for inspiration here - it's really useful in cases like this).

stdlib/javascript/patterns.mc
```diff
@@ -0,0 +1,97 @@
...
+      fun = JSEMember {
+        expr = JSEApp { fun = JSEMember { expr = e, id = "slice" }, args = [], curried = false },
+        id = "reverse"
+      },
+      args = [],
+      curried = false
+    }
+
+end
+
```
larshum, line 139: Just a note: I don't think this approach will work when we include more complicated kinds of patterns like and-, or-, and not-patterns. We'll need to do something more general.

stdlib/javascript/pprint.mc
```diff
@@ -40,109 +35,154 @@ let jsKeywords = [
...
+      (env, join [expr, ".", id])
     else never
   | JSEDef { id = id, expr = expr } ->
     match pprintEnvGetStr env id with (env,id) then
-      match printJSDef indent env id expr with (env,str) then
-        (env, join [str, ";"])
+      match (printJSExpr indent env) expr with (env, str) then
+        (env, join ["let ", id, " = ", str])
       else never
     else never
```
larshum, line 89: Here you could use the `match ... in ...` syntactic sugar since the else-branches are `never`.

stdlib/javascript/pprint.mc
```diff
@@ -40,109 +35,154 @@ let jsKeywords = [
...
-      match mapAccumL pprintEnvGetStr env params with (env,params) then
-        let params = join ["(", strJoin ", " params, ")"] in
-        match (printJSExpr ii) env body with (env,body) then
-          (env, join [id, params, " {", pprintNewline ii, body, pprintNewline i, "}"])
-        else never
+    match (printJSFunParams env true params) with (env, args) then
+      match (printJSExpr i) env body with (env,body) then
+        (env, join [args, " => ", body])
       else never
     else never
```
larshum, line 101: Another place where you could use the `match ... in ...` syntactic sugar.
```suggestion
    match (printJSFunParams env true params) with (env, args) in
    match (printJSExpr i) env body with (env,body) in
    (env, join [args, " => ", body])
```

stdlib/javascript/pprint.mc
```diff
@@ -40,109 +35,154 @@ let jsKeywords = [
...
+  | JSEIIFE { body = body } ->
+    let i = indent in
+    let ii = pprintIncr indent in
+    match (printJSExpr ii) env body with (env,body) then
+      (env, join ["(() => ", body, ")()"])
+    else never
```
larshum, line 107:
```suggestion
    match (printJSExpr ii) env body with (env,body) in
    (env, join ["(() => ", body, ")()"])
```

stdlib/javascript/pprint.mc
```diff
@@ -40,109 +35,154 @@ let jsKeywords = [
...
 
+  | JSETernary { cond = cond, thn = thn, els = els } ->
+    let i = indent in
+    match (printJSExpr 0 env) cond with (env, cond) then
+      match (printJSExpr i env) thn with (env, thn) then
+        match (printJSExpr i env) els with (env, els) then
+          (env, join ["(", cond, " ? ", thn, " : ", els, ")"])
+        else never
+      else never
+    else never
```
larshum, line 123:
```suggestion
    match (printJSExpr 0 env) cond with (env, cond) in
    match (printJSExpr i env) thn with (env, thn) in
    match (printJSExpr i env) els with (env, els) in
    (env, join ["(", cond, " ? ", thn, " : ", els, ")"])
```

stdlib/javascript/compile.mc
```diff
@@ -39,138 +46,251 @@ let _charSeq2String = use MExprAst in lam tms.
...
 -------------------------------------------
 -- MEXPR -> JavaScript COMPILER FRAGMENT --
 -------------------------------------------
 
-lang MExprJSCompile = MExprAst + JSProgAst
+lang MExprJSCompile = JSProgAst + PatJSCompile + MExprAst + JSOptimizeBlocks + JSOptimizeTailCalls + JSOptimizePatterns
 
   -- Entry point
-  sem compile =
+  sem compileProg (ctx: CompileJSContext) =
```
larshum, line 118: Use the "new" style for semantic functions.
```suggestion
  sem compileProg : CompileJSContext -> Expr -> JSProg
  sem compileProg ctx =
```

stdlib/javascript/compile.mc
```diff
@@ -39,138 +46,251 @@ let _charSeq2String = use MExprAst in lam tms.
...
     -- Run compiler
-    match compileExpr prog with expr then
+    match compileMExpr ctx prog with expr then
+      let exprs = match expr with JSEBlock { exprs = exprs, ret = ret }
+        then concat exprs [ret]
+        else [expr] in
       -- Return final top level expressions
-      JSPProg { imports = [], exprs = [expr] }
+      JSPProg { imports = [], exprs = exprs }
     else never
```
larshum, line 127: Use the `match ... in ...` syntactic sugar here.

stdlib/javascript/compile.mc
```diff
@@ -39,138 +46,251 @@ let _charSeq2String = use MExprAst in lam tms.
...
-
-  | TmApp _ & app ->
-    recursive let rec: [Expr] -> Expr -> (Expr, [Expr]) = lam acc. lam t.
-      match t with TmApp { lhs = lhs, rhs = rhs } then
-        if _isUnitTy (tyTm rhs) then rec acc lhs
-        else rec (cons rhs acc) lhs
-      else (t, acc)
-    in
-    match rec [] app with (fun, args) then
+  sem compileMExpr (ctx: CompileJSContext) =
```
larshum, line 218: Could use the new `sem` definition here.
```suggestion
  sem compileMExpr : CompileJSContext -> Expr -> JSExpr
  sem compileMExpr ctx =
```

stdlib/javascript/compile.mc
```diff
@@ -39,138 +46,251 @@ let _charSeq2String = use MExprAst in lam tms.
...
+        let isTrampolined = isTrampolinedJs ctx ident in
+        let jsApp = JSEApp {
+          fun = JSEVar { id = ident },
+          args = args,
+          curried = not isTrampolined
+        } in
+        if isTrampolined then
+          match mapLookup ident ctx.trampolinedFunctions with Some fun then
+            wrapCallToOptimizedFunction info fun (length args) jsApp
+          else never
```
larshum, line 234:
```suggestion
          match mapLookup ident ctx.trampolinedFunctions with Some fun in
          wrapCallToOptimizedFunction info fun (length args) jsApp
```

stdlib/javascript/compile.mc
```diff
@@ -39,138 +46,251 @@ let _charSeq2String = use MExprAst in lam tms.
...
-      else (t, acc)
-    in
-    match rec [] app with (fun, args) then
+  sem compileMExpr (ctx: CompileJSContext) =
+  | TmVar { ident = id } -> JSEVar { id = id }
+  | TmApp { info = info } & app ->
+    match foldApp [] app with (fun, args) then
       -- Function calls
+      let args = map (compileMExpr ctx) args in
       match fun with TmVar { ident = ident } then
```
larshum, line 224: Could use the `match ... in ...` pattern here to avoid the extra indentation of all code in the then-branch of this `match`.

stdlib/javascript/compile.mc
```diff
@@ -39,138 +46,251 @@ let _charSeq2String = use MExprAst in lam tms.
...
+    JSEObject { fields = map compileField fieldSeq }
 
-  | TmSeq {tms = tms, ty = ty, info = info} & t ->
-    -- Special handling of strings
+  | TmSeq {tms = tms} ->
     -- Check if sequence of characters, then concatenate them into a string
     if _isCharSeq tms then
       match (_charSeq2String tms) with Some str then JSEString { s = str }
-      else errorSingle [infoTm t] "Non-literal strings currently unsupported."
+      else never
```
larshum, line 261: Another place to apply the `match ... in ...` pattern.

larshum (commented):
A general comment: use the `match ... in ...` syntactic sugar for a `match` where the else-branch is `never`. I think this improves the readability of the code because it removes (unnecessary) nesting. I tried to mark it whenever I saw it (which to a big extent explains the large number of comments), but I probably missed a few places.

WilliamRagstad:
Continued in https://github.com/miking-lang/miking/pull/617.

src/main/options-type.mc
WilliamRagstad: This is what I am trying to do, except I manually parse the argument later.
I'm not sure how the argparser works and didn't want to think too much about it, maybe improve it in a later PR?

PR #626, opened 2022-09-06
dirichlet support check

gizemcaylak:
Checks whether the support is a simplex.


stdlib/ext/dist-ext.mc
```diff
@@ -68,7 +68,7 @@ let categoricalSample : [Float] -> Int =
 external externalDirichletLogPdf : [Float] -> [Float] -> Float
 external externalDirichletSample : [Float] -> [Float]
 let dirichletLogPdf : [Float] -> [Float] -> Float =
-  lam alpha. lam xs. externalDirichletLogPdf xs alpha
+  lam alpha. lam xs. if eqfApprox 1e-15 (foldl (lam sum. lam e. addf e sum) 0. xs) 1. then externalDirichletLogPdf xs alpha else negf inf
```
dlunde: Can be written `foldl addf 0. xs` instead to make it a bit cleaner

dlunde (changes requested):

PR #645, opened 2022-11-14
Mixed updates related to Miking DPPL

dlunde:
- Add `externalSetSeed` intrinsic.
- Generalize partial CPS to accept a `Name -> Bool` function instead of a `Set Name`.

Updates by @larshum  (maybe @larshum can provide further details if needed)
- Fix to `NoInfo` in `duplicate-code-elimination.mc`.
- Update lambda lifting so that it gives lifted parameters new symbols.


larshum:
No, I think that's a good summary.

PR #650, opened 2022-11-22
Add micro benchmarks for `map` and `foldl` implemented in MCore

lingmar:
* Adds micro benchmarks for `map` and `foldl` implemented in pure MCore, that is, without using intrinsic functions. See results below.
* Prefixes functions in `benchmarkcommon.mc` with `bc_` to avoid collisions with stdlib functions.
* Generate a temporary `dune-project` file to avoid getting warnings about that.
* Streamlines the list map micro benchmark to be similar to the others (it was using a heavier workload than the others).

The following are some quick results over 100 runs, showing a significant slowdown of the native versions. The size of the sequences is 10^5. 

Update: something seems to be off with complexities; the below statement is not valid.
~According to some quick tests I did, the complexities of the native versions are correct (that is, they seem to scale linearly).~

## Ropes
About 3x slowdown for `foldl`
```
$ ./run fold_native_rope 100 124
3. Miking compiler:      3.387486s  0.338615s
-------------------------------------------------------------------------------------------------------------------
$ ./run fold_rope 100 124
3. Miking compiler:      0.199286s  0.103352s
5. OCaml native:         4.938560s  0.113232s
```
and about 33x slowdown for `map`
```
$ ./run map_native_rope 100 124
3. Miking compiler:      4.963697s  4.881615s
-------------------------------------------------------------------------------------------------------------------
$ ./run map_rope 100 124
3. Miking compiler:      0.246004s  0.150056s
5. OCaml native:         0.256971s  0.150433s
```

## Lists
About 1.5x slowdown for `foldl`
```
$ ./run fold_native_list 100 124
3. Miking compiler:      0.867523s  0.759224s
-------------------------------------------------------------------------------------------------------------------
$ ./run fold_list 100 124
3. Miking compiler:      0.660040s  0.489972s
5. OCaml native:         0.620571s  0.459079s
```
and about 7x slowdown for `map`
```
$ ./run map_native_list 100 124
3. Miking compiler:      8.291943s  5.220912s
-------------------------------------------------------------------------------------------------------------------
$ ./run map_list 100 124
3. Miking compiler:      0.840659s  0.740034s
5. OCaml native:         0.817393s  0.709123s
```

### DIFF
--- a/test/microbenchmark/benchmarkcommon.mc
+++ b/test/microbenchmark/benchmarkcommon.mc
@@ -2,12 +2,12 @@
 include "string.mc"
 
 recursive
-  let work = lam f. lam n.
+  let bc_work : all a. (() -> a) -> Int -> () = lam f. lam n.
     if eqi n 0 then
       ()
     else
       f();
-      work f (subi n 1)
+      bc_work f (subi n 1)
 end
 
-let repeat = lam f. work f (string2int (get argv 1))
+let bc_repeat = lam f. bc_work f (string2int (get argv 1))
--- a/test/microbenchmark/factorial.mc
+++ b/test/microbenchmark/factorial.mc
@@ -10,4 +10,4 @@ let fact = lam n.
     else muli n (fact (subi n 1))
 in
 
-repeat (lam. fact 100)
+bc_repeat (lam. fact 100)
--- a/test/microbenchmark/fold_list.mc
+++ b/test/microbenchmark/fold_list.mc
@@ -12,4 +12,4 @@ in
 
 -- printLn (int2string (foldf n));
 
-repeat (lam. foldf n)
+bc_repeat (lam. foldf n)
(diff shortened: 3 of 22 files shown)


david-broman:
Interesting. Can you specify a bit more what you mean by native in this context? When you say native, is that without intrinsic functions or does it refer to compiling into native machine code code? It would be good if you could explain a bit more what the different test cases mean. Thanks.

lingmar:
The "native" micro benchmarks (`{map|fold}_native_{list|rope}`) use `map` and `foldl` implemented directly in MCore. These functions are copied from the `stdlib/seq-native.mc` file. The non-native micro benchmarks (`{map|fold}_{list|rope}`) use the intrinsic functions `map` and `foldl`. All `map` micro benchmarks iterate over a sequence and add `1` to each element. All the `fold` micro benchmarks fold over a sequence, also using `addi` as fold function.

PR #656, opened 2022-12-06
Update to use OCaml 5 (release candidate 1)

larshum:
This PR updates the repo to use the latest version of OCaml (5.0.0~rc1, as of writing). The changes are:
* Upgraded the `ocamlformat` version from `0.20.1` to `0.24.1` (we'll need to make another PR to update the git-refs as we did [last time](https://github.com/miking-lang/miking/pull/546)). This resulted in a minor change in `src/boot/lib/patterns.ml`.
* Updated `stdlib/multicore/thread.ext-ocaml.mc` because the API changed in the latest OCaml version.
* Rewrote the `README.md` to show how to install the new OCaml version (also naming the switch `miking-ocaml`, not sure if we want to do that).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Install OCaml and opam for Miking
         uses: ocaml/setup-ocaml@v2
         with:
-          ocaml-compiler: 5.0.0~beta2
+          ocaml-compiler: ocaml-base-compiler.5.0.0~rc1
           opam-repositories: |
             default: https://github.com/ocaml/opam-repository.git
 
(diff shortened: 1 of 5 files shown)


lingmar:
Wow! Noticed any speedup in bootstrapping?

larshum:
I haven't looked closely at the performance since I tried the OCaml 5 beta. Then I found that the total runtime of `make test-all` was ≈15% faster than the `4.12.0+domains` switch. Presumably, the performance hasn't changed much since it should include bug fixes.

david-broman:
Have you also changed in the scripts for the automatic tests on Github so that it installs and tests with version 5?

larshum:
Yes, that's the [main.yml file](https://github.com/miking-lang/miking/pull/656/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3).

PR #660, opened 2022-12-20
Always apply pattern lowering

larshum:
This PR fixes an issue that caused the pattern lowering to not always be applied and resolves a performance issue introduced when always using pattern lowering. In addition, it includes a significant number of changes that were required to make the tests pass. After this PR has been merged, we can update the OCaml backend to use a simpler approach for pattern compilation, which should (hopefully) improve performance as we no longer need to generate matches using `option`s. As we now run pattern lowering in `mi-lite`, the first two stages of bootstrapping are slowed down. I hope this will be improved after updating the compilation of patterns to OCaml.

@elegios co-authored the main changes in the pattern lowering and the changes in the type checker.

Additional changes included in this PR to get tests to pass after the above updates:
* Updated the type checker to replace `RecordVar`s with a `TyRecord` instead of a `TyUnknown`.
* Fixed untyped patterns being generated in the utest translation.
* Added explicit types to the `pprintVarName`, `pprintConName`, `pprintTypeName`, and `pprintLabelString` functions and fixed resulting type errors.
* Includes `MExprPrettyPrint` in `mi-lite` so that type errors are correctly printed
  * The type checker uses `type2str` when generating a unification error, but it does not include the pretty-print fragments needed to print anything. Perhaps this is considered an issue in the type checker rather than in `mi-lite`?
* Update the JS backend to pass its tests after making the changes discussed here. These changes will likely degrade performance, but I found correctness more critical (passing all tests). Feel free to optimize what I did @WilliamRagstad.
  * Simplified the compilation of pattern matching by using the lowering guarantees. I removed all special-case handling, which may have resulted in worse execution times.
  * Fixed a bug in the pretty printing of JavaScript objects.
  * Updated the JS tests to use `mi compile` instead of interpreting the compiler via `boot`.

Edit: Here are the results from running the bootstrapping, including the changes in this PR vs. `develop`:
<details id="detail-summary-1">
<summary>Current</summary>

```
Bootstrapping the Miking compiler (1st round, might take a few minutes)
50.08user 0.39system 0:50.49elapsed 99%CPU (0avgtext+0avgdata 996640maxresident)k
0inputs+50920outputs (0major+351999minor)pagefaults 0swaps
Bootstrapping the Miking compiler (2nd round, might take some more time)
27.89user 0.95system 0:29.79elapsed 96%CPU (0avgtext+0avgdata 2210864maxresident)k
0inputs+264600outputs (0major+945011minor)pagefaults 0swaps
Bootstrapping the Miking compiler (3rd round, might take some more time)
30.65user 1.17system 0:32.18elapsed 98%CPU (0avgtext+0avgdata 2247800maxresident)k
0inputs+267376outputs (0major+1046978minor)pagefaults 0swaps
```
</details>
<details id="detail-summary-2">
<summary>After this PR</summary>

```
Bootstrapping the Miking compiler (1st round, might take a few minutes)
73.11user 0.41system 1:13.54elapsed 99%CPU (0avgtext+0avgdata 1219044maxresident)k
0inputs+59024outputs (0major+421327minor)pagefaults 0swaps
Bootstrapping the Miking compiler (2nd round, might take some more time)
31.87user 1.14system 0:33.17elapsed 99%CPU (0avgtext+0avgdata 2264392maxresident)k
0inputs+297104outputs (0major+1082602minor)pagefaults 0swaps
Bootstrapping the Miking compiler (3rd round, might take some more time)
32.48user 1.15system 0:33.82elapsed 99%CPU (0avgtext+0avgdata 2429080maxresident)k
0inputs+297104outputs (0major+1122116minor)pagefaults 0swaps
```
</details>

### DIFF
--- a/src/main/mi-lite.mc
+++ b/src/main/mi-lite.mc
@@ -8,6 +8,7 @@
 include "options.mc"
 include "parse.mc"
 include "mexpr/boot-parser.mc"
+include "mexpr/pprint.mc"
 include "mexpr/shallow-patterns.mc"
 include "mexpr/symbolize.mc"
 include "mexpr/type-check.mc"
@@ -16,7 +17,7 @@ include "ocaml/mcore.mc"
 
 lang MCoreLiteCompile =
   BootParser + MExprSym + MExprTypeCheck + MExprUtestTrans + MCoreCompileLang +
-  MExprLowerNestedPatterns
+  MExprLowerNestedPatterns + MExprPrettyPrint
 end
 
 -- NOTE(larshum, 2021-03-22): This does not work for Windows file paths.
@@ -68,7 +69,6 @@ let compile : Options -> String -> () = lam options. lam file.
   let ast = symbolize ast in
   let ast = typeCheck ast in
   let ast = lowerAll ast in
-  printLn "After pattern lowering";
   let hooks = mkEmptyHooks (ocamlCompile options file) in
   compileMCore ast hooks;
   ()
--- a/src/main/mi-lite.mc
+++ b/src/main/mi-lite.mc
@@ -68,7 +68,6 @@ let compile : Options -> String -> () = lam options. lam file.
   let ast = utestStrip ast in
   let ast = symbolize ast in
   let ast = typeCheck ast in
-  let ast = lowerAll ast in
   let hooks = mkEmptyHooks (ocamlCompile options file) in
   compileMCore ast hooks;
   ()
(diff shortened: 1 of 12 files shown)


WilliamRagstad:
Good solution on removing deprecated compilation of nested patterns. I'll see what I can do optimization-wise 👍

Should maybe shallow patterns be a separate AST type? To indicate that nested patterns are not possible for MExpr backends?

PR #661, opened 2022-12-20
New flag: --debug-phases

elegios:
Introduce a new debug-flag for keeping track of some numbers related to phases, to make basic performance sanity checks easier.

Example output:

```
❯ MCORE_STDLIB=(pwd)/stdlib mi compile --debug-phases src/main/mi.mc
parsing
  Ast size: 1038556
  Phase duration: 14252.1191406
make-keywords
  Ast size: 1038556
  Phase duration: 343.031982422
accelerate
  Ast size: 1038556
  Phase duration: 137.394042969
tuning
  Ast size: 1038556
  Phase duration: 240.879150391
symbolize
  Ast size: 1038556
  Phase duration: 581.970947266
instrument-profiling
  Ast size: 1038556
  Phase duration: 0.
typecheck
  Ast size: 2723232
  Phase duration: 2389.10522461
runtime-checks
  Ast size: 2723232
  Phase duration: 0.001220703125
generate-tests
  Ast size: 2723232
  Phase duration: 144.854736328
symbolize2
  Ast size: 2723232
  Phase duration: 620.866943359
pattern-lowering
  Ast size: 2923683
  Phase duration: 1026.57299805
backend
  Ast size: 2923683
  Phase duration: 35104.894043
```

### DIFF
--- a/src/main/compile.mc
+++ b/src/main/compile.mc
@@ -12,6 +12,7 @@ include "mexpr/type-check.mc"
 include "mexpr/remove-ascription.mc"
 include "mexpr/utesttrans.mc"
 include "mexpr/shallow-patterns.mc"
+include "mexpr/phase-stats.mc"
 include "tuning/context-expansion.mc"
 include "tuning/tune-file.mc"
 include "ocaml/ast.mc"
@@ -30,7 +31,7 @@ lang MCoreCompile =
   MExprUtestTrans + MExprRuntimeCheck + MExprProfileInstrument +
   MExprPrettyPrint +
   MExprLowerNestedPatterns +
-  OCamlTryWithWrap + MCoreCompileLang
+  OCamlTryWithWrap + MCoreCompileLang + PhaseStats
 end
 
 lang TyAnnotFull = MExprPrettyPrint + TyAnnot + HtmlAnnotator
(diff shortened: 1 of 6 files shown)


larshum:
I think this seems like a useful flag. In the future, I imagine we could extend this to a more general debug flag for the compiler, showing execution time and memory usage of each pass (perhaps something like what you get from running `ocamlc` or `ocamlopt` with the `-dprofile` flag?).

But I don't think you should place the counting code in `src/main/mi.mc` for two reasons:
1. This results in it not being available from the standard library, so we cannot reuse the code in other projects. Because of this, we should aim to move as much code as possible out of the files in `src/main/`.
2. The code in `src/main/mi.mc` only includes the "front-end" part of the compiler, i.e., not the backend-specific parts. For the OCaml backend, we could add it as a new hook in `stdlib/ocaml/mcore.mc`.

elegios:
> I think this seems like a useful flag. In the future, I imagine we could extend this to a more general debug flag for the compiler, showing execution time and memory usage of each pass (perhaps something like what you get from running `ocamlc` or `ocamlopt` with the `-dprofile` flag?).

The future is now (except for memory usage). Currently this just treats each backend as a single opaque phase, might be nicer to split that.

larshum:
Awesome! Yeah, the memory usage stuff requires a bit more work, so I wasn't expecting that to happen in the near future. I think it looks good now, doing it in the backends as well could be a separate PR imo.

PR #662, opened 2022-12-22
Better type errors

aathn:
This pull request improves the error messages given by the type checker.  This consists primarily in
- Changing unification errors to present one type as 'expected' and the other as 'found'
- Updating the type checking of function applications to give better errors
- ~Making unification preserve aliases as long as possible~
- Presenting the errors in a nicer way

The pull request also changes the printing of unknown record types to be more helpful: `_a<:{x:_a1}` is now printed as `{x:_a1 ... }` instead.

Finally, the PR updates how errors with info fields are printed, from 
```
FILE "/path/to/file.mc" l1:c1-l2:c2 ERROR: ...
``` 
to
```
ERROR </path/to/file.mc l1:c1-l2:c2>:
...
```
I think this style is easier to read, but I know it's a controversial change which might warrant more discussion.  In that case, I can drop it from this PR.

To demonstrate the new style of type errors, consider the example I had in my workshop presentation, where we modify the `TmLet` case in `eval.mc` to not pass the `ctx` parameter in the recursive call.
```haskell
  sem eval ctx =
  | TmLet t ->
    eval {ctx with env = evalEnvInsert t.ident (eval t.body) ctx.env}
      t.inexpr
```
Previously, we would get the error
```
FILE "/home/aathn/Documents/projects/miking/miking/stdlib/mexpr/eval.mc" 120:48-120:52 ERROR: Type check failed: unification failure
LHS: {env: Env}
RHS: Expr
while unifying these:
LHS: EvalCtx -> Expr -> Expr
RHS: Expr -> a
    eval {ctx with env = evalEnvInsert t.ident (*eval* t.body) ctx.env}
```
whereas after this PR, we get
```
ERROR </home/aathn/Documents/projects/miking/miking/stdlib/mexpr/eval.mc 120:53-120:59>:
* Expected an expression of type: {env: Env}
* Found an expression of type: Expr
    eval {ctx with env = evalEnvInsert t.ident (eval *t.body*) ctx.env}
```
(here, I indicate highlighting with asterisks, although color is used in the actual output).

EDIT: I'll move the change to preserve aliases to a separate PR, as it requires a bit more work (see discussion below).

### DIFF
--- a/stdlib/mexpr/info.mc
+++ b/stdlib/mexpr/info.mc
@@ -51,20 +51,20 @@ let infoVal : String -> Int -> Int -> Int -> Int -> Info =
 
 -- Generate a string from an info
 let info2str : Info -> String = lam fi.
-  match fi with NoInfo () then "[No file info] " else
+  match fi with NoInfo () then "<No file info>" else
   match fi with Info (r & {row1 = 0}) then
-    join ["FILE \"", r.filename, "\" "]
+    join ["<", r.filename, ">"]
   else match fi with Info r then
-    join ["FILE \"", r.filename, "\" ", int2string r.row1, ":", int2string r.col1,
-    "-", int2string r.row2, ":", int2string r.col2, " "]
+    join ["<", r.filename, " ", int2string r.row1, ":", int2string r.col1,
+    "-", int2string r.row2, ":", int2string r.col2, ">"]
   else never
 
 -- Generate an info error string
 let infoErrorString : Info -> String -> String = lam fi. lam str.
-    join [info2str fi, "ERROR: ", str]
+    join ["ERROR ", info2str fi, ":\n", str]
 
 let infoWarningString : Info -> String -> String = lam fi. lam str.
-    join [info2str fi, "WARNING: ", str]
+    join ["WARNING ", info2str fi, ":\n", str]
 
 -- Print an error with info struct info and exit (error code 1)
 let infoErrorExit : Info -> String -> Unknown = lam fi. lam str.
--- a/stdlib/mexpr/pprint.mc
+++ b/stdlib/mexpr/pprint.mc
@@ -1122,7 +1122,7 @@ lang VarSortPrettyPrint = PrettyPrint + RecordTypeAst + VarSortAst
   | RecordVar r ->
     let recty = TyRecord {info = NoInfo (), fields = r.fields} in
     match getTypeStringCode indent env recty with (env, recstr) in
-    (env, concat (init recstr) " ... }")
+    (env, join [init recstr, " ... ", [last recstr]])
   | _ -> (env, idstr)
 end
 
--- a/stdlib/mexpr/pprint.mc
+++ b/stdlib/mexpr/pprint.mc
@@ -1122,7 +1122,7 @@ lang VarSortPrettyPrint = PrettyPrint + RecordTypeAst + VarSortAst
   | RecordVar r ->
     let recty = TyRecord {info = NoInfo (), fields = r.fields} in
     match getTypeStringCode indent env recty with (env, recstr) in
-    (env, join [idstr, "<:", recstr])
+    (env, concat (init recstr) " ... }")
   | _ -> (env, idstr)
 end
 
--- a/stdlib/mexpr/type-check.mc
+++ b/stdlib/mexpr/type-check.mc
@@ -181,7 +181,8 @@ lang FlexTypePrettyPrint = IdentifierPrettyPrint + VarSortPrettyPrint + FlexType
     match deref t.contents with Unbound t then
       match pprintVarName env t.ident with (env, idstr) in
       match getVarSortStringCode indent env idstr t.sort with (env, str) in
-      let weakPrefix = if t.isWeak then "_" else "" in
+      let weakPrefix =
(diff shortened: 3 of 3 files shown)


aathn:
Building currently fails with the following error, in the third stage of bootstrapping.
```
ERROR </home/aathn/Documents/projects/miking/miking/stdlib/mexpr/type-check.mc 528:40-528:84>:
I can't immediately see the record type of this pattern, it's a FlexTypeAst_FlexVarRec
    match deref t.contents with Unbound {ident = n, level = k, sort = s, isWeak = w} then
```
I guess this is because the type checker now preserves more aliases, so that the record gets the type `FlexTypeAst_FlexVarRec` instead of the actual record type referred to by the alias.

I don't know if this should be solved by resolving the aliases at the end of type checking, similar to `TyFlex` links, or if it should be the responsibility of later code to resolve aliases when needed... What do you think?

A third option would be if the type checker could return a mapping of aliases to types for later stages to use, but this is not entirely straightforward, as MExpr allows local type bindings so that we cannot just return a global mapping of aliases.

larshum:
> I guess this is because the type checker now preserves more aliases, so that the record gets the type FlexTypeAst_FlexVarRec instead of the actual record type referred to by the alias.

Yes, it is. This particular error occurs in the pattern lowering, where we need to know the shape of the record.

Regarding the resolution of aliases, I would prefer if they were removed after type-checking, but I don't have a strong opinion.

elegios:
> A third option would be if the type checker could return a mapping of aliases to types for later stages to use, but this is not entirely straightforward, as MExpr allows local type bindings so that we cannot just return a global mapping of aliases.

We could actually still do that given symbolize, but it's still somewhat tricky to use for later passes, since we'd need to handle parameterized aliases. There's some of this already, since ADTs can also be parameterized, which makes it inconvenient to, e.g., enumerate the constructors of an ADT accurately.

I would probably rather argue for resolving aliases fully after type-checking.

There are also at least two additional ways we could deal with aliases that would be more fancy than what we (and GHC, which is the primary other source of inspiration I have) do:
- Have an additional field on every type, similar to the `ty` field on `Expr`s, but this one represents "this used to be this alias", and use that in error messages in some useful way.
- Resolve aliases immediately, but remember their declarations. Treat the right-hand side of an alias as a (non-linear) pattern over types, then when printing, try to pattern match the type being printed using the aliases in scope.

The former approach has some risks with accidentally dropping aliases (which I guess is also true with the status quo), but could also preserve the aliases quite far without forcing code to resolve them. The latter is more general, and could simplify printed types even when the original wasn't written that way, but is likely to be more costly (which might be ok, given that it should only happen for error messages).

aathn:
I have reverted the change to aliases so that the PR can pass the tests.  I will make a separate PR for the aliases, where I try to implement a variant of @elegios's first suggestion above.

Instead of a new field on all types, I'm thinking we could have a new `syn` for aliased types, `TyAlias {display : Type, content : Type}`, which is treated as `content`, but printed as `display`.  Type checking will replace aliases with these link nodes, and I will provide a function `inspectTy : Type -> Type` for later code to use when it wants to inspect the structure of a type, which both passes through these aliases and discards any leading forall quantifiers.

This lets us remember the alias information with a low overhead on usage, without adding a new field to every type.  What do you think?

I got this idea from https://github.com/kztk-m/sparcl/blob/master/src/Language/Sparcl/Typing/Type.hs#L59

larshum:
I think that sounds good. I assume both types would have the same info field, so that `infoTy` could return the info of either of them (rather than adding an extra `info` for the `TyAlias` itself - or perhaps that's what we would want?). I'm not sure how the `smapAccumL_Type_Type` should work, though, in terms of what the expected behaviour would be.

By the way: we still have aliased types after type checking. Currently, we're explicitly unwrapping aliases in places such as in the pattern lowering and OCaml code generation, where we need to know the shape of the underlying records. This seems a bit inefficient since we're doing it multiple times (and we need an environment keeping track of which aliases are defined). I assume all such occurrences would be `TyAlias`es after your proposed changes?

aathn:
> I think that sounds good. I assume both types would have the same info field, so that infoTy could return the info of either of them (rather than adding an extra info for the TyAlias itself - or perhaps that's what we would want?). I'm not sure how the smapAccumL_Type_Type should work, though, in terms of what the expected behaviour would be.

I was thinking that the info of the type should be that of `display`, since that's what would be in the program source. For `smapAccumL_Type_Type`, my intuition would be to pass over `content`, as the type should be treated as `content`.

> By the way: we still have aliased types after type checking. Currently, we're explicitly unwrapping aliases in places such as in the pattern lowering and OCaml code generation, where we need to know the shape of the underlying records. This seems a bit inefficient since we're doing it multiple times (and we need an environment keeping track of which aliases are defined). I assume all such occurrences would be TyAliases after your proposed changes?

I guess so, yes. (Note that the pattern lowering isn't actually doing alias unwrapping at the moment though, which is what caused the build error for this PR). I started implementing these changes, making `symbolize` resolve aliases and insert `TyAlias` nodes, so that later code can simply use a semantic function `unwrapType` to get at the underlying type. It seems to work nicely, although I still haven't gotten the interaction with type lifting to work properly.

PR #667, opened 2023-01-01
Updated handling of aliases

aathn:
This PR changes the way aliases are handled, both simplifying the type checker code and later passes. This is done by introducing a new AST node `TyAlias {display : Type, content: Type}`, which is printed as `display`, but treated as `content`. `symbolize` will now resolve type aliases and insert these nodes, so that later code can easily access the underlying type.

The PR also makes the type checker preserve aliases during unification, leading to better error messages.

Two new semantic functions are added to the base `Ast` fragment:
- `unwrapType: Type -> Type`, which passes through any alias nodes to expose the underlying type.
- `inspectType : Type -> Type`, which also discards any leading forall quantifiers, useful e.g. when inspecting function types.

These are backed by a general semantic function `rappAccumL_Type_Type`. The fragment for unification variables in the type checker also extends this function, making it seamless to use `inspectType` to pass through links. This also means we could skip the elimination of unification variables and links at the end of type checking if we like, but I left it in place for now.

This PR is based on #662. After these changes, the type checker gives error messages along the lines of
```
ERROR </home/aathn/Documents/projects/miking/miking/stdlib/mexpr/eval.mc 120:53-120:59>:
* Expected an expression of type: EvalCtx
* Found an expression of type: Expr
    eval {ctx with env = evalEnvInsert t.ident (eval *t.body*) ctx.env}
```
Here, `EvalCtx` is an alias for `{env: Env}`, and the error message now uses the former instead of the latter.

~I expect there's a risk this PR breaks some code in `miking-dppl`.~ I have now made a [PR](https://github.com/miking-lang/miking-dppl/pull/122) for `miking-dppl` with the required changes there, which should be merged immediately after this one.


larshum:
Looks good to me!

I think this should be merged before #663 and #665 so I can update them to use the new `unwrapType` function.

PR #669, opened 2023-01-17
LR(k) parser generator

johnwikman:
This adds a LR(k) parser generator that hooks in to the existing parser/lexer.mc library. The parser generator is still in pretty rough shape with plenty of low-hanging fruits for improvement. But seeing as it at least is in a working state I suggest that it can be merged in so that other parts of the bootstrapping process that depends on this can start being worked on.

See the files under test/examples/parser for practical examples of how the parser generator can be used.


johnwikman:
@elegios Most of the functional code is now in place, just ironing out some issues and improving error reporting. There are some test grammars under `test/examples/parser` where I interface against your lexer. Please double-check if I understood the proposed interface correctly. I had to add some glue code to satisfy the current implementation.

PR #672, opened 2023-01-20
Fix bug in symbolization

larshum:
This PR fixes a nasty bug in the symbolization, which led to the type parameters of a free type variable would be removed. The bug shows up in Miking DPPL, where the `allowFree` flag is set to allow symbolizing types defined in the runtime (the `Dist` type).

I don't think we use this option elsewhere, and currently, there are no tests in `symbolize.mc`, which is why this bug managed to sneak through.

### DIFF
--- a/stdlib/mexpr/symbolize.mc
+++ b/stdlib/mexpr/symbolize.mc
@@ -411,7 +411,7 @@ lang ConAppTypeSym = ConTypeAst + AppTypeAst + AliasTypeAst + VariantTypeAst +
             appTy
         else
           if env.strictTypeVars then
-            if env.allowFree then ty
+            if env.allowFree then smap_Type_Type (symbolizeType env) ty
             else errorSingle [t.info] (join [
               "* Encountered an unknown type constructor: ", str, "\n",
               "* When symbolizing the type"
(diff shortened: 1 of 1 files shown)


larshum:
I found more bugs under the hood, so this PR isn't quite ready yet.

david-broman:
Good catch Lars! Can you add some tests in symbolize in this PR? At least some tests that test your bug fix? Thanks.

larshum:
Yeah, sure I can do that.

larshum:
I have added tests now. The remaining problems are unrelated to the symbolizer, so I'll mark this PR as ready for review.

stdlib/mexpr/symbolize.mc
```diff
@@ -411,7 +411,7 @@ lang ConAppTypeSym = ConTypeAst + AppTypeAst + AliasTypeAst + VariantTypeAst +
             appTy
         else
           if env.strictTypeVars then
-            if env.allowFree then TyCon t
+            if env.allowFree then smap_Type_Type (symbolizeType env) ty
```
aathn: I think `smap_Type_Type (symbolizeType env) ty` could be replaced by `mkAppTy (TyCon t) args`, which would be more efficient since the type arguments are symbolized once already above.

This bug must have been introduced by me in #667, I'm sorry about that.

larshum: Nice catch! I updated the code as you proposed.

PR #675, opened 2023-01-26
Print summary when utests fail in `mi`

larshum:
This PR updates the utest generation (used in both `mi eval` and `mi compile`) to print a summary of how many tests passed and failed, when at least one test failed (before exiting with code 1). For now, I just used the same format as in `boot`: `ERROR! <x> successful tests and <y> failed tests.`

Also included in this PR:
* Bugfix in the JS compiler in the handling of unit-type arguments.
* Removal of the `utesttrans.mc` code (this should have been removed in #665)


larshum:
It seems a few JS tests are failing after I changed the format of the generated code from utests. I will take a look at the JS codegen to see if I can fix it.

PR #680, opened 2023-02-10
Pretty print improvments

br4sco:
This PR:
- Changes the MExpr pretty printer to recognize the syntactic sugar `match t1 then t2 else never` = `match t1 in t2`;
- Adds pretty printing and equality checks of tensor values in MExpr `eval`;
- Prunes some dependencies from the `LamAst` langauge fragment.

### DIFF
--- a/stdlib/mexpr/ast.mc
+++ b/stdlib/mexpr/ast.mc
@@ -295,7 +295,7 @@ end
 
 
 -- TmLam --
-lang LamAst = Ast + VarAst + AppAst
+lang LamAst = Ast
   syn Expr =
   | TmLam {ident : Name,
            tyAnnot : Type,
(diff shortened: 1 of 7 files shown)


larshum:
Nice! How come you added pretty print and equality functions for tensors in `eval`? When are those needed?

larshum:
Oh I see, it's specifically for tensor values in the interpreter, right?

br4sco:
> 

I was not sure where to put these. I wanted to compare and pretty print a program that evaluated to a tensor value in a utest. Since the TmTensor only lives in eval.mc I figured these functions belonged there. I'm open to suggestions if you think it belongs somewhere else.

larshum:
Ah, I see. Putting them there makes more sense than anywhere else I can think of.

I mistakenly thought you were trying to add this specifically to improve the evaluation of `TmUtest` in `eval.mc`, which we're not using at the moment (we run `generateUtest` as in the compiler [before evaluation](https://github.com/miking-lang/miking/blob/736a3cc48ea1171dea29cff0b90d587e4a1c7d81/src/main/eval.mc#L71)).

PR #683, opened 2023-02-16
Start of JVM backend

asta12:
Introduces the start of the new JVM backend. Right now it only supports print and addi.

To compile to JVM bytecode use the option `--to-jvm`. This will print out JSON which can be parsed by the java program and turned into a classfile.


stdlib/jvm/compile.mc
```diff
@@ -0,0 +1,57 @@
...
+include "string.mc"
+include "jvm/ast.mc"
+
+lang MExprJVMCompile = MExprAst + JVMAst
+
+    -- go through AST and translate to JVM bytecode
+
+    sem toJSONExpr : JVMEnv -> Expr -> JVMEnv
+    sem toJSONExpr env =
+    | TmSeq {tms = tms} -> { env with bytecode = concat env.bytecode [(createBString "LDC" (_charSeq2String tms))]}
```
larshum: You shouldn't use the `_charSeq2String` here unless you explicitly include the file in which it is defined (`javascript/utils.mc`), or it will depend on a certain declaration order. Either way, I suppose this code is intended to be temporary (not all sequences are character literals).

stdlib/jvm/compile.mc
```diff
@@ -0,0 +1,57 @@
...
+    | TmConst { val = val } -> 
+        let bc = (match val with CInt { val = val } then (createBInt "LDC" val)
+        else never)
+        in { env with bytecode = concat env.bytecode [bc] }
```
larshum: You can use `snoc env.bytecode bc` when you're adding one element at the end of the list, instead of creating a new sequence.

stdlib/jvm/compile.mc
```diff
@@ -0,0 +1,57 @@
...
+    | TmApp { lhs = lhs, rhs = rhs } -> 
+        match lhs with TmConst { val = val } then 
+            match val with CPrint _ then 
+                -- push this to stack
+                let this = { env with bytecode = concat env.bytecode [(createBApply "GETSTATIC" "java/lang/System" "out" "Ljava/io/PrintStream;")] } in
+                -- push args to stack
+                let args = (toJSONExpr this rhs) in
```
larshum, line 24: You don't need parentheses around expressions in these cases - this also applies when they're an element in a sequence literal.

stdlib/jvm/ast.mc
```diff
@@ -0,0 +1,111 @@
+
+--helper functions
+
+recursive
+    let commaSeparateHelper = lam seq. lam res.
+        if (eqi 0 (length seq)) then 
+            res
+        else if (eqi 1 (length seq)) then
+            (concat res (concat "," (get seq 0)))
```
larshum:
```suggestion
            join [res, ",", get seq 0]
```

stdlib/jvm/ast.mc
```diff
@@ -0,0 +1,111 @@
...
+        else (commaSeparateHelper (tail seq) (concat res (concat "," (head seq))))
```
larshum:
```suggestion
        else commaSeparateHelper (tail seq) (join [res, ",", head seq])
```

stdlib/jvm/ast.mc
```diff
@@ -0,0 +1,111 @@
...
+
+    sem createClass : String -> [Field] -> Function -> [Function] -> Class
+    sem createClass name fields constructor =
+    | functions -> Class {name = name, fields = fields, constructor = constructor, functions = functions}
+
+    sem createField : String -> String -> Field
+    sem createField name =
+    | t -> Field {name = name, t = t} 
+
+    -- toString JSON
```
larshum, line 81: It would be nice to have the pretty printing in a separate language fragment and/or separate file.

larshum (commented):
Overall, it looks good; I just have minor comments about the style.

Also, you're indenting with four spaces, while we typically use two spaces. I don't know how strict we are about that.

PR #693, opened 2023-03-01
Update README.md

dlunde:


README.md
```diff
@@ -78,7 +78,7 @@ user, issue:
 make install
 ```
 
-This will install `mi` to `$HOME/.local/bin` and the standard library to `$HOME/.local/lib/mcore/stdlib`, according to the [systemd file system hierarchy overview](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html). If `MCORE_STDLIB` is unset, Miking will look in this installation folder as its default library location.
+This will install `mi` to `$HOME/.local/bin` (NOTE: on some systems, you may need to _manually_ add `$HOME/.local/bin` to your `PATH`) and the standard library to `$HOME/.local/lib/mcore/stdlib`, according to the [systemd file system hierarchy overview](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html). If `MCORE_STDLIB` is unset, Miking will look in this installation folder as its default library location.
```
johnwikman: What does this comment refer to? When I read it, it sounds like the comment refers to that we need to `$HOME/.local/bin` on our path to _install_ `mi`. Would be good to change the last bit of the note as

```diff
-your `PATH`
+your `PATH` environment variable to access the `mi` command from your shell.
```

dlunde: Updated now. Also moved it to a separate sentence at the end of the paragraph.

dlunde:
I am not sure why the Mac check for this PR fails, it's just a README update. @larshum Looks like it times out?

larshum:
It seems to get stuck on that test every once in a while. No idea why or how to fix it. It should pass if we manually re-run the tests.

PR #696, opened 2023-03-05
Include info into the LR(k) parse errors

johnwikman:
Compared to before, this at least provides _some_ info. This is to be improved in the future, but adding this as a stepping stone to good parse errors.

### DIFF
--- a/stdlib/parser/lrk.mc
+++ b/stdlib/parser/lrk.mc
@@ -674,7 +674,9 @@ lang LRParser = EOFTokenParser + MExprAst + MExprCmp
     /---- Assumed to exist "public" identifiers ----/
     let #var"global: result.err" = lam v. appf1_ (recordproj_ "err" (var_ "result")) v in
     let #var"global: result.ok" = lam v. appf1_ (recordproj_ "ok" (var_ "result")) v in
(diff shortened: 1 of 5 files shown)


johnwikman:
@elegios You mentioned during the meeting today that you had some comments about this PR. Could you please put them down here?

elegios:
Essentially that it seems worth the extra effort to call a function from error.mc directly, and that that should be fairly close to the error we want down the line, since you already have expected and unexpected tokens.

johnwikman:
> Essentially that it seems worth the extra effort to call a function from error.mc directly, and that that should be fairly close to the error we want down the line, since you already have expected and unexpected tokens.

Do you mean that the parser itself terminates the program instead of returning ResultErr or that we change the error type in ResultErr? One advantage I see with ResultErr is that it delays the termination until sufficiently helpful information has been gathered. E.g. if we want to show parse errors from multiple files, if we want to display the error in a slightly different format or highlight it in a web browser, etc.

elegios:
Ah, nvm, I was looking at the code for the example languages. Ignore me. 

What is the return type of the parse function now?

johnwikman:
> Ah, nvm, I was looking at the code for the example languages. Ignore me.
> 
> What is the return type of the parse function now?

It is `Result (Info, String) (Info, String) grammarEntrypointType` where `grammarEntrypointType` is the type returned by the semantic action of the entrypoint rules/productions

PR #709, opened 2023-03-16
Add peval keyword

adamssonj:
Adds the keyword 'peval' using the keyword-maker tool. Only works with the interpreter as of now.

### DIFF
--- a/stdlib/peval/ast.mc
+++ b/stdlib/peval/ast.mc
@@ -61,11 +61,10 @@ lang PEvalAst = KeywordMaker + MExpr + MExprEq + Eval + PrettyPrint
   | TmPEval e -> 
   switch eval ctx e.e 
     case clos & TmClos {ident=i, body=b, env=envi} then
-        match peval clos with res then
+        match peval clos with res in
             match res with TmLam t then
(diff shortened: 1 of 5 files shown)


stdlib/peval/ast.mc
```diff
@@ -0,0 +1,110 @@
...
+  | TmPEval r ->
+      match lhs with TmPEval l then
+        eqExprH env free l.e r.e
+      else None ()
+
+  sem eval (ctx : EvalCtx) = 
+  | TmPEval e -> 
+  switch eval ctx e.e 
+    case clos & TmClos {ident=i, body=b, env=envi} then
+        match peval clos with res then
```
br4sco, line 64: There is a syntactic sugar `match e with p in e` = `match e with p else never` if you prefer that

stdlib/peval/ast.mc
```diff
@@ -0,0 +1,110 @@
...
+            match res with TmLam t then
+                eval ctx res -- TmClos ...
+            else res
+        else never
```
br4sco: There is a syntactic sugar `match e1 with p in e2` = `match e1 with p then e2 else never` if you prefer this

PR #713, opened 2023-04-11
Record type inference fixes

aathn:
This PR fixes a number of subtle bugs related to the inference of record types, and also improves the inference to handle more cases than previously.

As a short recap, while the type checker infers the types of records, it often does not have precise information due to the way our record matches work.  For instance, if we write
```haskell
let g = lam r. r.x
```
we know that `r` is a record with a field `x`, but we do not know its precise structure.  Accordingly, the type checker will infer a signature
```haskell
g : {x : _a ... } -> _a
```
meaning that `g` takes a record containing at least the field `x` of some type `_a`.  Importantly, this parameter type is _not polymorphic_, but represents some fixed but as-of-yet undetermined record type.  For this reason, `_a` also must not be treated polymorphically.  The current type checker tries to enforce this by marking variables as 'weak' and refusing to generalize those variables, but there were a number of bugs in the implementation.

* Weak variables could be unified with type variables from a later let binding at the same 'level'.
```haskell
let g = lam r. r.x
let h : all a. {x : a} -> a = g -- This makes 'a' escape its scope into the type of 'g' 
```

* The contents of records could sometimes be unified with type variables unsoundly.
```haskell
let f : all a. a -> Unknown = lam x. lam r. {r with s = x} -- This places 'a' inside an unknown record type
```

* The occurs check would sometimes succeed prematurely for unknown records.
```haskell
let g = lam a. lam r.
  let h : all a. a -> a -> a = lam x. lam y. y in
  ({a with x = r}, h a r) -- This makes an infinite record type
```

In addition, the type checker was quite eager to mark variables as 'weak', meaning that sometimes it would fail to produce a general type even in cases when it was completely safe.  For instance, given
```haskell
recursive let work = lam eq. lam s1. lam s2.
  match (s1, s2) with ([h1] ++ t1, [h2] ++ t2) then
    if eq h1 h2 then work eq t1 t2
    else false
  else true
end
```
it would infer `work : (_a -> _a1 -> Bool) -> [_a] -> [_a1] -> Bool` instead of the most general type `work : all a. all a1. (a -> a1 -> Bool) -> [a] -> [a1] -> Bool`.  This is now fixed, by waiting until just before generalization to mark variables inside unknown records as weak.

### DIFF
--- a/stdlib/mexpr/type-check.mc
+++ b/stdlib/mexpr/type-check.mc
@@ -249,16 +249,8 @@ end
 lang FlexTypeUnify = UnifyFields + FlexTypeAst
   sem addSorts (env : UnifyEnv) =
   | (RecordVar r1, RecordVar r2) ->
-    let f = lam acc. lam b : (SID, Type).
-      match b with (k, ty2) in
-      match mapLookup k r1.fields with Some ty1 then
-        unifyTypes env (ty1, ty2);
-        acc
-      else
-        mapInsert k ty2 acc
-    in
-    let fields = foldl f r1.fields (mapBindings r2.fields) in
-    RecordVar {r1 with fields = fields}
+    let f = lam ty1. lam ty2. unifyTypes env (ty1, ty2); ty1 in
+    RecordVar {r1 with fields = mapUnionWith f r1.fields r2.fields}
   | (RecordVar _ & rv, ! RecordVar _ & tv)
   | (! RecordVar _ & tv, RecordVar _ & rv) ->
     rv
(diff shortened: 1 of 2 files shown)


elegios:
I vaguely feel like we've discussed this before, so forgive me if I've forgotten something important, but isn't the handling of unification variables inside records unnecessarily strict, e.g.,

```
let g : all a. {x : a} -> a = lam r. r.x
```

is a valid generalization, thus we'd ideally want the inferred and generalized type to be `g : all a. _r{x : a} -> a`, where I'm using `_r{...}` to denote a weak record unification variable with (at least) the given fields?

aathn:
Yes, I think you're right.  However, the way the record variables are dealt with currently makes it difficult to implement the handling you suggest.  `_r` corresponds to an entire record, and is essentially a reference which is updated once we know the concrete type.  If we naively generalize `a`, then we run into trouble when instantiating `g`, as `_r` will be unified with a specific record type for some specific instantiation of `a`.  For instance, applying
```haskell
g {x = 5}
```
would unify `_r` with `{x : Int}`, leaving `g` with the type `all a. {x : Int} -> a`, which of course is incorrect.  To get the right result, I think we would need to treat `_r` as a row variable representing only the _unknown_ fields.  This should be doable and is something we have discussed previously, but it might take a little more effort to get right.

PR #715, opened 2023-04-18
Print aliases as well

elegios:
This PR changes the error message on unification error in the type checker to print the definitions of all aliases, to provide more relevant information. Additionally, the unification error now also `pprint`s consistently, in case multiple `Name`s are involved with the same `String` but different `Symbol`s.

image

Edit: found and fixed a bug because `TyFlex` was not supported in `cmp.mc`. Note that because a `TyFlex` is essentially mutable it's a bit unsafe to depend on `cmpType a b` to remain constant for all types `a` and `b`, but the use here is safe because we never call `unify` after `cmpType`.

### DIFF
--- a/stdlib/mexpr/type-check.mc
+++ b/stdlib/mexpr/type-check.mc
@@ -106,6 +106,15 @@ lang FlexTypeAst = VarSortAst + Ast
     else (acc, ty)
 end
 
+lang FlexTypeCmp = Cmp + FlexTypeAst
+  sem cmpTypeH =
+  | (TyFlex l, TyFlex r) ->
+    -- NOTE(vipa, 2023-04-19): Any non-link TyFlex should have been
+    -- unwrapped already, thus we can assume `Unbound` here.
+    match (deref l.contents, deref r.contents) with (Unbound l, Unbound r) in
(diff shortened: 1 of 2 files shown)


stdlib/mexpr/type-check.mc
```diff
@@ -188,11 +189,33 @@ lang Unify = MExprAst + FlexTypeAst + PrettyPrint
...
+    match getTypeStringCode 0 pprintEnv env.foundType with (pprintEnv, found) in
+    recursive let collectAliases : Map Type Type -> Type -> Map Type Type
+      = lam acc. lam ty.
+        match ty with TyAlias x then
+          let acc = mapInsert x.display x.content acc in
+          collectAliases (collectAliases acc x.display) x.content
+        else sfold_Type_Type collectAliases acc ty
+    in
+    let aliases = collectAliases (mapEmpty cmpType) env.expectedType in
+    let aliases = collectAliases (mapEmpty cmpType) env.foundType in
```
aathn: `aliases` appears to be shadowed here - should the second line say `collectAliases aliases env.foundType`?

elegios: Good catch, fixed

PR #717, opened 2023-04-22
File namespaces

### LINKED ISSUE #716: File namespace proposal
To allow Miking code to refer to files outside of both the standard library and the current project, this proposal introduces a concept of *file namespaces*. This is a temporary solution, as we anticipate that a proper solution based on modules and packages or similar will require a more in-depth design discussion.

<a id="orgaf15e96"></a>

# Basic idea

File namespaces are simply names associated with directories, such as the following.

```bash
stdlib=/path/to/miking/stdlib
mypkg=/path/to/mypkg/src
```

We extend the `include` syntax to optionally accept a prefix of file namespace followed by a double colon (or some other delimiter, to be discussed).

```mcore
include "stdlib::error.mc"
include "mypkg::linear-time-travelling-salesman.mc"
```

In this case, the first include would resolve to the file `/path/to/miking/stdlib/error.mc`, and the second to `/path/to/mypkg/src/linear-time-travelling-salesman.mc`.

<a id="org9230320"></a>

# Specifying namespaces

We replace the `MCORE_STDLIB` environment variable with a ~`MCORE_LIBRARY_PATH`~ `MCORE_LIBS` variable containing the namespaces of installed libraries. This variable will be a colon-separated list of pairs `name=path` specifying the association between namespace and filesystem path.

For the example above, we would use the following.

```bash
export MCORE_LIBS="stdlib=/path/to/miking/stdlib:mypkg=/path/to/mypkg/src"
```

<a id="orgf21d2d5"></a>

# Why namespaces?

We could adopt an even simpler approach than the one above, where we simply expand `MCORE_STDLIB` to be a list of paths, and try each path in turn when resolving includes.

For instance, we could have `MCORE_STDLIB=/path/to/miking/stdlib:/path/to/mypkg/src` and then directly refer to the paths as follows.

```mcore
include "error.mc"
include "linear-time-travelling-salesman.mc"
```

However, we run into trouble if the same file, say `common.mc` occurs in both `stdlib` and `mypkg`. At this point, we would have no way of knowing which file was intended, and would either have to choose one arbitrarily or throw an error. Furthermore, this conflict cannot be easily resolved by the end user, as it requires changing the file structure of one of the packages (and packages cannot be expected to account for the possibility of other packages with the same file names).

Hence, we need some way for the user to differentiate between included files, and this is what namespaces give us.

<a id="org5d99fdd"></a>

# Special namespaces and local files

Always referring to namespaces explicitly may be tedious in some cases (and would require rewriting a lot of the existing code). For one, we would like to be able to refer to files in the standard library without explicitly naming `stdlib`. In addition, we might want to refer to files in relation to the project root (to avoid having to write paths like `../../common.mc`).

To account for this, I propose we use two special namespaces, `root` and `stdlib`, which are always implicitly available. [In the long run, this warrants more discussion, as it may be better to be explicit than ambiguous]

For instance, if I set `root=/path/to/mypkg/src` and `stdlib=/path/to/miking/stdlib`, I could write the following.

```mcore
include "error.mc"
include "linear-time-travelling-salesman.mc"
```

Then, the two includes would resolve to `/path/to/miking/stdlib/error.mc` and `/path/to/mypkg/src/linear-time-travelling-salesman.mc`, even from a subdirectory of `mypkg/src`.

In case the same file exists in more than one of {locally,project root,standard library}, the user would be required to disambiguate by writing the namespace, or prefixing the file name with `./` or `../` (and an error would be thrown otherwise).

```mcore
include "./error.mc"
include "root::error.mc"
include "stdlib::error.mc"
```

Supposing the user is in directory `/path/to/mypkg/src/subdir`, the three inclusions would resolve to `/path/to/mypkg/src/subdir/error.mc`, `/path/to/mypkg/src/error.mc` and `/path/to/miking/stdlib/error.mc`, respectively.

The `root` namespace would be treated specially, to ensure it always refers to the root of the namespace being handled. For instance, consider the following case, where we have four files `mypkgA/src/file{1,2}.mc`, `mypkgB/src/file{1,2}.mc`.

`mypkgA/src/file1.mc`

```mcore
include "mypkgB::file1.mc"
include "root::file2.mc"
```

`mypkgB/src/file1.mc`

```mcore
include "root::file2.mc"
```

Here, we would like `root::file2.mc` to resolve to `mypkgA/src/file2.mc` in the first file, but `mypkgB/src/file2.mc` in the second file. Hence, the inclusion parsing would need to update `root` to the value of `mypkgB` when passing through the inclusion of `mypkgB::file1.mc`.

The `root` namespace would be set through a separate variable `MCORE_ROOT` to make it easy to update on a per-folder basis using something like [direnv](https://direnv.net).

<a id="org0b7ca9f"></a>

# Problems with the approach

Although this approach avoids the most basic name clashes mentioned above, it still has some pretty serious limitations in terms of scaling.

Suppose we have the following scenario:

-   Package A depends on package C, referring to it as `pkg`.
-   Package B depends on package D, referring to it as `pkg`.
-   I want to depend on both package A and package B.

In this scenario, we are essentially out of luck, and need a more sophisticated approach. However, this name clash is on the level of *namespaces*, as opposed to *file names*, making it slightly less serious. As long as we are consistent about referring to every library with a unique name, and the ecosystem is small enough to make this reasonably easy, we should have no issues.

aathn:
This PR implements #716, except for the `root` namespace, as discussed during the last meeting.

Resolves #716.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,9 @@ and then run it using the command:
 build/mi hello.mc
 ```
 
-To help Miking find its standard library, you should define the
-environment variable `MCORE_STDLIB` to be the path to `stdlib`,
-for example by running the following:
+To help Miking find its standard library, you should define the environment variable `MCORE_LIBS` with an entry for `stdlib` pointing to the appropriate path, for example by running the following:
 
-    cd stdlib; export MCORE_STDLIB=`pwd`; cd ..;
+    export MCORE_LIBS=stdlib=`pwd`/stdlib
 
 To install the compiler along with the standard library for the current
 user, issue:
@@ -78,7 +76,7 @@ user, issue:
 make install
 ```
 
-This will install `mi` to `$HOME/.local/bin` and the standard library to `$HOME/.local/lib/mcore/stdlib`, according to the [systemd file system hierarchy overview](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html). If `MCORE_STDLIB` is unset, Miking will look in this installation folder as its default library location. NOTE: on some systems (e.g., macOS), you need to _manually_ add `$HOME/.local/bin` to your `PATH` environment variable to access the `mi` command from your shell.
+This will install `mi` to `$HOME/.local/bin` and the standard library to `$HOME/.local/lib/mcore/stdlib`, according to the [systemd file system hierarchy overview](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html). If `MCORE_LIBS` is unset, Miking will look in this installation folder as its default library location. NOTE: on some systems (e.g., macOS), you need to _manually_ add `$HOME/.local/bin` to your `PATH` environment variable to access the `mi` command from your shell.
 
 Conversely, to uninstall, issue:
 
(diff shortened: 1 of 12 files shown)


dlunde:
Great! Looking forward to having this merged

PR #726, opened 2023-05-08
Type checker improvements

aathn:
This PR makes a number of improvements to the type checker, fixing several bugs and previously known soundness problems.

- Fix a bug where bound type variables could escape their scope.
  The level-based approach to scope checking ensures type variables do not escape their lexical scope (i.e., their let binder), but there was no check ensuring that a unification does not bring a type variable out of its forall binder _within_ the type. For example, given the type `(all b. b -> %a) -> %a` (where `%a` denotes a unification variable), there was nothing stopping `%a` from being unified with `b`, giving the type `(all b. b -> b) -> b` where the last `b` is out of scope.  In fact, we never want to let flexible variables be unified with rigid variables whose binder is not outermost in the type.
- Fix a bug where type variables in recursive lets could escape to one of the parallel binders without being generalized.
  For example, in the program below, the signature of `none` would be inferred to `() -> Option a`, where `a` is out of scope.  This PR fixes the issue by making sure to quantify those variables as well (giving `none` type `all a. () -> Option a` as expected).
```haskell
recursive 
let f : all a. () -> Option a = none
let none = lam. None ()
end
```
- Fix a bug where annotations on polymorphic terms would not be checked correctly.
  The issue was that the type checker would always strip the `all` binders of the annotation before matching it with the inferred type.  Thus, when the inferred type had leading foralls, the check would always fail.  For instance, in the below code, we would try to unify `a -> a` with `all a. a -> a`, giving an error.
```haskell
let id = lam x. x
let f : all a. a -> a = #frozen"id"
```
- Implement the value restriction.
  The value restriction means that polymorphic types may only be inferred for terms with no visible side effects, and is typical of ML-like systems (including OCaml and Standard ML).  For example, in the program below, `x` would get the polymorphic type `all a. a ->  a` as usual, but `y`, which contains a function application, gets type `_a -> _a`, where `_a` is a _weak_ variable representing some fixed but as-of-yet unknown type.
```haskell
let x = lam x. x
let y = x x
```
The typical example where dropping the value restriction allows unsound behavior is the following program where a reference to an empty list is created.
```haskell
let r = ref [] in
modref r [true];
addi (head (deref r)) 1
```
Without the value restriction, `r` gets type `all a. Ref [a]`, meaning that we can treat its contents as having any type, allowing unsound operations like writing a bool and reading an int as in the example.  With the value restriction, `r` instead gets type `Ref [_a]`, where `_a` is determined to `Bool` at the call to `modref`, resulting in a consequent type error at the call to `addi`.
- Add scope check for type constructors.
  Previously, programs like
```haskell
let x =
  type Foo in
  con Foo : () -> Foo in
  Foo ()
```
or
```haskell
let y = ref []
type Foo
con Foo : () -> Foo
modref y (Foo ())
```
were possible.  In the first, a value of the local type `Foo` is returned, giving `x` the type `Foo` no longer in scope.  In the second, a value of type `Foo` is written into `y`, which was defined before the declaration of `Foo`, so that `Foo` again escapes its scope. This PR fixes the issue, using a level-based approach similar to how type variable scopes are handled.  We increment the current level whenever a type `X` is declared, , and check that unification variables are never unified with a type constructor of higher level.  We also check that no values of type `X` are returned from the `inexpr` of `X`.
In particular, this means that the following program is no longer valid:
```haskell
type Foo in
con Foo : () -> Foo in
Foo ()
```

### DIFF
--- a/stdlib/arg.mc
+++ b/stdlib/arg.mc
@@ -99,7 +99,8 @@ let argHelpOptions_general : all a.
 
 
 
-let argHelpOptions = argHelpOptions_general argHelpOptions_defaults
+let argHelpOptions = lam opConfig.
+  argHelpOptions_general argHelpOptions_defaults opConfig
 
 
 -- argument value conversion --
--- a/stdlib/assoc.mc
+++ b/stdlib/assoc.mc
@@ -163,25 +163,25 @@ mexpr
 let traits = {eq = eqi} in
 
 let length = assocLength in
-let lookup = assocLookup traits in
-let lookupOrElse = assocLookupOrElse traits in
+let lookup = lam x. assocLookup traits x in
+let lookupOrElse = lam x. assocLookupOrElse traits x in
 let lookupPred = assocLookupPred in
 let any = assocAny in
 let forAll = assocAll in
-let insert = assocInsert traits in
-let seq2assoc = seq2assoc traits in
-let assoc2seq = assoc2seq traits in
-let mem = assocMem traits in
-let remove = assocRemove traits in
-let keys = assocKeys traits in
-let values = assocValues traits in
-let map = assocMap traits in
-let mapWithKey = assocMapWithKey traits in
-let fold = assocFold traits in
-let foldOption = assocFoldlM traits in
-let mapAccum = assocMapAccum traits in
-let mergePreferLeft = assocMergePreferLeft traits in
-let mergePreferRight = assocMergePreferRight traits in
+let insert = lam x. assocInsert traits x in
+let seq2assoc = lam x. seq2assoc traits x in
+let assoc2seq = lam x. assoc2seq traits x in
+let mem = lam x. assocMem traits x in
+let remove = lam x. assocRemove traits x in
+let keys = lam x. assocKeys traits x in
+let values = lam x. assocValues traits x in
+let map = lam x. assocMap traits x in
+let mapWithKey = lam x. assocMapWithKey traits x in
+let fold = lam x. assocFold traits x in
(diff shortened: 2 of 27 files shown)


johnwikman:
In the first example of value restriction:
```ml
let x = lam x. x
let y = x x
```
Why cannot `y` get the type `all a. a -> a`? `x` has already been identified as `all a. a -> a`, so when `x` is applied to itself surely we would be able to not have to worry about any side effects, since we know what the function `x` does?

aathn:
Well, deciding whether `x` has side effects requires a more sophisticated analysis in general.  The value restriction is a very simple way to ensure type safety which has proven effective in practice.  OCaml's value restriction is slightly more sophisticated than the basic one I have implemented, and would allow inferring a polymorphic type for expressions like `(lam y. print "ok"; y) []`.  However, it still fails on the example:
```ocaml
# let x = fun x -> x;;
val x : 'a -> 'a = <fun>
# let y = x x;;
val y : '_weak1 -> '_weak1 = <fun>
```

aathn:
Well, deciding whether `x` has side effects requires a more sophisticated analysis in general.  The value restriction is a very simple way to ensure type safety which has proven effective in practice.  OCaml's value restriction is slightly more sophisticated than the basic one I have implemented, and would allow inferring a polymorphic type for expressions like `(lam y. print "ok"; y) []`.  However, it still fails on the example:
```ocaml
# let x = fun x -> x;;
val x : 'a -> 'a = <fun>
# let y = x x;;
val y : '_weak1 -> '_weak1 = <fun>
```

PR #730, opened 2023-05-10
Symbolize refactor and other changes

aathn:
This PR refactors `mexpr/symbolize.mc` to (1) reduce code duplication by a fair amount ~and (2) make it extensible in preparation of writing `mlang/symbolize.mc`~.  It also includes some other changes, listed below.

Note that this PR is based on #726, so the diff shown is not accurate (look  [here](https://github.com/miking-lang/miking/pull/730/files/8b7fd459b313ed5365fbb02c28f0eb7a7e485df7..049323263848b2c2b548e89cacbc08fec9efbdba) for the actual diff).

Changes:
- Refactor `symbolize.mc`.
- Move type alias resolution to `type-check.mc`.  The original motivation for putting it in `symbolize` was for convenience and efficiency, since it meant that alias resolution could be done in the same traversal as symbolization and the type checker didn't need to track and resolve aliases.  However, I think it is more logical to have a separation of concerns where the symbolization is only concerned with simple name resolution, and the type checker is responsible for the workings of type declarations.  The type checker will need to track type constructors anyways when we add kind checking, so it seems reasonable to make this change now.
- Rename `tyIdent` field of `TmLam` to `tyParam`.  `tyIdent` is used in `TmExt`, `TmType`, `TmConDef` to denote type declarations (part of the surface syntax), whereas only for `TmLam` it was a typechecker-added tag.
- Remove all remaining uses of `typeAnnot` except for the one in `ocaml/mcore.mc` which is still necessary for compilation.

These changes seem to make the first stage of bootstrapping about 10 seconds slower, I haven't looked into the details yet.

### DIFF
--- a/src/main/tune.mc
+++ b/src/main/tune.mc
@@ -13,7 +13,7 @@ include "tuning/instrumentation.mc"
 include "tuning/tune.mc"
 
 lang MCoreTune =
-  BootParser + MExprTypeAnnot + MExprTypeCheck +
+  BootParser + MExprTypeCheck +
   MExprHoles + MExprHoleCFA + NestedMeasuringPoints + DependencyAnalysis +
   Instrumentation + MExprTune
 end
--- a/stdlib/futhark/generate.mc
+++ b/stdlib/futhark/generate.mc
@@ -8,7 +8,7 @@ include "mexpr/ast-builder.mc"
 include "mexpr/cmp.mc"
 include "mexpr/pprint.mc"
 include "mexpr/symbolize.mc"
-include "mexpr/type-annot.mc"
+include "mexpr/type-check.mc"
 include "pmexpr/ast.mc"
 include "pmexpr/utils.mc"
 
@@ -533,7 +533,7 @@ lang FutharkGenerate = FutharkToplevelGenerate + MExprCmp
 end
 
 lang TestLang =
-  FutharkGenerate + FutharkPrettyPrint + MExprSym + MExprTypeAnnot
+  FutharkGenerate + FutharkPrettyPrint + MExprSym + MExprTypeCheck
 end
 
 mexpr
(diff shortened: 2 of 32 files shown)


david-broman:
I would like that we discuss and analyze the performance effects before merging this PR.

aathn:
I have tried to analyze the performance loss as best I can, and reached the following conclusions:
- Moving alias resolution to `type-check.mc` makes the first stage of bootstrapping about 6 seconds slower (on my machine), presumably because it adds an additional traversal of each type to resolve the aliases.
- The refactorization of `symbolize.mc` itself makes the first stage of bootstrapping about 6 seconds slower (on my machine), presumably because

(1) it increases the number of function calls (to reduce code duplication), giving an overhead when interpreting;
(2) it adds some unnecessary record updates.
  For example, the previous code might look something like:
```haskell
if nameHasSym t.ident then
  TmVar t
else
  let ident = match mapLookup (nameGetStr t.ident) env.varEnv with Some ident then ident else ... in
  TmVar {t with ident = ident}
```
  Here, the record update `{t with ident = ident}` occurs only if the name was actually updated.  On the other hand, the new code takes the form
```haskell
let ident = getSymbol env.varEnv t.ident in
TmVar {t with ident = ident}
```
  where `getSymbol` looks up `ident` only if it didn't already have a symbol.  This makes the record update be performed regardless if the name was updated or not.

For the former point (moving of alias resolution), I think the performance hit might be inevitable, as we will need to do some kind checking in the future anyways, requiring an extra traversal of the types.  For the latter point (symbolize refactorization) I would argue that the increase in code quality is worth it, but this is perhaps something we should discuss.

Note that the later stages of bootstrapping do not seem noticeably affected.

aathn:
Some measurements:
```
On develop:

Bootstrapping the Miking compiler (1st round, might take a few minutes)

real	2m12.364s
user	2m11.352s
sys	0m1.002s


On #726:

Bootstrapping the Miking compiler (1st round, might take a few minutes)

real	2m20.974s
user	2m19.948s
sys	0m1.020s


On #730:

Bootstrapping the Miking compiler (1st round, might take a few minutes)

real	2m33.200s
user	2m32.308s
sys	0m0.886s


On #726 + only symbolize refactorization:

Bootstrapping the Miking compiler (1st round, might take a few minutes)

real	2m26.742s
user	2m25.553s
sys	0m1.168s


On #726 + only alias change:

Bootstrapping the Miking compiler (1st round, might take a few minutes)

real	2m27.324s
user	2m26.289s
sys	0m1.024s

```

PR #746, opened 2023-05-30
Updates related to ProbTime (RTPPL)

larshum:
This PR includes the declaration of externals in ProbTime (RTPPL), which are defined in an OCaml library [outside of Miking](https://github.com/larshum/rtppl/tree/master/rtppl-support). It also consists of a few other changes that were added as part of the implementation effort:
* A pass for "demoting" (I don't know a good name for this) recursive bindings to let bindings or separate independent recursive bindings as far as possible (based on a dependency graph of the bindings). I use this in DPPL to ensure Daniel's optimizations apply to my code.
* Enables `ref` and `deref` (but notably not `modref`) in CFA transformation.
* Bugfix in extraction, where it would not extract a type referred to only through the type field of an external (which we do extract).
* Bugfix in the type erasure in lambda lifting.
* A new function, `lowerBoundBinarySearch`, for finding the first element that returns a non-negative value for a provided function. We assume the sequence `s` is ordered such that for the function `f`, it holds that `map f s` is increasing.


dlunde:
> A pass for "demoting" (I don't know a good name for this) recursive bindings to let bindings or separate independent recursive bindings as far as possible (based on a dependency graph of the bindings). I use this in DPPL to ensure Daniel's optimizations apply to my code.

I believe I did the same thing in Miking DPPL in my most recent PR, check `removeRedundantRec` in `extract.mc`. I would suggest that we use your version in Miking instead, so you could replace the `removeRedundantRec` from Miking DPPL in your other PR.

larshum:
> > A pass for "demoting" (I don't know a good name for this) recursive bindings to let bindings or separate independent recursive bindings as far as possible (based on a dependency graph of the bindings). I use this in DPPL to ensure Daniel's optimizations apply to my code.
> 
> I believe I did the same thing in Miking DPPL in my most recent PR, check `removeRedundantRec` in `extract.mc`. I would suggest that we use your version in Miking instead, so you could replace the `removeRedundantRec` from Miking DPPL in your other PR.

Yes. I am removing that function in my [Miking DPPL PR](https://github.com/miking-lang/miking-dppl/pull/136/files#diff-f22e89927b4710804cc431e1d07c023999863de58a07f537fae2816b0edf3215) and using this one instead.

stdlib/mexpr/cfa.mc
```diff
@@ -1371,9 +1371,9 @@ end
 -- probably be added.
 lang RefOpCFA = CFA + ConstCFA + RefOpAst
   sem generateConstraintsConst info ident =
-  -- | CRef _ -> []
+  | CRef _ -> []
+  | CDeRef _ -> []
```
dlunde: This is not correct in the general case, now we do not track abstract values (i.e., abstract lambdas and constants) flowing through references. We need to do something similar as in the sequence operations (but much simpler as references only contain one item). By the way, what is the purpose of having references without `modref`? I recall we discussed it, but I have forgotten.

dlunde:
> Yes. I am removing that function in my [Miking DPPL PR](https://github.com/miking-lang/miking-dppl/pull/136/files#diff-f22e89927b4710804cc431e1d07c023999863de58a07f537fae2816b0edf3215) and using this one instead.

Ah, great. Should have read the other PR description before replying

stdlib/mexpr/cfa.mc
larshum: I see. I'm storing a continuously updated timestamp, which I compute other relative timestamps to, and this can also take place inside a model. However, I know I will never use a `modref` inside a model as the time updates always take place outside of the model code. It's something along the lines of
```
let currTime = ref (getWallTime ()) in
...
let model = lam.
  let absTime = getWallTime () in
  let relTime = timeDiff absTime (deref currTime) in
  ...
in
...
infer ... (model ())
```
Because of the lambda lifting, I believe the `currTime` reference will be captured as a parameter to `model`. I could thread along the timestamp(s) manually (without using references) to all functions, but that's a lot of extra work, so I did it this way mainly to save time.

dlunde: Ah, right, now I remember. I can have a look tomorrow at implementing a correct version of `ref`/`deref` in the CFA framework.

PR #752, opened 2023-06-16
Fix incorrect print of negative integers in `pprint.mc`

dlunde:
`mexpr/pprint.mc` is currently printing negative integers as `-i`, which does not parse. This PR fixes this in the same way as for printing floats (`negi i` whenever the corresponding literal is negative).

### DIFF
--- a/stdlib/mexpr/pprint.mc
+++ b/stdlib/mexpr/pprint.mc
@@ -667,7 +667,9 @@ end
 
 lang IntPrettyPrint = IntAst + IntPat + ConstPrettyPrint
   sem getConstStringCode (indent : Int) =
-  | CInt t -> int2string t.val
+  | CInt t ->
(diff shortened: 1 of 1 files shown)


johnwikman:
Is this also reflected in the boot pretty printer?

I think that the better change would be to have parsing for negative numbers as well. Do we handle negative int's and float's at the moment?

dlunde:
Good question. No, it does not seem like either negative ints or floats are printed correctly in `boot/lib/pprint.ml`. Do we still use the pretty printing in boot?

If I recall correctly, we actively chose not to have negative integer and float literals in MExpr. I don't remember the motivation, however.

johnwikman:
There is a signed_integer option in `lexer.mll`, but it is not used. Regular languages the use the unary `-` operator to handle negative integers, but that should not be necessary since we don't have infix operators in miking. It also makes the code more readable to have lexing of negative numbers.

Issue #677, opened 2023-01-27
Controlling the pretty-print format for failing unit tests


larshum:
When a `utest` fails, it pretty-prints the contents of the left- and right-hand side using an expression printer (`boot`) or generated pretty-print functions for a particular type (`mi`). In most cases, this works fine, so it would be annoying if all unit tests required a user-defined pretty-print function. However, I commonly end up in situations where the default pretty printer could be (far) better. The result is that I end up re-running the program without tests but with a manually inserted print of what I'm testing to see what went wrong.

For example, consider the outputs at the bottom of this text (click to expand, they're pretty big), where I compare two almost equivalent ASTs consisting of one let-binding and an application. I find this completely unreadable, and I'm sure you agree. So what I end up doing in situations like this is to add a `printLn (expr2str lhs)` and then run without `--test`, which in this case would yield something like
```
let f =
  lam x.
    addi
      x
      1
in
f
  3
```

Furthermore, we want the flexibility to control how the `utest` arguments are pretty-printed. I think the format of `expr2str` works in most cases, but when testing the type-checker, this format would not be sufficient, as it does not show the underlying types. On the other hand, printing the types would be annoying when doing tests for the symbolizer, which runs before the type-checker. Therefore, I argue that we want the ability to control how the arguments of a `utest` are pretty-printed when a test fails.

However, I don't have any concrete suggestions for the syntax. The main problem is that we currently allow the left and right arguments to have different types when an equality function is provided. Therefore, we have to be able to specify which pretty-print function to use for which argument, and I cannot think of any syntax for that which I like. To get the discussion started, I'll propose the following syntax
```
utest a with b using {eq = c, lpp = d, rpp = e} in
...
```
where all fields of the record are optional.

<details id="boot output">
<summary>Boot output</summary>

```
/private/tmp/example.mc: 
 ** Unit test FAILED: FILE "/private/tmp/example.mc" 5:0-15:12 **
    LHS: LetAst_TmLet {tyAnnot = (UnknownTypeAst_TyUnknown {info = (NoInfo ())}),
              tyBody = (UnknownTypeAst_TyUnknown {info = (NoInfo ())}),
              inexpr = (AppAst_TmApp {info = (NoInfo ()),
                                      rhs = (ConstAst_TmConst {info = (NoInfo (
                                                               )),
                                                               val = (IntAst_CInt {
                                                               val = 3}),
                                                               ty = (UnknownTypeAst_TyUnknown {
                                                               info = (NoInfo (
[...]

PR #778, opened 2023-08-16
Custom pretty printing in utests

br4sco:
This PR adds the option to supply a function that specifies the output string for the left-hand and right-hand sides of failing utests. The semantics and syntax are best described with an example. The following code snippet:

```
...
mexpr
let eq : Int -> String -> Bool = lam l. lam r. eqi 0 (string2int r) in
let toString : Int -> String -> String = lam l. lam r.
  join ["left hand: ", int2string l, ", right hand: \"", r, "\""]
in
utest 0 with "1" using eq in               -- Failing
utest 0 with "1" using eq else toString in -- Failing
utest 0 with "0" using eq else toString in -- Passing
()
```

will on execution output:

```
 ** Unit test FAILED: <fileinfo> **
    LHS: 0
    RHS: ['1']
    Using: eq

 ** Unit test FAILED: <fileinfo> **
left hand: 0, right hand: "1"
    Using: eq
.ERROR! 1 successful tests and 2 failed tests.
```
 
The else function has the signature `a -> b -> String` and is only applied to the left-hand side and the right-hand side if a test fails (the standard formatting is omitted in this case). Currently, you always have to specify the custom equality function if you want to also specify the else function due to how the bootparser is implemented. We can relax this later on if we want to.
This PR addresses https://github.com/miking-lang/miking/issues/677.

You can find the above example in [test/examples/utest/utest-with-onfail.mc](https://github.com/miking-lang/miking/commit/92f56c8491aa75bc523fcf98cc878e64b35f5040#diff-af7d058521a4b491cc049871147b55496c02287e98f89adae9d1e658a1622dbe).


johnwikman:
Looks good! But I would also like to have the option to get `LHS: ... RHS: ...` formatting for free, i.e. that I just provide string conversion on either side:
```
let toString : Int -> String -> (String, String) =
    lam l. lam r. (int2string l, concat "strint: " r)
in
utest 0 with "1" using eq onfail toString in
-- This would output
-- ** Unit test FAILED: <fileinfo> **
--    LHS: 0
--    RHS: strint: 1
--    Using: eq

-- or even more convenient
utest 0 with "1" using eq onfail (int2string, concat "strint: ") in
```

br4sco:
I propose that we implement that functionality as a library function `defaultOnfail : all a. all b. (a -> String) -> (b -> String) -> a -> b -> String` rather than adding more keywords and variations on the `utest` term. Or do you want to replace the current signature so that "RHS:" and "LHS:" are always printed?

johnwikman:
Yes that sounds like a good solution. Can probably be part of a separate PR then

br4sco:
Changed `onfail` to `else`.

PR #784, opened 2023-09-14
Fixes to conform to the JSON standard

vsenderov:
The JSON specification does not allow for trailing or leading 0's such as `1.` or `.1`.  For this reason I have changed the `json2string` function on the string level and added tests.

Caveats:
- a better fix may be to write a new intrinsic to use instead of `float2string` that makes use of the OCaml Printf module and specify the required output there;
-  unsure if this works for locales that do not have dot as the decimal separator and use `,` instead.

This is an urgent PR as JSON compatibility is critical for TreePPL.


vsenderov:
I made the small changed requested during the Miking meeting; this should be okay to merge now.

PR #787, opened 2023-09-20
JSON does not support `nan`, `-inf`, `inf`

vsenderov:
JSON does not support `nan`, `-inf`, `inf`.
Unofficially one can use the JavaScript `NaN`, `Infinity`, and `-Infinity`.
This PR implements that.

### DIFF
--- a/stdlib/json.mc
+++ b/stdlib/json.mc
@@ -7,6 +7,7 @@ include "map.mc"
 include "string.mc"
 include "option.mc"
 include "tensor.mc"
+include "math.mc"
 
 type JsonValue
 con JsonObject: Map String JsonValue -> JsonValue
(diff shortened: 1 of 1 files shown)


johnwikman:
Is there precedence for this extension? A risk we should not take is to start generating JSON files that are unparsable by other established JSON parsers.

vsenderov:
I am converting this to draft, as I don't have enough time to implement before Oct 15.

I did talk to vipa and I think the situation is as follows.

Right now the serializer does produce `nan`, `-inf` and `inf`.  This is not legal JSON according to @jkudlicka.

Using `NaN, `Infinity` and `-Infinity` _is_ supported by the Python parser, but is as far as I understand semi-official.

The proper behavior is a flag, which tells the serializer to use semi-official style or fail.  We should never produce `nan`, etc.

I will get back to this after Oct 15.

PR #802, opened 2023-11-07
ProbTime external updates

larshum:
This PR updates the external declarations used by ProbTime. Also, it improves the performance of the `toRope` function in `stdlib/seq.mc` in two ways:
1. Do nothing if the sequence representation is already a rope.
2. Traverse the list more efficiently and, finally, force the rope to flatten to ensure constant-time random access.

Update: This PR now also contains #806


larshum:
The most recent failure on MacOS is due to errors with Python rather than due to my changes in this PR. I'm making an empty commit to force them to re-run.

PR #805, opened 2023-11-12
Add run_all, timeout, and constant folding to the microbenchmarks

br4sco:
This PR adds the miking compiler with optimizations (currently only constant folding) to the micro-benchmark entries. Moreover, it adds an optional timeout and a script to run all micro-benchmarks conveniently. It also improves the argument parsing of the benchmark runner and fixes a minor bug.

### DIFF
--- a/test/microbenchmark/run
+++ b/test/microbenchmark/run
@@ -1,2 +1,2 @@
 #!/bin/sh
-ocaml -I +unix unix.cma run.ml $1 $2 $3 $4 $5
+ocaml -I +unix unix.cma run.ml "$@"
--- a/test/microbenchmark/run
+++ b/test/microbenchmark/run
@@ -1,2 +1,2 @@
 #!/bin/sh
-ocaml unix.cma run.ml $1 $2 $3 $4 $5
+ocaml -I +unix unix.cma run.ml $1 $2 $3 $4 $5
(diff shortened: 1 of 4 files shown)


larshum:
The only test that I found failed for that reason was `split_rope`, and that's because it's using `Rope.create` instead of `Rope.create_array`. I would be fine with renaming it to `Rope.create` since there's only one version now, but then we'd have to update other benchmarks and also the code in other places.

Also, I had trouble running the benchmarks at first because the `--enable-constant-fold` flag isn't supported by the compiler in this branch.

br4sco:
> Also, I had trouble running the benchmarks at first because the `--enable-constant-fold` flag isn't supported by the compiler in this branch.

Ah right, I should say that this depends on https://github.com/miking-lang/miking/pull/800

br4sco:
> The only test that I found failed for that reason was `split_rope`, and that's because it's using `Rope.create` instead of `Rope.create_array`. I would be fine with renaming it to `Rope.create` since there's only one version now, but then we'd have to update other benchmarks and also the code in other places.

Would it not be fine to just use `Rope.create_array` in the micro-benchmark?

larshum:
> Would it not be fine to just use Rope.create_array in the micro-benchmark?

Yes, that's the easiest way to fix it.

PR #828, opened 2024-03-21
Adds a mapFindLower and mapFindUpper to map.mc

br4sco:
This PR:
- Adds two new functions to `map.mc`, `mapFindLower`, and `mapFindUpper` that find the nearest upper/lower key and associated value w.r.t. to a given key.
- Adds a new function, `mapSingleton`, which creates a singleton map.
- Re-organizes `map.mc` to clarify the map interface.

### DIFF
--- a/stdlib/avl.mc
+++ b/stdlib/avl.mc
@@ -205,6 +205,36 @@ lang AVLTreeImpl
     else if gti d 0 then avlLookup cmp k t.r
     else Some t.value
 
+  -- `avlFindUpper cmp k t` returns (k', v) in the tree `t`, where k' is the
+  -- minimum key in the tree and k≤k', according to `cmp`. Returns `None ()` if
+  -- no such key k' exists in the tree.
+  sem avlFindUpper : all k. all v. (k -> k -> Int) -> k -> AVL k v -> Option (k, v)
+  sem avlFindUpper cmp k =
+  | Node t ->
+    let d = cmp k t.key in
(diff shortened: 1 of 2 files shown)


johnwikman:
LGTM, but this PR got me thinking, why are we limiting ourselves to `k -> k -> Int` comparisons for maps? Why not have something like `k1 -> k2 -> Int` if we want to compare different types? Feels like this feature would be great if you are looking for something in an integer keyed map and want to find the number closest to some float.

br4sco:
> why are we limiting ourselves to `k -> k -> Int` comparisons for maps?

That would be nice. It is not immediately clear to me how you can get that expressiveness to work with AVL trees but maybe I'm not thinking hard enough.

johnwikman:
> > why are we limiting ourselves to `k -> k -> Int` comparisons for maps?
> 
> That would be nice. It is not immediately clear to me how you can get that expressiveness to work with AVL trees but maybe I'm not thinking hard enough.

I was thinking that you could have something like
```
let mapFindLowerOn : all k. all v. (k -> Int) -> Map k v -> Option (k, v) =
  ...

mexpr

let m : Map Int String = ... in

let x = 3.5 in

switch mapFindLowerOn (lam k. cmpf x (int2float k)) m
case Some (k, v) then ...
case _           then ...
end
```

I think this kind of custom comparison really only makes sense with this "find the best option" kind of operation when there might not be a direct conversion from lookup values to key values. For all other map operations you expect perfect matches, so a direct conversion should be possible.

PR #842, opened 2024-05-20
MLang Pipeline

marten-voorberg:
This PR aims to add a pipeline for MLang compilation that does not use any of the `boot` transformations for compiling MLang to MExpr but does this in MLang itself.


src/main/mi.mc
```diff
@@ -68,21 +68,21 @@ type SubConfig = {
...
+-- in
 
 -- Commands map, maps command strings to functions.
 let commandsMap : [(String, SubConfig)] = map (lam c : SubConfig. (c.name, c))
 [ {name = "run", cmd = run, config = optionsConfig}
 , {name = "eval", cmd = eval, config = optionsConfig}
 , {name = "compile", cmd = compile, config = optionsConfig}
 , {name = "tune", cmd = tune, config = tuneOptionsConfig}
-, {name = "syn", cmd = parserGen, config = optionsConfig}
+-- , {name = "syn", cmd = parserGen, config = optionsConfig}
```
elegios: This should be re-included before merge, but I'm guessing there's something that needs updating there because of AST changes or something?

src/main/options-config.mc
```diff
@@ -134,5 +134,15 @@ let optionsConfig : ParseConfig Options = [
   ([("--help", "", "")],
     "Display this list of options",
     lam p: ArgPart Options.
-      let o: Options = p.options in {o with printHelp = true})
+      let o: Options = p.options in {o with printHelp = true}),
+  -- TODO: I personally prefer the arg --wasm but to be consistent
```
elegios: Consistency is probably the right thing for now. Our flags and options have grown somewhat haphazardly, so there might be a point to take a pass and look over them, but that should probably not be in this pr.

stdlib/fileutils.mc
```diff
@@ -0,0 +1,74 @@
...
+
+let parseMCoreLibs : String -> Map String String = lam str.
+  match str with "" then mapEmpty cmpString else 
+  let libs = strSplit ":" str in 
+  let pairs = map (lam lib. strSplit "=" lib) libs in 
+
+  let isValid = lam pair. 
+    if neqi (length pair) 2 then 
+      error (join [
+        "Invalid MCore_LIBS environment: '",
```
elegios: Casing

stdlib/fileutils.mc
```diff
@@ -0,0 +1,74 @@
...
+  using mapEq eqString
+utest parseMCoreLibs "stdlib=a/b/c:foo=bar" 
+  with mapFromSeq cmpString [("stdlib", "a/b/c"), ("foo", "bar")] 
+  using mapEq eqString
+
+let parseMCoreLibsEnv : () -> Map String String = lam.
+  match sysGetEnv "MCORE_LIBS" with Some s then
+    parseMCoreLibs s
+  else
+    error "Environment variable $MCORE_LIBS is not set!"
```
elegios, line 41: I believe that we don't actually require it to be set presently, that it essentially defaults to something like `stdlib=some/default/path/per/system`. Do you handle that elsewhere?

stdlib/mexpr/lamlift.mc
```diff
@@ -1040,5 +1040,19 @@ let nestedUtest = preprocess (bindall_ [
     addi_ (var_ "x") (int_ 1))))
 ]) in
 utest liftLambdas nestedUtest with nestedUtest using eqExpr in
-
-()
+();
+(let innerFunction = preprocess (bindall_ [
```
elegios: I'm assuming this is some form of temporary testing?

stdlib/mexpr/symbolize.mc
```diff
@@ -23,31 +23,103 @@ include "repr-ast.mc"
...
 
-type SymEnv = {
-  varEnv: Map String Name,
-  conEnv: Map String Name,
-  tyVarEnv: Map String Name,
-  tyConEnv: Map String Name,
-  allowFree: Bool,
-  ignoreExternals: Bool,
-  reprEnv: Map String Name
+-- type SymEnv = {
```
elegios: probably remove if it's not what we're actually doing

stdlib/mexpr/symbolize.mc
```diff
@@ -23,31 +23,103 @@ include "repr-ast.mc"
...
+--   tyVarEnv: Map String Name,
+--   tyConEnv: Map String Name,
+--   allowFree: Bool,
+--   ignoreExternals: Bool,
+--   reprEnv: Map String Name,
+--   langEnv: Map String SymEnv
+-- }
+
+type NameEnv = {
+  varEnv : Map String Name,   -- sem f
```
elegios: The comments seem mlang specific, but that's not the only use, right?

stdlib/mexpr/symbolize.mc
```diff
@@ -23,31 +23,103 @@ include "repr-ast.mc"
...
+type LangEnv = {
+  ident : Name,
+
+  -- Map from the string of a syn identifier to a tuple 
+  -- in which the first element is the symbolized name of the syn identifier
+  -- and the second element is a list of symbolized constructor names.
+  -- The third element is the parameter list of the syn. 
+  syns: Map String (Name, [Name], [Name]),
+  -- Map from the string of a sem identifier to a tuple
+  -- in which teh first element is the list of parameters
```
elegios: teh

stdlib/mexpr/symbolize.mc
```diff
@@ -23,31 +23,103 @@ include "repr-ast.mc"
...
+}
+
+type LangEnv = {
+  ident : Name,
+
+  -- Map from the string of a syn identifier to a tuple 
+  -- in which the first element is the symbolized name of the syn identifier
+  -- and the second element is a list of symbolized constructor names.
+  -- The third element is the parameter list of the syn. 
+  syns: Map String (Name, [Name], [Name]),
```
elegios: Storing things in this way during `symbolize` feels a bit odd, I'll update this more when I've seen its use.

stdlib/mexpr/symbolize.mc
```diff
@@ -23,31 +23,103 @@ include "repr-ast.mc"
...
+}
+
+type SymEnv = {
+  allowFree : Bool, 
+  ignoreExternals : Bool,
+  currentEnv : NameEnv,
+  langEnv : Map String LangEnv
+}
+
+let updateVarEnv = lam env : SymEnv . lam varEnv : Map String Name.
```
elegios: These names might not play well with the current `include` handling, which I think we still have to consider. For example, type-checking will have similarly named environments, that might have similarly named functions.

stdlib/mexpr/symbolize.mc
```diff
@@ -77,12 +149,25 @@ lang SymLookup
...
-  sem setSymbol env =| ident ->
+  sem setSymbol env =| ident -> 
     if nameHasSym ident then (env, ident)
     else
       let ident = nameSetNewSym ident in
       (mapInsert (nameGetStr ident) ident env, ident)
 
+  -- Take an identifier that already has a symbol and add it to the env
+  -- without generating a new symbol
+  sem updateSymbol : Map String Name -> Name -> (Map String Name, Name)
```
elegios: Naming feels a bit weird here, especially in conjunction with `setSymbol` (which admittedly isn't the most descriptive name, but you know).

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,718 @@
...
+include "digraph.mc"
+
+type CompositionCheckEnv = {
+  nextId : Int,
+  -- Mapping from the symbolized name of a syn or sem to their base declaration
+  baseMap : Map Name Name,
+  -- Mapping from symbolized name of a syn to the amount of parameters
+  paramMap : Map Name Int,
+  -- Mapping form a symbolized sem name to the symbolized names of its arguments if they are defined.
+  semArgsMap : Map Name (Option [Name]),
```
elegios: Are the parameters required to have the same `Name` across fragments? I would like to lift that restriction, since it should be relatively simple now compared to doing it in `boot`.

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,718 @@
...
+
+  -- Mapping from symbolized names of a sem to its cases that are ordered by
+  -- the subset relation on the patterns. 
+  -- We also introduce a unique id for each case. We need this id to be able 
+  -- to remove duplicate cases under languaage composition.
+  semPatMap : Map Name [use MLangAst in {pat: Pat, thn : Expr, id : Int}]
```
elegios: I think it'd be better to save the digraph here, rather than an already ordered list. In `mlang.ml` each case has a pre-computed `pos_pat` and `neg_pat` for performance, and then each `sem` has the digraph.

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,718 @@
...
+  let incl2pats = lam i : Name. 
+    match mapLookup i env.semPatMap with Some pats then
+      pats
+    else error (concat 
+      "Illegal state during composition-check! The name patterns for each "
+      "included sem should have already been included!")
+  in 
+  join (map incl2pats includes)
+
+let carthesianProduct : all a. all b. [a] -> [b] -> [(a, b)] = lam xs. lam ys.
```
elegios: This is `liftA2 (lam a. lam b. (a, b))`, and I've sometimes also declared a `pairWithLater` function that does what your `indexPairs` does, don't quite remember if it's in `seq.mc` or not.

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,718 @@
...
+    synIdent : Name,
+    info : Info
+  }
+  | InvalidSemPatterns {
+    semIdent : Name,
+    info : Info
+  }
+
+  syn CompositionWarning = 
+  | MismatchedSemArgNames {
```
elegios: Probably remove

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,718 @@
...
+      if eqi 1 (setSize includeSet) then
+        _ok (insertBaseMap env s.ident (head includeList))
+      else
+        _err (DifferentBaseSem {
+          semIdent = s.ident, 
+          info = s.info
+        })
+
+  sem parseCases env = 
+  | DeclSem s ->
```
elegios, line 281: I feel like this is longer than it needs to be and also recomputes things unnecessarily. I'd recommend looking at https://github.com/miking-lang/miking/blob/dcdbdab33a52312d9e0195dcad994b81c1e59854/src/boot/lib/mlang.ml#L248-L265 and https://github.com/miking-lang/miking/blob/dcdbdab33a52312d9e0195dcad994b81c1e59854/src/boot/lib/mlang.ml#L190-L221 which does the corresponding work.

The annoying thing is that `mlang.ml` has the advantage of having all of this information in an environment, which can be shaped differently from the AST, but here we would like to put it on the AST, which would *ideally* be done with product extension, which we of course don't have presently. I still think it should be doable with minor changes, e.g., we still store cases as a list under each `sem`, then construct the set/map when needed.

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,718 @@
...
+end
+
+lang TestLang = MLangSym + MLangCompositionCheck end
+
+mexpr 
+use TestLang in 
+use MLangPrettyPrint in 
+
+let handleResult = lam res.
+  switch _consume res
```
elegios: Use of private function, `result.consume` instead. We mentioned this earlier, so I'll only put the comment here as a reminder, and ignore it for other places.

stdlib/mlang/include-handler.mc
```diff
@@ -0,0 +1,86 @@
...
+  sem flattenIncludes : Ref (Set String) -> String -> Map String String -> Decl -> [Decl]
+  sem flattenIncludes included dir libs =
+  | DeclInclude {path = path, info = info} ->
+    let path = findPath dir libs info path in 
+    let prog = handleIncludesFile included (eraseFile path) libs path in 
+    prog.decls
+  | other -> [other]
+
+  sem findPath : String -> Map String String -> Info -> String -> String
+  sem findPath dir libs info =| path ->
```
elegios, line 70: As far as I can tell, these are not the current semantics for include:

- `foo/bar.mc` is either `./foo/bar.mc` or `stdlib:foo/bar.mc`
- `lib:foo/bar.mc` is `foo/bar.mc` in library `lib` as described using `MCORE_LIBS=lib=stuff`
- `./foo/bar.mc` is either treated like `foo/bar.mc`, or required to be relative, I don't quite remember.

I think this code will look for it in any library regardless of prefix, right?

stdlib/mlang/include-handler.mc
```diff
@@ -0,0 +1,86 @@
...
+    let existingFiles = filter sysFileExists paths in 
+    let existingFilesAsSet = setOfSeq cmpString existingFiles in 
+
+    switch (setSize existingFilesAsSet)
+      case 0 then 
+        errorSingle [info] "File not found!"
+      case 1 then 
+        head (setToSeq existingFilesAsSet)
+      case _ then 
+        -- TODO(voorberg, 09/05/2024): This happens because we dont properly
```
elegios, line 84: This should be a matter of porting the canonicalization code from boot, https://github.com/miking-lang/miking/blob/dcdbdab33a52312d9e0195dcad994b81c1e59854/src/boot/lib/utils.ml#L126-L167, most of which is a general helper thing.

src/main/compile.mc
```diff
@@ -137,43 +138,49 @@ let compileWithUtests = lam options : Options. lam sourcePath. lam ast.
...
-    (if options.debugParse then printLn (expr2str ast) else ());
-
-    compileWithUtests options file ast; ()
-  in
-  if options.accelerate then compileAccelerate files options args
-  else iter compileFile files
+
+  if options.mlangPipeline then
+    printLn "WARNING: You are using an experimental, unstable pipeline.";
+    use MainLang in
```
elegios: Probably a bit too general of a name for now

stdlib/mlang/ast.mc
```diff
@@ -55,28 +73,63 @@ lang SynDeclAst = DeclAst
...
+             -- The first string identifies the langauge of the include
+             -- and the second string identifies the name.
+             includes : [(String, String)],
              info : Info}
+
+  sem infoDecl = 
+  | DeclSyn d -> d.info
 end
 
+-- DeclSynProdExt
```
elegios: Should this be inhere before we actually start work on that? I think at least it would be nice to separate such code PR-wise

stdlib/mlang/ast.mc
```diff
@@ -55,28 +73,63 @@ lang SynDeclAst = DeclAst
...
+                    extIdent : Name,
+                    globalExt : Type,
+                    specificExt : [{ident : Name, tyIdent : Type}],
+                    info : Info}
+
+  sem infoDecl = 
+  | DeclSynProdExt d -> d.info
+end
 -- DeclSem --
 lang SemDeclAst = DeclAst
```
elegios: A thing that might be nice to do here (and in the rest of the compiler) is to order/separate fields that we would prefer were product extensions from those we would consider core. In this case I'm guessing core fields are `ident`, `tyAnnot`, `args`, and `cases`, while `includes` and `info` would be extensions?

stdlib/mlang/ast.mc
```diff
@@ -55,28 +73,63 @@ lang SynDeclAst = DeclAst
...
   | DeclSem {ident : Name,
              tyAnnot : Type,
              tyBody : Type,
-             args : [{ident : Name, tyAnnot : Type}],
+             args : Option [{ident : Name, tyAnnot : Type}],
              cases : [{pat : Pat, thn : Expr}],
+             -- The list of semantic function s whose cases should be included.
+             -- The first string identifies the langauge of the include
+             -- and the second string identifies the name.
+             includes : [(String, String)],
```
elegios, line 110: Should these truly be strings? They seem like names to me, even if I'm guessing you're initially populating them before symbols have been assigned.

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,744 @@
...
+let insertArgsMap : CompositionCheckEnv -> (String, String) -> Option [Name] -> CompositionCheckEnv =  
+  lam env. lam k. lam v.
+    {env with semArgsMap = mapInsert k v env.semArgsMap}
+
+let insertSemPatMap = lam env. lam k. lam v.
+  {env with semPatMap = mapInsert k v env.semPatMap}
+
+lang MLangCompositionCheck = MLangAst + MExprPatAnalysis + MExprAst + MExprPrettyPrint
+  syn CompositionError =
+  | DifferentBaseSyn {
```
elegios, line 104: We mentioned including more info in these constructors, probably at least the fragments that conflict in their assumed base (and their `Info`s).

stdlib/mlang/composition-check.mc
```diff
@@ -0,0 +1,744 @@
...
+      nameGetStr e.semIdent,
+      "' includes or defined patterns which are overlapping or equal!"
+    ] in 
+    errorMulti [(e.info, "")] msg
+
+  sem checkComposition : MLangProgram -> Result CompositionWarning CompositionError CompositionCheckEnv
+  sem checkComposition =| prog -> 
+    _foldl parseAll _emptyCompositionCheckEnv prog.decls 
+
+  sem parseAll : CompositionCheckEnv ->
```
elegios: I'm guessing the naming here is from the "parse don't validate" idea, but `parseAll` is fairly non-descript and could easily collide with something else.

stdlib/mlang/compile.mc
```diff
@@ -0,0 +1,492 @@
...
+
+  compositionCheckEnv : CompositionCheckEnv,
+
+  -- A map from identifier strings of semantic functions to the 
+  -- symbolized names that the function has in different fragments.
+  semSymbols : Map String [Name]
+}
+
+-- Substitute the identifier stored in a TmVar based on the provided substitution
+recursive let subTmVarSymbol = lam subst : (Name -> Name). lam expr.
```
elegios: This feels somewhat suspect, either because it should exist elsewhere already, or that it only touches `TmVar`s and that it's not extensible.

stdlib/mlang/compile.mc
```diff
@@ -0,0 +1,492 @@
...
+  {ctx with semSymbols = mapInsert s newValue ctx.semSymbols}
+
+-- Create a substitution function by partially applying the first two elements
+-- This substitution function maps symbols belonging to semantic function in 
+-- included language fragments to the symbol of the semantic function in the 
+-- current fragment
+let createSubst = lam semSymbols. lam semNames. lam n. 
+  let s = nameGetStr n in 
+  match mapLookup s semSymbols with Some xs then
+    if optionIsSome (find (lam x. nameEqSym x n) xs) then
```
elegios: This looks like `xs` should be a set

stdlib/mlang/compile.mc
```diff
@@ -0,0 +1,492 @@
...
+      match mapLookup s semNames with Some res then res
+      else n
+    else
+      n
+  else 
+    n
+
+let createSubst2 = lam semSymbols. lam semNames. lam n.
```
elegios: Naming, also somewhat suspect that it would use `nameEqStr` presumably after symbolization

stdlib/mlang/compile.mc
```diff
@@ -0,0 +1,492 @@
...
+  -- TODO(voorberg, 2024-04-23): Add test case for the compilation of externals.
+  | DeclExt d -> _ok (
+    withExpr ctx (TmExt {ident = d.ident,
+                         tyIdent = d.tyIdent,
+                         effect = d.effect,
+                         info = d.info,
+                         ty = tyunknown_,
+                         inexpr = uunit_}))
+  -- TODO(voorberg, 2024-04-23): Add test case for error on DeclInclude
+  | DeclInclude d -> _err (FoundIncludeError {info = d.info, path = d.path})
```
elegios: Similar to before, might be better to just not have this case.

stdlib/mlang/compile.mc
```diff
@@ -0,0 +1,492 @@
...
+  | DeclLang l -> 
+    let langStr = nameGetStr l.ident in
+
+    let typeDecls = filter isTypeDecl l.decls in
```
elegios, line 174: You could do a `mapOption (lam x. match x with SemDecl x then Some x else None ())`, whereby later uses of these lists won't have to check that things are `SemDecl` etc.

I more and more tend to define the record carried separately so that it's easy to annotate such lists, e.g.:

```
lang Foo
  type FooRec = {...}
  syn Thing =
  | Foo FooRec
end
```

stdlib/result.mc
```diff
@@ -348,6 +348,40 @@ utest
   ()
 with ()
 
+let _foldl : all w. all e. all a. all b.
```
elegios: `_foldlM` to be consistent with the other functions, and include in the interface at the bottom.

Do consider if you can structure things in a different way if possible though, since this will very effectively suppress the ability to collect multiple errors, since we can't run a computation for a `b` if we already found a failure.

stdlib/result.mc
```diff
@@ -348,6 +348,40 @@ utest
...
+            case ResultOk c then
+              workOK (mapUnion accWarn c.warnings) (c.value) tl
+            case ResultErr c then
+              ResultErr {c with warnings = mapUnion accWarn c.warnings }
+            end
+          else
+            ResultOk { warnings = accWarn, value = acc }
+    in workOK (_emptyMap ()) acc
+
+let _foldlfun : all w. all e. all a. all b.
```
elegios: This does not feel general enough to have in the library here I think. I think it would be implemented with

```
lam a. lam b. lam fs. result.foldlM (lam a. lam f. f a b) a fs
```

looking purely at the type signature

stdlib/seq.mc
```diff
@@ -472,16 +472,59 @@ utest filterOption [Some 3, Some 2, None (), Some 4] with [3, 2, 4] using eqSeq
...
 
+-- Find the first element in a sequence satisfying a predicate in a list.
 recursive
   let find : all a. (a -> Bool) -> [a] -> Option a = lam p. lam seq.
     if null seq then None ()
     else if p (head seq) then Some (head seq)
     else find p (tail seq)
 end
 
+let eqRec = lam lhs. lam rhs.and (eqi lhs.x rhs.x) (eqi lhs.y rhs.y)
```
elegios: This adds to the global namespace, which might not be ideal. What does this test case capture that the others do not?

test/mexpr/float-test.mc
```diff
@@ -53,8 +53,7 @@ utest 0.25 with 7.5 using neg geqf in
 utest 0.25 with 0.25 using geqf in
 utest 2.2 with 2.2 using eqf in          -- equal (float)
 utest 2.2 with 2.3 using neg eqf in
-utest subf 3.0 5.0 with negf 2.0
-      using eqf in
+utest subf 3.0 5.0 with negf 2.0 using eqf in
```
elegios: I'm assuming changes in this file are related to wasm changes? How come there are changes for the constants? (I'm assuming the commented out `stringIsFloat` are due to lack of intrinsics or some such)

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+
+let fst : all a. all b. (a, b) -> a = lam p.
+    match p with (res, _) in res
+
+let snd : all a. all b. (a, b) -> b = lam p.
+    match p with (_, res) in res
+
+
+let mapDisjoint : all a. all b. (Map String a) -> (Map String b) -> Bool = lam l. lam r.
+    let leftKeys = setOfSeq cmpString (mapKeys l) in
```
elegios: `setOfKeys` exists

elegios: Also, this function is probably more efficiently implemented with `mapIntersectWith`, to not create more intermediate values.

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+        -- This can be rewritten to use a shallow map on declarations. E.g.
+        -- smap (symbolizeExpr env) (DeclUtest t) 
+        let decl = DeclUtest {t with test = symbolizeExpr env t.test,
+                                     expected = symbolizeExpr env t.expected,
+                                     tusing = optionMap (symbolizeExpr env) t.tusing} in 
+        (env, decl)
+    | DeclExt t -> 
+        -- When ignoreExternals is set, the DeclExt should have been filtered
+        -- out of the program before attempting to symbolize the declarations.
+        if env.ignoreExternals then
```
elegios: Is this consistent with handling for externals in regular `mexpr`?

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+        -- 1. Symbolize ident and params of SynDecls in this langauge
+        let symbSynStep1 = lam langEnv : NameEnv. lam synDecl.
+            match synDecl with DeclSyn s in
+            let env = updateEnv env langEnv in 
+
+            let ident = nameSym (nameGetStr s.ident) in 
+            match mapAccumL setSymbol env.currentEnv.tyVarEnv s.params with (_, params) in
+
+            let synn = DeclSyn {s with params = params,
+                                       ident = ident} in
```
elegios: This should probably not be a new unique `Name`, rather the `Name` of the type to be declared, which is either the one here or one that's included.

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+
+        let isTypeDecl = lam d. match d with DeclType _ then true else false in 
+        let typeDecls = filter isTypeDecl t.decls in 
+
+        -- 1. Symbolize ident and params of SynDecls in this langauge
+        let symbSynStep1 = lam langEnv : NameEnv. lam synDecl.
+            match synDecl with DeclSyn s in
+            let env = updateEnv env langEnv in 
+
+            let ident = nameSym (nameGetStr s.ident) in
```
elegios: There is a helper for this combination, something like `nameNewSym` or `nameSetSym`.

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+        let isSemDecl = lam d. match d with DeclSem _ then true else false in 
+        let semDecls = filter isSemDecl t.decls in 
+
+        let isTypeDecl = lam d. match d with DeclType _ then true else false in 
+        let typeDecls = filter isTypeDecl t.decls in 
+
+        -- 1. Symbolize ident and params of SynDecls in this langauge
+        let symbSynStep1 = lam langEnv : NameEnv. lam synDecl.
+            match synDecl with DeclSyn s in
+            let env = updateEnv env langEnv in
```
elegios: It seems like `updateEnv` will be called a lot, is that truly required?

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+
+        -- 1. Symbolize ident and params of SynDecls in this langauge
+        let symbSynStep1 = lam langEnv : NameEnv. lam synDecl.
+            match synDecl with DeclSyn s in
+            let env = updateEnv env langEnv in 
+
+            let ident = nameSym (nameGetStr s.ident) in 
+            match mapAccumL setSymbol env.currentEnv.tyVarEnv s.params with (_, params) in
+
+            let synn = DeclSyn {s with params = params,
```
elegios: I feel like the first pass should only assign symbols to `ident`, because those will be in scope outside the `syn` itself, and then everything internal is symbolized after in a phase 2, much like how it's done for rec lets.

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+        -- 2. Symbolize DeclType and params
+        let symbDeclType = lam langEnv : NameEnv. lam typeDecl. 
+            match typeDecl with DeclType t in 
+
+            -- Symbolize ident
+            let ident = nameSym (nameGetStr t.ident) in 
+
+            -- -- Check for name conflicts with syns and other types.
+            -- -- Throw an error if DeclType is included with the same identifier
+            -- errorOnNameConflict includedTypes ident langIdent t.info ;
```
elegios: Why are these commented out?

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+                mapInsert (nameGetStr ident) ident langEnv.tyConEnv
+            else 
+                langEnv.tyConEnv
+            in
+
+            ({langEnv with tyConEnv = tyConEnv}, synn)
+        in
+        match mapAccumL symbSynStep1 langEnv synDecls with (langEnv, synDecls) in 
+
+        -- 2. Symbolize DeclType and params
```
elegios: Comment implies that the body (rhs) of `DeclType` isn't symbolized.

stdlib/mlang/symbolize.mc
```diff
@@ -0,0 +1,787 @@
...
+            in
+            let cases = map symbCases s.cases in
+
+            let decl = DeclSem {s with cases = cases, 
+                                       tyAnnot = tyAnnot,
+                                       args = args} in 
+
+            (langEnv, decl)
+        in
+        match mapAccumL symbSem2 langEnv semDecls with (langEnv, semDecls) in
```
elegios: No need for a `mapAccumL` here, using a `map` makes it clearer that no names escape from the sem bodies

stdlib/mlang/main.mc
```diff
@@ -0,0 +1,108 @@
...
+                MLangSym + MLangCompositionCheck +
+                MExprPrettyPrint + MExprEval + MExprEq + 
+                MLangConstTransformer + MLangIncludeHandler +
+                PhaseStats + LanguageComposer
+
+  sem myEval : Expr -> Expr
+  sem myEval =| e ->
+    eval (evalCtxEmpty ()) e 
+
+  -- TODO: add node count for MLang programs to phase-stats
```
elegios: I ran into the same issue in some other places as well, which suggest that phase-stats maybe shouldn't assume things to be `Expr`, rather take a function or something like that.

Also, why is this code commented out but present?

stdlib/mlang/language-composer.mc
```diff
@@ -0,0 +1,355 @@
...
+
+include "bool.mc"
+include "common.mc"
+include "error.mc"
+include "tuple.mc"
+include "name.mc"
+include "seq.mc"
+include "map.mc"
+
+type DeclInfo
```
elegios, line 60: How do these relate to the `Decl` type? Are they the same as the corresponding constructors, or do they carry different data?

stdlib/mlang/language-composer.mc
```diff
@@ -0,0 +1,355 @@
...
+    if not (null conflicts) then
+      errorMulti errors "The declared type has an identifier that conflicts with included types or syns!"
+    else
+      let info = decl2info langStr decl in 
+      (ctxWithDeclInfo ctx (langStr, nameGetStr d.ident) info, decl)
+  | _ -> error "Only Type, Syn, and Sem declarations can be contained inside of a langauge!"
+
+  sem addImplicitIncludes langStr includes definedSynsSems =
+  | ctx ->
+    -- We are going to remove elements from the ctx.langMap that
```
elegios: I think there's something odd about the negations in this comment relative to the code, and generally double negations are hard.

stdlib/mlang/language-composer.mc
```diff
@@ -0,0 +1,355 @@
...
+    -- (1) do not belong to an included langauge
+    -- (2) that have already been included explicitly through a syn or sem
+    -- (3) that are not Type declarations.
+    let pred = lam k. lam v. 
+      match k with (origLang, ident) in 
+        (and (not (isTypeInfo v))
+             (and (seqMem eqString includes origLang)
```
elegios: `seqMem` smells like a performance bug to me

elegios (commented):
We have 2 space indentation as a convention, I don't think that's the case for some of the new files in this PR.

src/main/compile.mc
marten-voorberg: I agree, but I also could not think of any better name.

stdlib/mexpr/lamlift.mc
marten-voorberg: Yeah, I think these were some test cases from when I was playing around with lambda lifting months ago that somehow made it into this PR.

stdlib/mlang/ast.mc
marten-voorberg: I agree. I've removed it now.

stdlib/mlang/ast.mc
marten-voorberg: Yeah, I currently have them as String pairs, but I think that names would be slightly better as it would allow you to use the same language name in multiple (unrelated) places once we have better include handling.

I might make this change in a future PR though.

stdlib/mlang/language-composer.mc
marten-voorberg: They carry a subset of the data. I thought it would be rather expensive to carry all of the cases and constructors around so I created this `Info` type.

stdlib/mlang/language-composer.mc
marten-voorberg: How so? A set would typically be more efficient, but it is quite common to have to check for membership in a list anyways right? It doesn't really make sense to me to construct a set each time.

stdlib/mlang/main.mc
marten-voorberg: Yeah, it would be good if phase-stats was a bit more general.

marten-voorberg: The code for MLang became outdated. I'll re-add it at some point, but for now I've removed the comment.

stdlib/result.mc
marten-voorberg: I don't think it makes sense to gather multiple errors in most contexts. If the new value of the accumulator can not be computed for some element in the list, then I would argue the computation is undefined for all further computations since there is no accumulator. 

E.g., if you are traversing the spine of a program like this:
```
let f = ...
let x = ...
let y = some error here
let z = f x y
```
If we have an error at the computation of `y`, then the accumulator will maybe not know enough about `y` and we will get another error at `z` which can then also propegate upwards. So you get a ton of irrelevant errors instead of the single **actual** error.

stdlib/result.mc
marten-voorberg: Agreed, moved it to `composition-check.env` with your definition.

test/mexpr/float-test.mc
marten-voorberg: Yeah, this is old stuff. I removed it.

stdlib/mlang/composition-check.mc
marten-voorberg: Agreed, but I think this should come in a future PR.

stdlib/mexpr/symbolize.mc
marten-voorberg: What would you suggest? Using names that make it clearer that these belong to symbolization?

stdlib/mexpr/symbolize.mc
marten-voorberg: This function isn't actually used anymore in my current implementation so I've removed it.

stdlib/result.mc
elegios: I meant more to consider if it is possible to formulate things such that they're more independent of each other, so you can report more errors. For example, in `tool.mc` there's an environment being passed around with type `Map Name (Either (Res TypeInfo) (Res TokenInfo))` (where `Res` is an alias around `Result`), so that even if a definition is faulty, I can still keep track of whether it's a type or a token and error if it's used incorrectly. Key point being that the type isn't `Res (Map Name (Either TypeInfo TokenInfo))`, since that's less precise and allows fewer errors. It's rarely *truly* the case that a particular value depends on the entirety of an accumulator, it's just sometimes inconvenient to insert `Result` at useful places.

stdlib/mexpr/symbolize.mc
elegios: probably yes, the most common way we do it is to name things `<name-of-file><name-of-value>`, so `symbolizeUpdateVarEnv`.

src/main/compile.mc
elegios: `MlangPipeline`? It's clear, and easy to change later if we come up with something better

stdlib/mlang/language-composer.mc
elegios: Probably document the connection.

(Depending on precise needs it might not be an issue to carry the original `Decl` around, perhaps with a subsetted type, since it's already a value that exists, while this will have to create a new value with a subset of the data. Probably reasonable to have a separate type though if the extra data isn't needed, and especially if it would get in the way somehow.)

stdlib/mlang/language-composer.mc
elegios: `pred` will be called multiple times (so at least complexity "size of ctx.langMap" * "size of includes", vs. "size of ctx.langMap" * "log of size of includes" + "n log n of size of includes").

If you make `includes` a set/map you could also work with `mapIntersect`.

stdlib/mlang/compile.mc
```diff
@@ -122,8 +139,11 @@ let isSynDecl = use MLangAst in
...
   lam d. match d with DeclSem _ then true else false
 
-lang MLangCompiler = MLangAst + MExprAst
+lang DeclCompiler = DeclAst + Ast
   sem compileDecl : CompilationContext -> Decl -> CompilationResult
+end
+
+lang LetDeclCompiler = DeclCompiler + LetDeclAst + LetAst
   sem compileDecl ctx = 
   | DeclLet d -> _ok (
```
elegios: These should not be using the `_` versions of these names, as they're intended to be private. There are a number of other similar uses as well, not only of the `_ok` function.

stdlib/mlang/composition-check.mc
```diff
@@ -60,15 +69,9 @@ let collectPats = lam env. lam includes.
...
-let carthesianProduct : all a. all b. [a] -> [b] -> [(a, b)] = lam xs. lam ys.
-  let work1 : a -> [b] -> [(a, b)] = lam x. lam ys. 
-    map (lam y. (x, y)) ys
-  in 
-  join (map (lam x. work1 x ys) xs)
-
 let indexPairs : Int -> [(Int, Int)] = lam n. 
   let indices = range 0 n 1 in 
-  let pairs = carthesianProduct indices indices in
+  let pairs = seqLiftA2 (lam a. lam b. (a, b)) indices indices in
```
elegios: This does at least twice the work it needs because of the filter below. I would suggest you port https://github.com/miking-lang/miking/blob/dcdbdab33a52312d9e0195dcad994b81c1e59854/src/boot/lib/mlang.ml#L26-L30 and put it in `seq.mc` (your choice if it should create the pair or take a function to call for each pair).

src/main/compile.mc
marten-voorberg: isnt that the current name? Or am I missing something?

stdlib/mlang/postprocess.mc
```diff
@@ -0,0 +1,61 @@
+-- This transformation converts the `Name`s of semantic functions to `Name`s 
+-- with the same symbol, but also have a unique string in the shape of 
+-- <LangName>_<SemName>. 
+-- 
+-- This transformation is added because some parts of the exising MExpr
+-- transformations to OCaml do not respect the symbols fully and also require
+-- the strings of names to be unique.
```
elegios, line 7: This sounds rather worrying, do you have pointers to what doesn't handle things correctly?

src/main/compile.mc
elegios: A suggestion for what to change `MainLang` to.

stdlib/mlang/ast.mc
elegios: I know from experience that using `String` temporarily has a tendency to stick around and be very annoying to fix properly, so I would prefer to do that already now. If I understand what's needed it shouldn't be that big of a change, right?

stdlib/mlang/composition-check.mc
marten-voorberg: done.

stdlib/mlang/postprocess.mc
marten-voorberg: I discussed this a little bit with @larshum and it seems that the utest-generation sometimes uses `nameGetStr` for the names of generated code which can cause incorrect bheaviour when you have names with the same strings but different symbols.

larshum: Correct. I think those cases should be fairly easy to fix, by adding the symbol to the string we generate. There's also code in Miking DPPL that assumes the MLang name mangling works in a particular way, which would surely break as well.

stdlib/mlang/compile.mc
marten-voorberg: Use `mexpr/utils.mc`

stdlib/mlang/symbolize.mc
marten-voorberg: Ill fix this

stdlib/mlang/symbolize.mc
marten-voorberg: Add a comment!

stdlib/mlang/symbolize.mc
marten-voorberg: Move symbolilzation of params to step 3.

stdlib/mlang/postprocess.mc
marten-voorberg: I have created an issue for this (see #850). The MLang pipeline will just use the `postprocess` work around until this issue is resolved.

PR #846, opened 2024-05-27
Pure arrays

david-broman:
This is a small library with arrays without side effects. Random access operations have O(n log n) complexity, with a rather small constant factor.


marten-voorberg:
The PR description says O(n log n) but should be O(log n) for access operations.

stdlib/parray.mc
```diff
@@ -0,0 +1,369 @@
...
+-- of O(log n).
+
+-- Pure Array
+type PA a
+con PAData  : all a. (Int,a,a,a,a,a,a,a,a,a,a) -> PA a
+con PANode  : all a. (Int,Int,PA a,PA a,PA a,PA a,PA a,PA a,PA a,PA a,PA a,PA a) -> PA a
+con PAEmpty : all a. () -> PA a
+con PANext  : all a. (PA a) -> PA a
+
+let empty = PAEmpty()
```
marten-voorberg: This should be renamed to avoid name collisions

stdlib/parray.mc
```diff
@@ -0,0 +1,369 @@
...
+  match pa with PAData(7,x0,x1,x2,x3,x4,x5,x6,x7,x8,x9)
+           then PAData(8,x0,x1,x2,x3,x4,x5,x6,y ,x8,x9) else
+  match pa with PAData(8,x0,x1,x2,x3,x4,x5,x6,x7,x8,x9)
+           then PAData(9,x0,x1,x2,x3,x4,x5,x6,x7,y ,x9) else
+  match pa with PAData(9,x0,x1,x2,x3,x4,x5,x6,x7,x8,x9) then
+           if eqi l 0 then PANode(1,10,PAData(
+                   10,x0,x1,x2,x3,x4,x5,x6,x7,x8,y),
+                   empty,empty,empty,empty,empty,empty,empty,empty,empty)
+           else
+            PANext(PAData(10,x0,x1,x2,x3,x4,x5,x6,x7,x8,y)) else
```
aathn: I think these could be written less verbosely, as
```
match pa with PAData t in
switch t.0
case 1 then PAData{t with #label"0" = 2, #label"2" = y}
case 2 then PAData{t with #label"0" = 3, #label"3" = y}
...
end
```
Other parts in the file could also benefit from the `switch` sugar

stdlib/parray.mc
```diff
@@ -0,0 +1,369 @@
...
+-- Either keep track of the number of elements in your pure array in another
+-- variable, or use `lengthPA` to get the current number of elements.
+let addPA = lam pa. lam y.
+  workAddPA pa y 0
+
+
+
+/-
+recursive
+let addPA : all a. PA a -> a -> PA a = lam pa. lam y.
```
aathn: Is this a duplicate definition of `addPA`?

stdlib/parray.mc
david-broman: Well, that is a matter of taste. I will think about it.

stdlib/parray.mc
david-broman: It is commented-out code that I forgot to remove. It is removed in the next commit.

stdlib/parray.mc
david-broman: Fixed in next version.

stdlib/parray.mc
david-broman: I have fixed this now when I think it makes it less verbose, and still clear. Thanks for the feedback!

PR #858, opened 2024-08-29
Add 'nameRemoveSym' to stdilb/name.mc

marten-voorberg:


johnwikman:
Loving the title of this PR :slightly_smiling_face: 

I would suggest adding these unit tests as well for completeness
```
let _t2 = nameNoSym "Foo" in

utest nameEq _t1 _t2 with false in
utest nameEq (nameRemoveSym _t1) _t2 with true in
```

PR #862, opened 2024-09-09
Extend bool stdlib with 'and' and 'or' on arrays

marten-voorberg:
Currently, we have `and` and `or` with arity 2. This leads to some rather ugly syntax when you have a list of conditions. E.g. `(and p1 (and p2 (and p3 p4)))`. With this pr, you'd be able to write `allb [p1, p2, p3, p4]`. The name comes from `all` (all must be true) and `b` for boolean.


marten-voorberg:
Closing because I used the wrong base branch...

PR #863, opened 2024-09-09
Add `allb` `someb` `tupleEq2` `setFilter` `strStartsWith` and `strEndsWith` to stdlib

marten-voorberg:
Currently, we have `and` and `or` with arity 2. This leads to some rather ugly syntax when you have a list of conditions. E.g. `(and p1 (and p2 (and p3 p4)))`. With this pr, you'd be able to write `allb [p1, p2, p3, p4]`. The name comes from `all` (all must be true) and `b` for boolean.

### DIFF
--- a/stdlib/string.mc
+++ b/stdlib/string.mc
@@ -236,14 +236,7 @@ utest strTrim "   bbbbb  bbb " with "bbbbb  bbb"
 utest strTrim "ccccc c\t   \n" with "ccccc c"
 
 -- `strEndsWith str suffix` returns true if `str` ends with the provided `suffix`
-let strEndsWith : String -> String -> Bool
-  = lam str. lam suffix. 
-    let strLen = length str in 
-    let suffixLen = length suffix in 
-    if lti strLen suffixLen then
-      false
-    else
-      eqString (subsequence str (subi strLen suffixLen) strLen) suffix
+let strEndsWith : String -> String -> Bool = lam s1. lam s2. isSuffix eqChar s2 s1
 
 utest strEndsWith "Something" "thing" with true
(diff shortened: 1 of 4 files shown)


stdlib/string.mc
```diff
@@ -235,6 +235,40 @@ utest strTrim " aaaa   " with "aaaa"
...
 
+-- `strEndsWith str suffix` returns true if `str` ends with the provided `suffix`
+let strEndsWith : String -> String -> Bool
+  = lam str. lam suffix. 
+    let strLen = length str in 
+    let suffixLen = length suffix in 
+    if lti strLen suffixLen then
+      false
+    else
+      eqString (subsequence str (subi strLen suffixLen) strLen) suffix
```
johnwikman: These functions are reimplementations of `isSuffix` and `isPrefix` from `seq.mc`. I think it makes sense to have string specific versions (and keep the tests!), but they should be defined as
```
let strEndsWith : String -> String -> Bool = isSuffix eqChar

let strStartsWith : String -> String -> Bool = isPrefix eqChar
```

marten-voorberg: Done! Although I did change it to `lam s1. lam s2. isSuffix eqChar s2 s1` because I feel that order of arguments makes more sense.

johnwikman: I would say that the regular order makes more sense due to that it plays nicer with partial application. As an example, let's say that I want to find all MCore files in a directory:
```
let mcore_files = filter (strEndsWith ".mc") (sysListDir ".") in
```

marten-voorberg: Good point. I've changed it.

PR #867, opened 2024-10-14
feat: start adding readBytes

didrikmunther:


src/boot/lib/mexpr.ml
```diff
@@ -1881,6 +1885,21 @@ and delta (apply : info -> tm -> tm -> tm) fi c v =
       TmSeq (fi, tms)
   | CreadLine, _ ->
       fail_constapp fi
+  | CreadBytes (Some TmRecord (_, r)), TmConst (_, CInt v) ->
```
didrikmunther, line 1888: I'm having problems here, I don't know which type I'm supposed to define in `src/boot/lib/ast.ml` in order to get a TmRecord in here.

PR #868, opened 2024-10-16
Add fileReadBytes external

didrikmunther:
Adds the ability to read X amount of characters from a `ReadChannel`. Commonly required in protocols such as JSON-RPC to read the required amount of characters as specified by a `Content-Length: X` header.

### DIFF
--- a/stdlib/ext/file-ext.mc
+++ b/stdlib/ext/file-ext.mc
@@ -59,6 +59,12 @@ external externalReadLine ! : ReadChannel -> (String, Bool)
 let fileReadLine : ReadChannel -> Option String =
   lam rc. match externalReadLine rc with (s, false) then Some s else None ()
 
+-- Reads a given number of bytes from the file.
+-- Returns None if end of file or error.
+external externalReadBytes ! : ReadChannel -> Int -> (String, Bool)
+let fileReadBytes : ReadChannel -> Int -> Option String =
+  lam rc. lam len. match externalReadBytes rc len with (s, false) then Some s else None ()
+
 -- Reads everything in a file and returns the content as a string.
 -- Should support Unicode in the future.
 external externalReadString ! : ReadChannel -> String
@@ -111,6 +117,18 @@ utest
     (l1,l2,l3,l4)
   else ("Error reading file","","","")
 with ("Hello", "Next string", "Final", "EOF") in
+
+-- Test reading x amount of characters from the file
+utest
+  match fileReadOpen filename with Some rc then
+    let l1 = match fileReadBytes rc 3 with Some s then s else "" in
+    let l2 = match fileReadBytes rc 4 with Some s then s else "" in
+    let l3 = match fileReadBytes rc 0 with Some s then s else "" in
+    let l4 = match fileReadBytes rc 1 with Some s then s else "" in
+    fileReadClose rc;
+    (l1,l2,l3,l4)
+  else ("Error reading file","","","")
+with ("Hel", "lo\nN", "", "e") in
 
 -- Check that the file size is correct
 utest fileSize filename with 23 in
--- a/stdlib/ext/file-ext.mc
+++ b/stdlib/ext/file-ext.mc
@@ -60,10 +60,15 @@ let fileReadLine : ReadChannel -> Option String =
   lam rc. match externalReadLine rc with (s, false) then Some s else None ()
 
 -- Reads a given number of bytes from the file.
--- Returns None if end of file or error.
-external externalReadBytes ! : ReadChannel -> Int -> (String, Bool)
-let fileReadBytes : ReadChannel -> Int -> Option String =
-  lam rc. lam len. match externalReadBytes rc len with (s, false) then Some s else None ()
+-- Returns (bytes, eof, error) where bytes is the read bytes
+external externalReadBytes ! : ReadChannel -> Int -> ([Int], Bool, Bool)
+let fileReadBytes : ReadChannel -> Int -> Option [Int] =
+  lam rc. lam len.
+    switch externalReadBytes rc len
+      case ([], true, _) then None ()
+      case (s, _, false) then Some s
+      case _ then None ()
+    end
 
 -- Reads everything in a file and returns the content as a string.
 -- Should support Unicode in the future.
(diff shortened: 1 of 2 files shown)


stdlib/ext/file-ext.mc
```diff
@@ -54,6 +54,12 @@ external readLine ! : ReadChannel -> (String, Bool)
 let readLine : ReadChannel -> Option String =
   lam rc. match readLine rc with (s, false) then Some s else None ()
 
+-- Reads a given number of bytes from the file.
```
marten-voorberg, line 62: What happens if in situations where I read more bytes than that are left? E.g., there are 5 bytes left and I attempt to read 10?

stdlib/ext/file-ext.mc
```diff
@@ -105,6 +111,18 @@ utest
   else ("Error reading file","","","")
 with ("Hello", "Next string", "Final", "EOF") in
 
+-- Test reading x amount of characters from the file
```
marten-voorberg, line 126: Consider adding a test case for the case in which `None ()` is returned

stdlib/ext/file-ext.mc
```diff
@@ -105,6 +111,18 @@ utest
...
+utest
+  match readOpen filename with Some rc then
+    let l1 = match readBytes rc 3 with Some s then s else "" in
```
marten-voorberg: It seems slightly weird to default to "". 

Maybe `match readBytes rc 3 with Some s in s` 

This might be preferable since not being able to read bytes seems to be an error which defaulting to "" handles silently

stdlib/ext/file-ext.mc
```diff
@@ -105,6 +111,18 @@ utest
...
+    let l2 = match readBytes rc 4 with Some s then s else "" in
+    let l3 = match readBytes rc 0 with Some s then s else "" in
+    let l4 = match readBytes rc 1 with Some s then s else "" in
+    readClose rc;
+    (l1,l2,l3,l4)
+  else ("Error reading file","","","")
```
marten-voorberg: Maybe just throw an `error` instead? E.g. `else error "..."`

marten-voorberg (changes requested):
Looks good! Glad you figured it out. 

I added a couple comments covering some very minor changes.

stdlib/ext/file-ext.mc
didrikmunther: That's a good suggestion, I've added it!

stdlib/ext/file-ext.mc
didrikmunther: It's written this way in the `readLine` function above, but I think this solution is much cleaner. Is there any downside to throwing an `error` in a `utest` though?

stdlib/ext/file-ext.mc
didrikmunther: Fixed!

stdlib/ext/file-ext.mc
didrikmunther: The solution to this turned out to be a completely different implementation in the OCaml code. I have added it now!

didrikmunther: The behavior is that it will return `Some a` where `len(a) == 5`. All calls after will return `None ()`

marten-voorberg: I think the comment could be a bit clearer on this. E.g.
```
Read the provided number of bytes from a `ReadChannel`.
If there is no more content, `None` is returned.
If the number of remaining bytes is smaller than `len`, 
all remaining bytes are returned

```

marten-voorberg: Also, what happens if some error occurs when reading? Is `None` returned or does this raise an error?

stdlib/ext/file-ext.mc
```diff
@@ -105,6 +114,21 @@ utest
   else ("Error reading file","","","")
 with ("Hello", "Next string", "Final", "EOF") in
 
+-- Test reading x amount of characters from the file
+utest
```
marten-voorberg, line 127: Consider nesting the test cases like this:
```
  match readOpen filename with Some rc then
  	utest readBytes rc 3 with Some "Hel" using optionEq eqString in
  	utest readBytes rc 4 with Some "lo\n" using optionEq eqString in 
  	utest readBytes rc 0 with Some "" using optionEq eqString in 
  	utest readBytes rc 1 with Some "e" using optionEq eqString in 
        -- If there are fewer bytes remaining than we requested, all remaining bytes
       -- are returned.
  	utest readBytes rc 1000 with Some "xt string \nFinal" using optionEq eqString in
  	utest readBytes rc 1000 with None () using optionEq eqString in 
  	utest readBytes rc 1000 with None () using optionEq eqString in 
  	()
  else 
  	error "File could not be read in tests for readBytes"
```

stdlib/ext/file-ext.mc
marten-voorberg: Well, I feel the utest should fail normally if the tested function behaves incorrectly (i.e. it does not crash but just returns the wrong value).

However, when there is a problem with the test case itself (e.g., the file can not be opened) this should raise an error IMO. This is because this is not a bug in the tested function, but in the test suite itself.

stdlib/ext/file-ext.mc
didrikmunther: That looks better, thanks!

stdlib/ext/file-ext.mc
didrikmunther: I agree!

stdlib/ext/file-ext.mc
didrikmunther: I've changed it so that it returns `None` if there was an error, or there was an EOF and empty string returned.

stdlib/ext/file-ext.mc
```diff
@@ -54,6 +56,36 @@ external readLine ! : ReadChannel -> (String, Bool)
...
+    -- tuple: (Content, reached EOF, had error)
+    case ("", true, _) then None () -- EOF
+    case (s, _, false) then Some s -- Success
+    case (_, _, _) then None () -- Error
+  end
+
+-- returns Option content if the requested number of bytes could be read
+-- otherwise, None is returned
+recursive
+  let readBytesBuffered : ReadChannel -> Int -> Option String =
```
didrikmunther, line 76: I think this could be removed from `stdlib`.

PR #871, opened 2024-10-22
Add .editorconfig

br4sco:
Rather than just specify code formating in the wiki I suggest that we add an https://editorconfig.org/ configuration with our specification. The benefit is that many editors has builtin support for editorconfig (and even more have optional plugins), meaning that they will automatically indent code etc. according to our specification when editing files in this repo.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# EditorConfig is awesome: https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Default rules
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf                # Use Unix line endings
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length = 80            # A line should not have more than this amount
+                                # of chars (not supported by all plugins)
+
+# Makefiles require tabs
+[{Makefile,**.mk}]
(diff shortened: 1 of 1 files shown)


johnwikman:
I generally like this, but the root directory is becoming really crowded now. Can this file replace other things? Such as `mcore-comby.json`, `.ocamlformat`, etc.

br4sco:
It is more general (in the sense that it is agnostic to the language), but it only governs indentation style and whitespace. It cannot replace ocamlformat, which is a full-fledges code formatter, for example. I still think this is better than only specifying this in the wiki (as we do now).

br4sco:
Regarding `.ocamlformat`, we could move that to the `src/boot` where the OCaml code lives.

johnwikman:
By the way, could we perhaps have this file format as an interface for pretty printing in the future as well? I.e. the pretty printer uses a editorconfig config (a default setting or loaded from file) to determine the style of the output.

Something like `mi fmt <file>` would then look for an .editorconfig file to determine how it should pretty print the program.

From the spec ([link](https://spec.editorconfig.org/#id11)):
> EditorConfig plugins shall ignore unrecognized keys and invalid/unsupported values.

So I guess that we can add basically any additional information we need to a .editorconfig file?

PR #876, opened 2024-11-20
Json debug in phase stats

elegios:
This is based on #738 until it is merged.

This PR adds a new flag `--debug-phase <phase>` that uses `mexpr/json-debug.mc` to dump the AST at a named phase, which is specified with `endPhaseStats` in the code. The option can be given multiple times to print multiple phases. No spell-checking is done, since we don't have a data structure that stores the existing phases, so you'd typically use this by running with `--debug-phases` first (note the plural) then with the appropriate `--debug-phase` (or looking in the code of course).

I also add a new function in `json.mc` for printing the json representation directly, because building the interim string was *very* prohibitive. I suppose this is something of a known thing performance-wise, but it might also be relevant for `pprint.mc` when it's used during compilation, maybe we could make compilation faster if we could change this.

### DIFF
--- a/src/main/compile.mc
+++ b/src/main/compile.mc
@@ -41,7 +41,8 @@ lang MCoreCompile =
   MExprConstantFold +
   OCamlTryWithWrap + MCoreCompileLang + PhaseStats +
   SpecializeCompile +
-  PprintTyAnnot + HtmlAnnotator
+  PprintTyAnnot + HtmlAnnotator +
+  MExprToJson
 end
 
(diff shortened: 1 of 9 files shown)


marten-voorberg:
I love this!

I think it might be nice to have a pretty print option too. So you can choose whether you want your debug information as a pprint or as a JSON.

I have some changes on one of my branches making it so the `debugPhases` works on the `MLangProgram` type as well by the way. So these changes will conflict when we eventually merge but the conflict should be minor.

PR #878, opened 2024-11-21
Add printErrorLn to common.mc

br4sco:


marten-voorberg:
Minor note on the naming here. Since we have `print` and `printLn` for outputing to stdout, should we maybe call this `errorLn` for consistency?

br4sco:
> Minor note on the naming here. Since we have `print` and `printLn` for outputing to stdout, should we maybe call this `errorLn` for consistency?

We have an intrinsic `error`, which exists in a program with an error code (and a message). I named it `printErrorLn` because we have an intrinsic, `printError`, which prints a string to stderr without a newline without exiting the program (hence the same naming convention as `print` and `printLn`)

PR #884, opened 2024-12-16
feat: Add `mapErrors` to `Result`

didrikmunther:


stdlib/result.mc
```diff
@@ -149,6 +149,21 @@ let _mergeErrors
...
     { warnings = mapUnion a.warnings b.warnings, errors = mapUnion a.errors b.errors }
 
+-- Map the errors, if any, inside the `Result`. Preserves all warnings.
+let _mapErrors
+  : all w. all e1. all e2. all a. (e1 -> e2) -> Result w e1 a -> Result w e2 a
+  = lam f. lam start.
+    switch start
+    case ResultOk r then ResultOk r
+    case ResultErr { warnings = warnings, errors = errors } then
+      let errors = map f (mapValues errors) in
```
br4sco: I suggest you use `mapMap` from `map.mc` instead of retrieving the values as a sequence and building a new map from the mapped values.

stdlib/result.mc
```diff
@@ -149,6 +149,21 @@ let _mergeErrors
...
+      let f = lam acc. lam err. mapInsert (gensym ()) err acc in
+      ResultErr {
+        warnings = warnings,
+        errors = foldl f (_emptyMap ()) errors
+      }
+    end
+
```
br4sco, line 161: It would be good to add some tests for this function.

stdlib/result.mc
elegios: The current implementation makes new symbols for the keys, i.e., it's not just a `mapMap`. It could be a `mapFoldWithKey` though to avoid creating an intermediate value.

I think the reason I didn't add this function originally is that it's not quite obvious what the semantics should be. The issue is that errors aren't referentially transparent, i.e., `let r = result.err x in r == r` is true while `result.err x != result.err x`, each call to `result.err` creates a new unique error. This implementation of `mapErrors` creates a new distinct error for each old error, so you end up needing to be very careful about not calling `mapErrors` on the same error multiple times.

The other implementation choice, making a `mapErrors` that preserves the identity of each error, but changes the resulting value, could conceptually lose errors, e.g., `result.withAnnotations (result.mapErrors foo r) (result.mapErrors bar r)` would only contain the errors transformed via `bar`, not `foo`. It now becomes important that we only perform equivalent transforms when using `mapErrors`, or we might have surprising results.

As a result I don't think we should have a bare `mapErrors`, it's a footgun no matter the semantics. We could have, e.g., a `mapErrorsNewIdentities` and `mapErrorsKeepIdentities` or something, so you can be explicit. I *think* the `KeepIdentities` version is the one you want most of the time.

stdlib/result.mc
didrikmunther: Yes!

stdlib/result.mc
didrikmunther: Makes sense! I changed it to be named `mapErrorsKeepIdentities` and to use `mapMap` instead.

PR #885, opened 2024-12-17
feat: Add cleanFilePath

didrikmunther:


stdlib/fileutils.mc
```diff
@@ -61,6 +61,39 @@ utest eraseFile "foo.mc" with ""
 utest eraseFile "a/b/c/foo.mc" with "a/b/c"
 utest eraseFile "a/b/c/../foo.mc" with "a/b/c/.."
 
+-- Clean up a file path by removing "." and ".." parts
+let cleanFilePath : String -> String = lam filepath.
```
elegios, line 65: I have an alternate implementation of this on my in progress branch: https://github.com/elegios/miking/commit/43aafa9cdbd948f4c73ffdbeb75d731b394bb22c#diff-f9a9b7d9bc00feb5e0b44f710652e081c01e967d5d077bd8294d3b58583f35f3R25-R38

Mine is intended to fully normalize a path, i.e., make it absolute and such as well. We should probably make sure sync implementations.

didrikmunther: I see, then I will use this locally and wait until your solution is merged! Closing this.

PR #894, opened 2025-01-15
Combine lambdas in OCaml output

elegios:
This PR uses syntactic sugar in OCaml to write, e.g., `let f = fun a -> fun b -> ...` as `let f a b = ...`, and similarly for merging anonymous functions. This has some benefits:
- Reduces whitespace, before we would indent one level per `let` and `fun`, now just once for `let`.
- Shorter code, might be easier to read.
- After switching from dune to ocamlfind (#873) I eventually found that backtraces in gdb had every function named as just `fun`, i.e., it only saw anonymous functions. I have no idea how the two tools invoke `ocamlopt` differently to get that effect, but merging arguments like done in this PR recovers a useful backtrace.

### DIFF
--- a/src/stdlib/ocaml/pprint.mc
+++ b/src/stdlib/ocaml/pprint.mc
@@ -373,35 +373,36 @@ lang OCamlPrettyPrint =
                 "\"", nameGetStr t.nativeIdent, "\";;"])
   | OTopLet t ->
     let indent = 0 in
-    match pprintVarName env t.ident with (env, ident) then
-      match pprintCode (pprintIncr indent) env t.body with (env, body) then
-        (env, join ["let ", ident, " =", pprintNewline (pprintIncr indent),
-                    body, ";;"])
-      else never
-    else never
+    match pprintVarName env t.ident with (env, ident) in
+    match collectParameters env t.body with (env, parameters, body) in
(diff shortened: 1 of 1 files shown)


elegios:
We concluded during the meeting we wanted to see how this might impact performance. Here's the bootstrapping process before:

```
 2) [163.162s] [build] src: ./boot eval main/mi-lite.mc -- 0 main/mi-lite.mc ../build/src/mi-lite        
 1) [49.822s] [build] src: ./mi-lite 1 main/mi.mc ../build/src/mi1                                       
 0) [50.277s] [build] src: ./mi1 compile main/mi.mc --output ../build/src/mi 
 ```
and after:
```
 2) [153.074s] [build] src: ./boot eval main/mi-lite.mc -- 0 main/mi-lite.mc ../build/src/mi-lite        
 1) [31.709s] [build] src: ./mi-lite 1 main/mi.mc ../build/src/mi1                                       
 0) [31.805s] [build] src: ./mi1 compile main/mi.mc --output ../build/src/mi
```

I've done a few runs, and the first step seems to be roughly equal time taken in the two versions, while step 2 and 3 seem to consistently go from ~50s to ~30s, which is... weird, to put it mildly. The intermediate `.ml` file that's passed to `ocamlopt` goes from 33MB to 25MB.

PR #905, opened 2025-02-17
.editorconfig should follow EditorConfig spec regarding inline comments

wmuth:
Hi,

The EditorConfig spec does not include inline comments since version 0.15.0.

See the relevant part of the spec [here](https://spec.editorconfig.org/#no-inline-comments).

This means that whenever I open my editor the built-in EditorConfig parser fails as it parses the value of "end_of_line" to be

```
lf                # Use Unix line endings
```

instead of

```
lf
```

and therefore resulting in the error

```
end_of_line must be one of "lf", "crlf", or "cr"
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,12 +5,14 @@ root = true
 
 # Default rules
 [*]
-end_of_line = lf                # Use Unix line endings
+# Use Unix line endings
+end_of_line = lf
 charset = utf-8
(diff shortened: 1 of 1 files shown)


wmuth:
I also made the identical change to miking-dppl. If you merge both (the dppl PR is linked above this comment) and I'll remove the dppl change from the request reviews channel.

PR #906, opened 2025-02-20
feat: Add LSP skeleton architecture

didrikmunther:


src/stdlib/lsp/utils.mc
```diff
@@ -0,0 +1,21 @@
+include "ext/file-ext.mc"
+
+-- returns Option content if the requested number of bytes could be read
+-- otherwise, None is returned
+recursive
+  let readBytesBuffered : ReadChannel -> Int -> Option [Int] =
```
didrikmunther: Rewrite to tail recursive!

Issue #927, opened 2025-03-17
Test script does not check if a cheat compiler exists, crashes


wmuth:
# Issue
When running the `misc/test` script, it will crash if a cheat compiler does not exist since it is used by default.

## Approx error output
```bash
> misc/test
Using 'make' as a test runner
Included collections:
Excluded collections: ...

CHEAT MI compile ...
No output on stdout

stderr output:
1:        /bin/sh: Line 1: build/mi-cheat: No such file or directory
```

## Steps to reproduce
1. ```rm build/mi-cheat```
2. ```misc/test```

### Current fix
Simply run `make cheat` or pass the `--bootstrapped` or `--installed` flags.

### Ideal fix
Ideally the script should check the existance of a cheat compiler and if it does not exist, it should compile it or use a BOOT/INSTALLED compiler which does exist.

PR #935, opened 2025-03-25
Test script checks existance of mi

wmuth:
## TLDR: This PR addresses issue #927

- `make cheat` implicitly relied on `install` since it just called `mi`, made the dependency explicit
- created `misc/test-utils.sh` for containing extra logic to not bloat up `misc/test`. The file contains logic for checking which mi version the user requested and making sure it exists.
- `misc/test` sources `test-utils.sh` and calls the function inside of it.


Makefile
```diff
@@ -73,7 +73,7 @@ bootstrap: $(if $(wildcard build/$(BOOT_NAME)),,boot)
 	$(SET_STDLIB) $(SET_OCAMLPATH) build/$(MI_MID_NAME) compile src/main/mi.mc --output build/$(MI_NAME)
 
 .PHONY: cheat
-cheat:
+cheat: install
```
elegios: I would prefer to have a check and an informative error message instead. This dependency will unconditionally run `install` each time you run `cheat` (because both are `PHONY`), potentially overwriting your environment with something you didn't intend (e.g., breaking your install if you're working on something that produces a buggy compiler).

I also use a slightly different setup, where I treat the most recently built `build/mi` as a "known good" compiler that I put on my `PATH` when I'm working on miking (via `misc/direnv/nix.envrc`), meaning that `install` is the wrong target to make sure there's an `mi` in `PATH` for me.

misc/test
```diff
@@ -24,7 +31,7 @@ if [ misc/test-spec -ot misc/test-spec.mc ]; then
     # NOTE(vipa, 2024-11-22): Use bootstrapped mi, with repo-local stdlib and boot
     MCORE_LIBS=stdlib=$DIR/src/stdlib OCAMLPATH=$DIR/build/lib:$OCAMLPATH build/mi compile misc/test-spec.mc --output misc/test-spec
   else
-    echo "You need 'mi' to setup tup, e.g., by bootstrapping via 'make'"
+    echo "'mi' seems to have been removed after existance check. Please re-reun this script."
```
elegios: I think changing this error message is good, because I seem to have messed up and copied the message from the `setup-tup` script rather than make it relevant to here.

On a related note, maybe it'd be possible to move this code to your new `test-utils.sh`? I believe it's shared with at least the `setup-tup` script.

misc/test
```diff
@@ -16,6 +18,11 @@ if [ ! "$DIR" -ef "$TARGETDIR" ]; then
   exit 1
 fi
 
+# NOTE(wmuth, 2025-03-25): This function makes sure that installed mi exists if
+# user passes --installed and same for --bootstrapped, --cheated or none of them
+# Defined in ./test-utils.sh
+ensure_mi_exists "$@"
```
elegios: This isn't necessarily doing the right thing: if you use the `tup` runner instead of `make` then `tup` will build its own bootstrapped or cheated version of `mi`, so you don't need to build it ahead of time. `tup` can do this properly because it has accurate dependency tracking without assistance from other tools, so it can know precisely when things are out of date, which is not viable with `make` without _a lot_ of manual work and tool support, which is why the `make` targets just unconditionally build things when you ask for them.

misc/test-utils.sh
```diff
@@ -0,0 +1,55 @@
...
+        exit 1
+      }
+      echo "Compiled 'mi-cheat'."
+    fi
+  fi
+
+  if [ "$INSTALLED" == "true" ]; then
+    if ! command -v mi >/dev/null 2>&1; then
+      echo "'mi' is not installed. Installing... "
+      make -C "$DIR" install || {
```
elegios: Generally I really don't want to implicitly do things that change the user's environment without them explicitly asking for it or at the very least us presenting a confirm dialog beforehand.

misc/test
```diff
@@ -7,6 +7,8 @@ CURR=$(pwd)
 TARGETDIR=".${CURR#"$DIR"}"
 cd "$DIR"
 
+source "$DIR/misc/test-utils.sh"
```
elegios: This script should probably be in the `scripts` sub-directory instead. We're generally trying to reduce the amount of files immediately visible here, to varying degrees of success :P

elegios (changes requested):
_Very_ nice that someone is looking at these scripts with an outside perspective. I'd be a bit cautious of automatically and implicitly doing things that aren't the main task of each script. Instead, I'd prefer to either give an informative error message or an explicit confirmation dialog.

wmuth:
@elegios

I could put all the installs behind y/N prompts. Of course it could also just print error messages but I personally much preferred the script having the ability to also perform the needed change, but I see your point about asking the user before doing anything. EDIT: Should I add a `--noninteractive` flag potentially which skips the y/N prompt and assumes N? Potentially github actions could use this as well though it shouldn't have issues as it bootstraps mi before running the test script so the file exists check is fine.

For the Tup part, potentially I could move the checks to happen after the check wether to use tup or make as the test-runner? In that case the command/file exists check would only be run if we are using make and can't use tup's versions of the binaries. And of coure it can ask the user before actually running the make targets.

elegios:
I think asking before running another `make` target makes sense, as well as taking `tup` or not into account. A non-interactive flag would be nice, but not super-necessary I think, depends on how much it affects the complexity of the script(s), so I think that's up to you. I agree that it shouldn't matter for our automated uses.

wmuth:
Merging this PR closes #927 .

elegios (approved):

PR #939, opened 2025-03-31
FIX: Send array to tup

wmuth:
@david-broman Please merge asap

Currently there is a bug such that only the first arg is sent to tup, not the whole array. Make works fine.


wmuth:
The CI check had an internal error, seems to be an issue unrelated to this change. E: Seems like it had an internet connectivity error while running the setting up of the CI.

PR #940, opened 2025-04-01
Feat: Add -j flag to test script and have make & tup obey it

wmuth:
This PR adds parallelization to the test script `misc/test` using a `-j` flag the same way `make` and `tup` has.

Parallelization is disabled by default and is enabled using the `-j%d` flag. If you provide `-j` without a number then the script automatically detects your number of logical cpu cores and automatically runs that many jobs in parallel.

Before this PR, running `make test -j%d`, would run the tests in parallel since recursive `make` invocations inherit flags trough the `MAKEFLAGS` environment variable. This still works but now `tup` also follows the specified `-j` flag, even if provided to `make` trough e.g. `make test -j5`.

### DIFF
--- a/misc/scripts/test-utils
+++ b/misc/scripts/test-utils
@@ -48,7 +48,6 @@ util_set_vars() {
         echo "'tup' installed, but not set up."
         if util_user_consent "Would you like set up 'tup' now?"; then
           util_make_setup_tup
-          util_vars[use_tup]=1
         else
           echo "Run 'make setup-tup' if you want to use 'tup'."
         fi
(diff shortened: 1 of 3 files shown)


misc/scripts/test-utils
```diff
@@ -8,6 +8,7 @@ declare -A util_vars=(
   [is_cheated]=0
   [is_installed]=0
   [non_interactive]=0
+  [use_threads]=1
```
johnwikman: I don't know how this affects the behavior of the generated Makefile from the test script, but from a practical point of view I would prefer for parallelism to be opt-in. It would be preferable that the default option is not to starve the system resources, especially if running the tests on a server shared by others.

Can we not replicate the Makefile interface for this, with the `-j` flag and an argument to it?

misc/scripts/test-utils
```diff
@@ -8,6 +8,8 @@ declare -A util_vars=(
   [is_cheated]=0
   [is_installed]=0
   [non_interactive]=0
+  [num_threads]=0
+  [use_threads]=0
```
johnwikman: LGTM, but I am guessing that these could be merged from a stylistic point of view? `num_threads = 1` implies `use_threads = 0`, so no need to have both. In this way we only provide the `-j` flag to `make` if `num_threads != 1`.

johnwikman (approved):

PR #956, opened 2025-09-25
Add docgen implementation in stdlib and new command in src/main/mi.mc.

ZivoMartin:
## Summary
Introduces the docgen module in the stdlib, containing the full implementation of the documentation generator.  
The `docgen` command enables automatic generation of stdlib documentation.
Also adds the necessary integration in `/main` to expose it as a command, following the same structure as the `tune` subcommand.

---

## New files in `/main`
- `docgen.mc`  
  Defines the `docgen` command entry point, which delegates to the implementation in `stdlib/docgen/docgen.mc`.  
- `docgen-config.mc`  
  Implements the parser for the `DocgenOptions` type, based on `ParseConfig`.

---

## Modified files in `/main`
- `mi.mc`  
  - Added `docgen` usage line.  
  - Added a new entry in the `commandsMap`.  
- `options-type.mc`  
  Added a `docgenOptions` field for storing command options.  
- `options.mc`  
  Integrated the default `DocgenOptions` into the default `Options` type.

---

## How to test
Run the following on any valid miking file:
```bash
mi docgen file.mc
```


elegios:
The tests fail for this PR, for a somewhat boring reason: it uses some externals that aren't available in the interpreter. The typical fix is to mark the files in the test suite that end up needing externals such that they're only tested in compiled mode.

You can look in `misc/test-spec.mc`, find the comment saying `Files using externals not available in the interpreter`, then add files to that list. Since most of your changes are in `src/stdlib/docgen` I'm guessing you can run `misc/test --installed smart -- src/stdlib/docgen/**/*.mc` while looking for files to mark, then `make test` to be sure you've caught everything.

See [here](https://miking.org/docs/reference/testing-miking) for more information on the test suite, or ping me if you have questions.

ZivoMartin:
Hey, I fixed the tests, let me know if any problems remain :)

PR #963, opened 2025-10-27
Optimize compiler code generation

larshum:
This PR includes a number of smaller changes to the compiler intended to reduce the amount of generated code, thereby reducing compilation times (as a lot of time is spent in the OCaml compiler). It includes the code for PRs https://github.com/miking-lang/miking/pull/961 and https://github.com/miking-lang/miking/pull/962.

After this PR, the compiler will:
- Print ASCII strings as string literals in the OCaml AST, drastically reducing the amount of generated code.
  - Previously, each character was printed as an encoded integer on a separate line.
  - New function were added in boot to efficiently perform this conversion.
- Omit the introduction of unused variables in the `TmSeqEdge` pattern.
  - Also, the length is computed in the then-branch rather than before the match, and only when it is needed.
- Omit the never case of a sequence of matches when all constructors are matched on.


build/tup.config
elegios, line 1: This file is actually important to keep for the `tup` version of building and testing, it let's `tup` to know to build in this folder rather than adjacent to each original file.

larshum: Oh sorry, I don't know why it was removed.

PR #968, opened 2025-12-18
add of_array

gizemcaylak:
Adds `of_array` and integer kind to the external array library.

### DIFF
--- a/src/stdlib/ext/arr-ext.mc
+++ b/src/stdlib/ext/arr-ext.mc
@@ -135,7 +135,7 @@ external externalExtArrGet : all a. ExtArr a -> Int -> a
 external externalExtArrSet ! : all a. ExtArr a -> Int -> a -> ()
 external externalExtArrCopy : all a. ExtArr a -> ExtArr a
 external externalExtArrFill : all a. ExtArr a -> a -> ()
-external externalExtArrOf : all a. ExtArrKind a -> Arr a -> ExtArr a
+external externalExtArrOfArr : all a. ExtArrKind a -> Arr a -> ExtArr a
 
 --------------------------------------------------------------------------------
 -- ExtArr interface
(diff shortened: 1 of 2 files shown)


src/stdlib/ext/arr-ext.mc
```diff
@@ -135,6 +135,7 @@ external externalExtArrGet : all a. ExtArr a -> Int -> a
 external externalExtArrSet ! : all a. ExtArr a -> Int -> a -> ()
 external externalExtArrCopy : all a. ExtArr a -> ExtArr a
 external externalExtArrFill : all a. ExtArr a -> a -> ()
+external externalExtArrOf : all a. ExtArrKind a -> Arr a -> ExtArr a
```
br4sco: May I suggest calling it `externalExtArrOfArr`?

src/stdlib/ext/arr-ext.mc
```diff
@@ -243,3 +247,23 @@ utest
   utest extArrGetExn a 2 with 1. in
   ()
   with ()
+
+-- Build a one-dimensional Bigarray from a given array
+let extArrOf : all a. ExtArrKind a -> Arr a -> ExtArr a
```
br4sco: See name suggestion above.

br4sco: Also, in the comment, the notion of a Bigarray is specific to the OCaml backend. Perhaps we should call it "shareable array" or "external array".

gizemcaylak: Fixed, thanks for checking!

src/stdlib/ext/arr-ext.mc
gizemcaylak: Fixed!

PR #987, opened 2026-04-20
Update flake and ocamlformat

elegios:
When setting up a new test environment on a new computer I wanted to give it a later version of OCaml. Unfortunately this makes it difficult to install the version of `ocamlformat` we require, so this PR updates that. It also updates the nix flake, since that was required to install the new version of `ocamlformat`.

### DIFF
--- a/misc/packaging/flake.nix
+++ b/misc/packaging/flake.nix
@@ -20,7 +20,7 @@
           devShells.default = pkgs.mkShell {
             name = "Miking dev shell";
             inputsFrom = [ packages.miking-lib packages.miking-unwrapped ];
-            buildInputs = [ pkgs.tup pkgs.ocamlformat_0_29_0 pkgs.ocamlPackages.owl pkgs.ocamlPackages.utop pkgs.gdb ];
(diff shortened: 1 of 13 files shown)


misc/packaging/flake.nix
```diff
@@ -20,7 +20,7 @@
           devShells.default = pkgs.mkShell {
             name = "Miking dev shell";
             inputsFrom = [ packages.miking-lib packages.miking-unwrapped ];
-            buildInputs = [ pkgs.tup pkgs.ocamlformat_0_24_1 ];
+            buildInputs = [ pkgs.tup pkgs.ocamlformat_0_29_0 pkgs.ocamlPackages.owl pkgs.ocamlPackages.utop pkgs.gdb ];
```
br4sco, line 23: Since this list grows a bit, how about:
```suggestion
            buildInputs = with pkgs; [ tup ocamlformat_0_29_0 ocamlPackages.owl ocamlPackages.utop gdb ];
```
I know, I'm procrastinating on other things.

PR #993, opened 2026-05-28
Make multiple runs of the effects.mc test not interact

elegios:
There were (very) occasional failures in the test suite, caused by a test case for interacting with files interacting with a fixed path. This is an issue if the same file is tested in parallel with itself, e.g., through running a compiled executable and interpreting the original file, which would sometimes happen.

### DIFF
--- a/src/test/mexpr/effects.mc
+++ b/src/test/mexpr/effects.mc
@@ -3,6 +3,8 @@
 --
 -- Test different instrinsics with side effects
 
+include "sys.mc"
+
 mexpr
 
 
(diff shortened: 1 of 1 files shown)


src/test/mexpr/effects.mc
```diff
@@ -19,7 +21,8 @@ utest if false then dprint ["message\n","hi"] else () with () in
 -- 'writeFile fname s' writes a string 's' to a file named 'fname'.
 -- String -> String -> ()
 let str1 = "A unicode string.\n島" in
-let file = "_testfile" in
+let dir = sysTempDirMake () in
```
johnwikman: Is there a reason for not using `sysWithTempDir` here? Does that mess up the utests for the effects?

PR #1000, opened 2026-06-11
First steps on the new parser

mrxdst:
The parser is designed to be as extensible as possible.
It is built on top of the breakable library.

This is a new self-contained file, should not affect anything else.
The parser is still very much work in progress but it works with limited functionality.

Currently the parser is tested for equivalence with the boot parser.

Known issues as of now.
* Much is not implemented
* Ambiguity error reporting does not use the correct source string.
* Precedence handling (groupingsAllowed) will need a new solution.
* No tests for parsing that should fail.

### DIFF
--- a/src/stdlib/parser/parser.mc
+++ b/src/stdlib/parser/parser.mc
@@ -7,10 +7,12 @@ include "mexpr/json-debug.mc"
 include "json.mc"
 include "seq.mc"
 include "parser/breakable.mc"
+include "name.mc"
 
 type BreakableOp lstyle rstyle
-con OpAtom: all self. self -> BreakableOp LClosed RClosed
-con OpNeg: all self. self -> BreakableOp LClosed ROpen
+con OpAtom: use Ast in Expr -> BreakableOp LClosed RClosed
+con OpNeg: Info -> BreakableOp LClosed ROpen
+con OpApp: Info -> BreakableOp LClosed ROpen
 
 lang AstParserBase = Lexer
   sem config: () -> Config BreakableOp
@@ -21,7 +23,7 @@ lang AstParserBase = Lexer
   sem parenAllowed: ParenAllowedFunc BreakableOp
   sem groupingsAllowed: GroupingsAllowedFunc BreakableOp
 
-  sem constructPrefix: BreakableOp LClosed ROpen -> Expr -> Expr
+  sem constructPrefix: (BreakableOp LClosed ROpen, Expr) -> Expr
 
   sem parseExpr: NextTokenResult -> (Expr, NextTokenResult)
   sem parseExprRClosed : State BreakableOp RClosed -> NextTokenResult -> (Expr, NextTokenResult)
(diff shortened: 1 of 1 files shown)


src/stdlib/parser/parser.mc
```diff
@@ -0,0 +1,912 @@
...
+  sem canStartAppArgExpr =
+  | { token = LIdentTok { } } -> true
+
+  sem parseExprROpen state =
+  | { token = LIdentTok { val = val } } & cur ->
+    let expr = TmVar {
+      ident = nameNoSym val,
+      ty = ityunknown_ cur.info,
+      info = cur.info,
+      frozen = false -- TODO: Always false?
```
elegios: Yes, it's a separate syntax for when it's true: `#frozen"foo"` becomes a `TmVar {ident = nameNoSym "foo", frozen = true, ...}`.

src/stdlib/parser/parser.mc
```diff
@@ -0,0 +1,912 @@
...
+      lhs = lhs,
+      rhs = rhs,
+      info = info
+    }
+end
+
+lang ParenParser = AstParserBase + RecordAst + RecordTypeAst
+  sem canStartAppArgExpr =
+  | { token = LParenTok {} } -> true
+  | { token = RParenTok {} } -> false
```
elegios: I think it's probably better to just work with (and strongly assume) that the default case is `false`. That gives fewer cases, and also makes it possible for some other fragment to make the given token permissible after `TmApp` (not that that's particularly likely for `RParenTok` in particular, but eh).

src/stdlib/parser/parser.mc
```diff
@@ -0,0 +1,912 @@
...
+          parseTypeRClosed state (nextToken close.stream)
+        else
+          parseErr (cur.info, "Missing closing parenthesis")
+      )
+end
+
+lang BoolParser = AstParserBase + BoolAst
+  sem canStartAppArgExpr =
+  | { token = LIdentTok { val = "true" } } -> true
+  | { token = LIdentTok { val = "false" } } -> true
```
elegios: These could be just one case:

```
| { token = LIdentTok { val = "true" | "false" } } -> true
```

They should compile to the same thing, so it's just a stylistic thing.

src/stdlib/parser/parser.mc
```diff
@@ -0,0 +1,912 @@
...
+        parseTypeRClosed state (nextToken cur.stream)
+      else
+        parseErr (cur.info, "Missing right bracket")
+    )
+
+end
+
+lang LetDeclParser = AstParserBase + LetDeclAst
+  syn BrkOpExpr lstyle rstyle =
+  | OpExprLet Decl
```
elegios: This should probably be called something with `Decl` rather than `Let`. It's quite possible also a good idea to move the definition to the main fragment, since it'll probably be shared with all other `Decl`s.

elegios (approved):
Very nice progress. I've left some (rather minor) notes for continued work, but I think this PR can be merged as is.

PR #1007, opened 2026-08-13
Finish the new parser

mrxdst:
Continuation of #1000

* adds parsers for the remaining language syntax.
* fixes bugs
* reworks keyword handling
* added tests for things that should fail to parse.
* added equality comparison for MLang ast nodes
* updated lexer string escape handling
* created a script that compares entire files with the boot parser, and also checks info field consistency.
* added test to test-spec that tests most files using the above mentioned script.
* changed groupings allowed to be category based.
* added cli flag `--native-parser`

points of interest
* TmApp and TmConApp was deliberately made ambiguous. no test failures.


src/stdlib/mexpr/shallow-patterns.mc
```diff
@@ -364,6 +364,14 @@ lang DedupBranchesAndInformativeNever = ShallowBase + NeverAst + AppAst + SeqAst
   sem lowerToExpr env scrutinee branches += | fallthrough ->
     let inconsistentError = lam info. lam name.
       errorSingle [info] (join ["Inconsistent pattern; '", nameGetStr name, "' is not always bound."]) in
+
+    let infoOrFallback = lam info.
```
elegios, line 368: Do you know where the missing `info` field appears? Ideally we make sure there is one everywhere, so this kind of fix is slightly antithetical.

src/stdlib/mlang/eq.mc
```diff
@@ -0,0 +1,151 @@
...
+include "option.mc"
+include "string.mc"
+include "seq.mc"
+include "mexpr/info.mc"
+
+lang UseEq = Eq + UseDeclAst
+  sem eqDeclH (env : EqEnv) (free : EqEnv) (lhs : Decl) +=
+  | DeclUse {ident = i2} ->
+    match lhs with DeclUse {ident = i1} then
+      if nameEqStr i1 i2 then Some (env, free) else None ()
```
elegios, line 18: I think this comparison should look more like the one for `TmVar`.

It might also be the case that you'd rather use `cmp.mc` and its functions for your comparisons, because that's a stricter equality than the alpha-equivalence that `eq.mc` is trying to check for, and also simpler to reason about the correctness of implementation.

src/stdlib/parser/lexer.mc
```diff
@@ -423,6 +452,27 @@ lang OperatorTokenParser = TokenParser
...
+
+  -- Operator characters are lexed by greedily merging a maximal run into a
+  -- single token (e.g. `<&>` lexes as one token, per the utests above),
+  -- which is broader than boot's fixed operator set and can over-merge at
+  -- a grammar position that expects two separate operators back-to-back
+  -- (e.g. `sem f =| pat -> e`, where `=` and `|` are adjacent with no
+  -- space). This splits a known operator prefix off of a token and
+  -- re-lexes the remainder, so callers can recover the intended
+  -- tokenization at the specific grammar positions that need it.
+  sem splitOperatorPrefix: NextTokenResult -> String -> Option NextTokenResult
```
elegios, line 464: I don't really like this kind of re-lexing solution, I'd rather use more precise lexing language fragments to start with, or maybe disallow writing `=|` without a space (the latter would have to be a discussion at a meeting). Do you have a sense of how often it appears?

src/stdlib/mexpr/shallow-patterns.mc
mrxdst: this is due to the info field consistency requirement.
re-parsing the source indicated by the info field should yield the same result.
but this can not work in the case of syntax sugar.

examples are:
* chars in strings.
* TyVariant if a type declaration is missing =.
* Implicit never in match, switch,
* temporary variable for switch
* match from dot projectection.
* match pattern from if/else.
* con type restriction
* .. might be others.

boot parser adds fake info fields. we could do that as well but we would have to get rid of the consistency checks.
the only use of info fields that i could identify as having an impact on program output was in shallow-patterns.
a fallback was added here only.

src/stdlib/mlang/eq.mc
mrxdst: cmp and eq for mlang was not implemented before, only for mexpr. i can try implementing cmp instead. in that case should i keep the eq file for mlang as it is now or just get rid of it?

src/stdlib/parser/lexer.mc
mrxdst: i do not like it either. it was made like this in order to only make backwards compatible changes to the lexer.
but maybe changing the lexer is fine.
a quick search finds about 50 uses of =| in stdlib.

PR #1010, opened 2026-09-08
Add externals for sumlog

gizemcaylak:
For a faster calculation of adding logarithms, needed in coreppl models

### DIFF
--- a/src/stdlib/ext/arr-ext.ext-ocaml.mc
+++ b/src/stdlib/ext/arr-ext.ext-ocaml.mc
@@ -108,5 +108,11 @@ let arrExtMap =
         expr = "(fun kind a -> Bigarray.Array1.of_array kind Bigarray.c_layout a)",
         ty = (tyarrows_ [otyopaque_, otyopaque_, otyopaque_])
       }
+    ]),
+    ("externalExtArrSumLogFloat64", [
+      impl {
+        expr = "(fun (a : (float, Bigarray.float64_elt, Bigarray.c_layout) Bigarray.Array1.t) (n : int) -> let acc = ref 0.0 in for i = 0 to n - 1 do acc := !acc +. Float.log (Bigarray.Array1.unsafe_get a i) done; !acc)",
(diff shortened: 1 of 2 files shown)


src/stdlib/ext/arr-ext.ext-ocaml.mc
```diff
@@ -108,5 +108,11 @@ let arrExtMap =
         expr = "(fun kind a -> Bigarray.Array1.of_array kind Bigarray.c_layout a)",
         ty = (tyarrows_ [otyopaque_, otyopaque_, otyopaque_])
       }
+    ]),
+    ("externalExtArrSumLogFloat64", [
+      impl {
+        expr = "(fun (a : (float, Bigarray.float64_elt, Bigarray.c_layout) Bigarray.Array1.t) (n : int) -> let acc = ref 0.0 in for i = 0 to n - 1 do acc := !acc +. Float.log (Bigarray.Array1.unsafe_get a i) done; !acc)",
```
br4sco, line 114: I think you need to verify that the input `n` is in the range `0` ... `len(arr)  - 1` before the loop, since you use `unsafe_get`. Otherwise, `extArrKindFloat64 [2., 3., 5.]) 4` gives undefined behavior since we read outside of the array memory, for example.

br4sco: Another option is to mark this function as unsafe, but I don't know if we want that.

br4sco: A third option would be to remove the `n` parameter completely and always sum over the whole array. You could then add another external array operation that retrieves a sub-array (cf. https://ocaml.org/manual/5.5/api/Bigarray.Array1.html#VALsub) to cover the use-case where you only want to sum over a subset of the array.

gizemcaylak: Thanks for the review! Now, it always sums over the whole array

